PR #14, opened 2021-04-15
Setup MDX/Markdown pages

codemonkey800:
## Description

This PR adds MDX tooling to enable support for MDX/Markdown pages in Next.js. This sets up an example page using `/about`. Right now it's a very basic page, but future work will replace the page with the correct designs.

## Demo

image


kne42 (approved):
yay MDX is suuper cool! LGTM :)

tihuan (approved):
Wow this is indeed super cool! Didn't know about this library 😆 TIL! Thanks for tagging me on this 🎉

PR #18, opened 2021-04-21
Workflows for linting and testing client code

codemonkey800:
## Description

This PR adds GitHub workflows for testing PRs and pushes to the main branch. PRs will have jobs that check for code formatting using prettier, code linting using stylelint and eslint, and unit / integration tests with Jest.

When a PR is merged, it'll also run E2E tests for chromium and firefox on different screen sizes to test functional changes in the UI (for example, the smaller layouts have a menu drawer while larger layouts do not).

### [Workflow for PR](https://github.com/codemonkey800/napari-hub/actions/runs/771849475)

image

### [Workflow for Pushing to main](https://github.com/codemonkey800/napari-hub/actions/runs/771851307)

image

## Follow Ups

- E2E tests are only testing against chromium and firefox. From the survey results, several people were using Webkit and Edge for their browsers, so we'll need to investigate if we need to E2E tests for those engines. I know that Edge is now based on chromium, so we may not need to do any work there, but it would be good to confirm first.


kne42 (approved):
ah this CI code is so beautiful! 😻

PR #28, opened 2021-04-30
Plugin Support Info

codemonkey800:
## Description

Depends on #27

This PR adds the plugin support info section to the plugin details page.

## Changes

- Added 3 fractional unit column grid class
- New components:
  - MediaFragment: Component for rendering media query components without a wrapper `<div />`
  - SupportInfo: Component for rendering project authors, support links, and source code links

## Demos

**Demo Link**: https://napari-hub-support-info-demo.vercel.app/plugin/name

### Responsive Support Info

https://user-images.githubusercontent.com/2176050/116639386-4a434380-a91d-11eb-8302-f63769331697.mp4

### Missing Information

info-not-submitted


justinelarsen (approved):
Looks good to me - Nice work!

client/src/components/PluginDetails/MetadataList.module.scss
```diff
@@ -0,0 +1,13 @@
...
+ */
+.commaList {
+  &::after {
+    content: ', ';
+  }
+
+  &:last-child::after {
+    content: '';
+  }
+}
```
kne42, line 13: interesting! I'm guessing if someone were to copy/paste this, it would result in a `ul` instead of having the commas? also is there a performance reason or something why you choose to use two statements here instead of a `not`

kne42: also why is the css class name in camelcase? I thought it was automatically translated from snake to camel but maybe that's not the best practice to do, could you clarify please?

kne42 (commented):
nice, LGTM too! also no need to wait up for my reviews since they're probably not useful lol

also left some more questions generally about best practices and whatnot

client/src/components/PluginDetails/MetadataList.module.scss
codemonkey800: > also why is the css class name in camelcase?

oh so they actually don't get translated into camel case automatically. the reason why we're using camel case is because it's nicer to work with in JS. If we named this `.comma-list`, then in JS I would need to access it using `styles['comma-list']` vs `styles.commaList`.

kne42: ohhh, that makes sense

PR #29, opened 2021-05-05
Client proxy for local hub API access

codemonkey800:
## Description 

This PR adds a script for locally forwarding a port to the hub API securely over SSH. With this, it's possible to interact with the hub API on `localhost`.

Check out #30 for a demo of this being used.


client-proxy/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+## SSH Setup
+
+This client assumes the infrastructure is using
+[blessclient](https://github.com/chanzuckerberg/blessclient) for managing SSH
+certificates. Install blessclient and run the following to gain access to the
+AWS imaging infrastructure:
+
+```sh
+blessclient import-config git@github.com:chanzuckerberg/sci-imaging-infra/blessconfig.yml
```
codemonkey800, line 15: @justinelarsen @ziyangczi I just realized this is a private repo, is there any harm in including this in the README or should I change it to ta placeholder?

ziyangczi: for dev iteration there are probably other ways to access the API, I can discuss this separately with you. I didn't know that you need something like this (Should have thought about setting up dev connections) but we can chat

codemonkey800: Yeah I'm open to easier ways to do this. I wrote this script out of necessity while working on the hub API in #30. I'm completely fine with this being throwaway code if there's an easier way 😄

codemonkey800:
Closing in favor of simpler methods discussed with @ziyangczi. We may revisit adding a script like this in the future

PR #30, opened 2021-05-05
Fetch plugin data from backend

codemonkey800:
## Description

This PR sets up fetching plugin data from the hub backend API.

## Changes

- Fixes lodash-es issue on server
  - We use `lodash-es` for tree shaking in the browser
  - `lodash-es` doesn't work on the server: https://github.com/vercel/next.js/issues/2259
  - This can be resolved by importing from `lodash` everywhere and using [webpack aliases](https://webpack.js.org/configuration/resolve/#resolvealias) to replace `lodash` with `lodash-es` in the browser
- Moved PluginStateContext up to the page level
  - This makes sense architecturally because the page should handle fetching and providing data through context
- Renamed route `/plugin/:name` to `/plugins/:name` to be consistent with backend API
- Mock server for hub API
- ErrorMessage component
- Show list of plugin links on `/plugins` page for testing

## Demos

The below demo uses the client proxy (#29) to render the plugin listing and data.

https://user-images.githubusercontent.com/2176050/117193991-679e6480-ad98-11eb-9a88-d135c27e33d5.mp4

PR #31, opened 2021-05-06
add backend lambda that handles the API

ziyangczi:
Add the backend logic in lambda py to handle the API calls

### DIFF
--- a/backend/lambda.py
+++ b/backend/lambda.py
@@ -375,7 +375,7 @@ def send_alert(message):
         "text": message
     }
     if slack_url is None:
-        print("Unable to send alert")
+        print("Unable to send alert because slack URL is not set")
     else:
         try:
             requests.post(slack_url, json=payload)
--- a/backend/lambda.py
+++ b/backend/lambda.py
@@ -1,0 +1,433 @@
+import concurrent.futures
+import boto3
+from botocore.exceptions import ClientError
+from google.cloud import bigquery
+import requests
+from requests.exceptions import HTTPError
+from requests.utils import requote_uri
+import json
+import os
+import re
+from datetime import datetime, timedelta, timezone
+import tempfile
+import yaml
+
+bucket = os.environ.get('BUCKET')
+slack_url = os.environ.get('SLACK_URL')
+cache_ttl = int(os.environ.get('TTL', "4"))
+plugins_key = 'cache/plugins.json'
+index_key = 'cache/index.json'
+exclusion_list = 'excluded_plugins.json'
+index_subset = {'name', 'summary', 'description', 'description_content_type',
+                'authors', 'license', 'python_version', 'operating_system',
+                'release_date', 'version', 'first_released',
+                'development_status'}
+
+s3 = boto3.resource('s3')
+s3_client = boto3.client("s3")
+cache_ttl = timedelta(minutes=cache_ttl)
+
+
+def handler(event, context):
+    """
+    Entrypoint for lambda handler.
+
+    :param event: event for the lambda invoke to handle
+    :param context: context about this invoke
+    :return: result for the invoke
+    """
+    if 'index' in event:
+        return get_index(context)
+    elif 'exclusion' in event:
+        return get_exclusion_list()
+    elif 'version' in event and 'plugin' in event:
+        return get_plugin(event['plugin'], event['version'])
+    elif 'plugin' in event:
+        return get_plugin(event['plugin'])
+    else:
+        return get_plugins(context)
+
+
+def get_attribute(obj, path):
+    """
+    get attribute iteratively from a json object.
+
+    :param obj: object to iterate on
+    :param path: list of string to get subpath within json
+    :return: the value if the path is accessible, empty string if not found
+    """
+    part = obj
+    for token in path:
+        if isinstance(part, dict) and token in part:
+            part = part[token]
+        elif isinstance(part, list) and token < len(part):
+            part = part[token]
+        else:
+            return ""
+    return part
+
+
+def filter_prefix(str_list, prefix):
+    """
+    filter the list for strings with the given prefix.
+
+    :param str_list: list of strings to filter
+    :param prefix: prefix to filter on
+    :return: list of filtered strings
+    """
+    filtered = []
+    for string in str_list:
+        if string.startswith(prefix):
+            filtered.append(string)
+    return filtered
+
+
+def filter_index(plugin, version):
+    """
+    filter index based to only include specified entries.
+
+    :param plugin: name of the plugin
+    :param version: version of the plugin
+    :return: filtered json metadata for the plugin
+    """
+    plugin_info = get_plugin(plugin, version)
+    ret = {}
+    for k, v in plugin_info.items():
+        if k in index_subset:
+            ret[k] = v
+    return ret
+
+
+def get_index(context):
+    """
+    get the index page related metadata for all plugins.
+
+    :param context: context for the run to raise alerts for
+    :return: json for index page metadata
+    """
+    if cache_available(index_key, cache_ttl):
+        return get_cache(index_key)
+    results = []
+    with concurrent.futures.ThreadPoolExecutor(max_workers=32) as executor:
+        futures = [executor.submit(filter_index, k, v)
+                   for k, v in get_plugins(context).items()]
+    for future in concurrent.futures.as_completed(futures):
+        results.append(future.result())
+    return cache(results, index_key)
+
+
+def get_file(download_url, file):
+    """
+    get file from github.
+
+    :param download_url: github url to download from
+    :param file: filename to get
+    :return: file context for the file to download
+    """
+    api_url = download_url.replace("https://github.com/",
+                                   "https://api.github.com/repos/")
+    try:
+        response = requests.get(
+            f'{api_url}/{file}')
+        if response.status_code != requests.codes.ok:
+            response.raise_for_status()
+        info = json.loads(response.text)
+        if "download_url" in info:
+            return requests.get(info["download_url"]).text
+    except HTTPError:
+        pass
+
+    return None
+
+
+def get_extra_metadata(download_url):
+    """
+    extract extra metadata from the github download url
+
+    :param download_url: github url to download from
+    :return: extra metadata dictionary
+    """
+    extra_metadata = {}
+
+    description = get_file(download_url, "contents/.napari/DESCRIPTION.md")
+
+    if description is not None:
+        extra_metadata['description'] = description
+
+    yaml_file = get_file(download_url, "contents/.napari/config.yml")
+    if yaml_file:
+        config = yaml.safe_load(yaml_file)
+        for k, v in config.items():
+            extra_metadata[k] = v
+
+    return extra_metadata
+
+
+def format_plugin(plugin):
+    """
+    format the plugin dictionary to extra relevant information.
+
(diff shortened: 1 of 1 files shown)


backend/lambda.py
```diff
@@ -0,0 +1,433 @@
...
+
+
+def get_plugins(context):
+    """
+    get all valid plugins list.
+
+    :param context: context for the run to raise alerts
+    :return: json of valid plugins and their version
+    """
+    if cache_available(plugins_key, cache_ttl):
```
justinelarsen, line 229: I think a comment would be helpful here to explain the methods we're using for getting plugin list like first try cache, then query pypi, then try bigQuery

backend/lambda.py
```diff
@@ -0,0 +1,433 @@
...
+
+def filter_excluded_plugin(packages):
+    """
+    filter excluded plugins from the plugins list
+    :param packages: all plugins list
+    :return: only plugins not in the filtered list
+    """
+    filtered = {}
+    excluded = get_exclusion_list()
+    for package in packages:
```
justinelarsen, line 318: Since excluded packages will be a subset of the package list, would it be more efficient to loop through the exclusion list instead and remove them from the full list? It probably doesn't make a difference now, but thinking about scalability to plan for many plugins.

backend/lambda.py
```diff
@@ -0,0 +1,433 @@
...
+    name_pattern = re.compile('class="package-snippet__name">(.+)</span>')
+    version_pattern = re.compile(
+        'class="package-snippet__version">(.+)</span>')
+    url = requote_uri(f"https://pypi.org/search/?c=Framework :: napari&page=")
+
+    while True:
+        try:
+            response = requests.get(f'{url}{page}')
+            if response.status_code != requests.codes.ok:
+                response.raise_for_status()
```
justinelarsen, line 355: Is this going to be called every time we get to the last page as we have no way of knowing what the last page on pypi is?

backend/lambda.py
```diff
@@ -0,0 +1,433 @@
...
+    """
+    send alert to slack with a message.
+
+    :param message: message to send alongside the alert
+    """
+    payload = {
+        "text": message
+    }
+    if slack_url is None:
+        print("Unable to send alert")
```
justinelarsen, line 378:
```suggestion
        print("Unable to send alert because slack URL is not set")
```

backend/lambda.py
```diff
@@ -0,0 +1,433 @@
...
+
+    :return: list of plugin name and version
+    """
+    client = bigquery.Client()
+    results = client.query(
+        """
+            select name, max(version) as version from 
+            `bigquery-public-data.pypi.distribution_metadata`  
+            where "Framework :: napari" in UNNEST(classifiers) 
+            group by name;
```
justinelarsen, line 398: Should these be ordered ascending alphabetically?

backend/lambda.py
```diff
@@ -0,0 +1,433 @@
...
+def cache(content, key):
+    """
+    cache the given content to the key location.
+
+    :param content: content to cache
+    :param key: key path in s3
+    :return: content that is cached
+    """
+    if bucket is None:
+        send_alert(f"({datetime.now()}) Unable to find bucket for lambda "
```
justinelarsen, line 424: I love all the detailed messages! This will be very helpful for troubleshooting :)

justinelarsen (commented):
Great job @ziyangczi ! Overall, it looks great. I added some comments which are minor or questions.

backend/lambda.py
```diff
@@ -0,0 +1,433 @@
...
+def filter_index(plugin, version):
+    """
+    filter index based to only include specified entries.
+
+    :param plugin: name of the plugin
+    :param version: version of the plugin
+    :return: filtered json metadata for the plugin
+    """
+    plugin_info = get_plugin(plugin, version)
+    ret = {}
```
neuromusic, line 94: `ret = {k: plugin_info[k] for k in index_subset}` avoids looping through the full dict

similarly if you don't want a dict comprehension...

```
ret = {}
for k in index_subset:
    ret[k] = plugin_info[k]
```

backend/lambda.py
```diff
@@ -0,0 +1,433 @@
...
+
+def filter_prefix(str_list, prefix):
+    """
+    filter the list for strings with the given prefix.
+
+    :param str_list: list of strings to filter
+    :param prefix: prefix to filter on
+    :return: list of filtered strings
+    """
+    filtered = []
```
neuromusic, line 78: `filtered = [string for string in str_list if string.startswith(prefix)]`

backend/lambda.py
```diff
@@ -0,0 +1,433 @@
...
+
+def handler(event, context):
+    """
+    Entrypoint for lambda handler.
+
+    :param event: event for the lambda invoke to handle
+    :param context: context about this invoke
+    :return: result for the invoke
+    """
+    if 'index' in event:
```
neuromusic, line 39: I'm assuming that `event` is a string, is that right?

Is there a risk that there's an event called `plugin_version` where its ambiguous which if statement should take precedence?

neuromusic (approved):
looks good!

gave some feedback, since you asked for it 😄  

mostly stylistic. nothing blocking.

backend/lambda.py
codemonkey800: +1 to this, although it means the list will need to be updated in-place vs the more favorable immutable approach of creating a new list 😆

backend/lambda.py
```diff
@@ -0,0 +1,433 @@
...
+    """
+    query pypi to get all plugins.
+
+    :return: all plugin names and latest version
+    """
+    packages = {}
+    page = 1
+    name_pattern = re.compile('class="package-snippet__name">(.+)</span>')
+    version_pattern = re.compile(
+        'class="package-snippet__version">(.+)</span>')
```
codemonkey800, line 348: It's usually better to parse the HTML and use CSS selectors to query the elements: 
https://docs.python-guide.org/scenarios/scrape/
http://htmlparsing.com/regexes.html

There are some scenarios where using a parser is better. What if the pypi developers add a new attribute:

```html
<span class="package-snippet__version" data-package-version>0.1.0</span>
```

or a new class:

```html
<span class="package-snippet__version some-text-class">0.1.0</span>
```

Using a parser here wouldn't require code changes, but the regexes will. I don't think the HTML will change _that_ often, but definitely something to keep in mind 😆

codemonkey800 (commented):
Awesome work! Overall looking good to me. Left a few comments but nothing major 🔥

codemonkey800:
Also just a suggestion (maybe something for a future PR if you want), but it would be awesome to have type hints so it's easier to understand the type of the data being passed around

backend/lambda.py
neuromusic: copy & pop

backend/lambda.py
neuromusic: +1

backend/lambda.py
ziyangczi: In my initial version in the napari code I implemented parsing for the page number, but talley pushed a commit to use this approach, just doing the same thing for consistency

backend/lambda.py
ziyangczi: currently the pypi list doesn't have ordering by alphabetically, I don't think the ordering matters, the web front end can probably use whatever ordering makes the most sense to users later

backend/lambda.py
ziyangczi: event is a dictionary, we actually control the invoke part for the lambda, so there is no risk of an ambiguous event. However given a second thought, had I have a chance to clean it up I could have used value instead of key to control the flow. Gonna leave it here for now

backend/lambda.py
```diff
@@ -0,0 +1,433 @@
...
+def get_plugin(plugin, version=None):
+    """
+    get plugin metadata for a particular plugin, get latest if version is None.
+
+    :param plugin: name of the plugin to get
+    :param version: version of the plugin
+    :return: plugin metadata dictionary
+    """
+    if version is None:
+        url = f"https://pypi.org/pypi/{plugin}/json"
```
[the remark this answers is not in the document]
ziyangczi: that being said, you do have a 5min window cache from api gateway side

backend/lambda.py
codemonkey800: I'm not sure when we'll encounter this, but it'll be important to have the backend do the ordering later on since we'll need to handle pagination / infinite loading. Maybe if/when we move to the search to the backend, it'll make more sense to move ordering to the backend.

backend/lambda.py
ziyangczi: Agreed, leaving for now for code consistency, but I think we definitely want to transition to selectors in the future

ziyangczi:
thanks all for your review comments! I have addressed them and hopefully we can merge the file in for iterative improvements now :)

backend/lambda.py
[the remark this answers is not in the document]
codemonkey800: > I probably should advertise heavily then that front end should avoid as much as possible using unversioned call

This is a good callout, currently the implementation in #30 is making unversioned API calls:
https://github.com/chanzuckerberg/napari-hub/blob/307103777700ada3d0d6594f8ed7e70238291756/client/src/pages/plugins/%5Bname%5D.tsx#L54-L59

But then that would mean for the plugin details page, we would need to make an additional API call to `/plugins` to fetch the entire plugin list only to get the latest version for one plugin.

We can add this to the frontend, but I wonder if it would be better to do it in the backend by doing something like:

```py
    if version is None:
        plugins = get_plugins(context)
        version = plugins[plugin]
        
    if cache_available(f'cache/{plugin}/{version}.json', None):
        return get_cache(f'cache/{plugin}/{version}.json')
    url = f"https://pypi.org/pypi/{plugin}/{version}/json"
```

That way the latest plugin will use whatever is latest according to the index. Not sure if this works though, so would love to hear what you think

backend/lambda.py
codemonkey800: Of course! It works well right now so we can look into it if necessary. Hopefully pypi doesn't change their HTML 😆

justinelarsen (approved):

backend/lambda.py
```diff
@@ -0,0 +1,433 @@
...
+    Query pypi to get all plugins.
+
+    :return: all plugin names and latest version
+    """
+    packages = {}
+    page = 1
+    name_pattern = re.compile('class="package-snippet__name">(.+)</span>')
+    version_pattern = re.compile(
+        'class="package-snippet__version">(.+)</span>')
+    url = requote_uri(f"https://pypi.org/search/?c=Framework :: napari&page=")
```
neuromusic, line 349: comments by @justinelarsen & @codemonkey800 reminded me....

I'm not confident that the default "relevance" ordering on PyPI is even *stable*. There are multiple times that I've clicked to the second page and seen a plugin that I swear I had just seen on the prior page.

it might be better to use `https://pypi.org/search/?q=&o=-created&c=Framework+%3A%3A+napari` as the remote URI here. there's still a chance that a plugin gets updated while we're traversing the pages, but that should be rare

ziyangczi: I had the same doubt, so I did run loops to check for paging consistency last week, from my experiment I didn't observe any inconsistency in the ordering, without looking at their implementation, I would think if there is a problem in the default ordering, the problem would exist in the date last update ordering as well. If there are phantom entries on the second page, I don't think ordering would be the issue

PR #32, opened 2021-05-07
adds "file:" prefix for `long_description`

neuromusic:
fyi @DragaDoncila


neuromusic:
this fixes a type (see https://github.com/napari/napari/blob/master/setup.cfg#L8 for correct syntax)

justinelarsen (approved):

PR #36, opened 2021-05-10
Improve Dockerfile for production

codemonkey800:
## Description

This PR improves the frontend Dockerfile. It splits the image into two using [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/) so that build dependencies and source code aren't included in the final production image. It also switches the base image to an alpine based runtime.

This ultimately reduces the image size from 2.51 GB to 600 MB.

### DIFF
--- a/client/Dockerfile
+++ b/client/Dockerfile
@@ -1,6 +1,28 @@
-FROM node:15
-COPY . /app
+# Build image
+FROM node:16-alpine AS builder
+
 WORKDIR /app
-RUN yarn install && yarn build
+
+# Install dependencies
+COPY package.json yarn.lock postinstall.js /app/
+RUN yarn install
+
+# Build for production
+COPY . /app
+RUN yarn build
+
+# Application image
+FROM node:16-alpine
+
+WORKDIR /app
+
+# Install dependencies for production
+COPY package.json yarn.lock postinstall.js /app/
+ENV NODE_ENV=production
+RUN yarn install
+
+# Copy application build to image
+COPY --from=builder /app/.next /app/.next
+
 EXPOSE 8080
 ENTRYPOINT [ "yarn", "start" ]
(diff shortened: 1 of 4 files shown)


client/Dockerfile
```diff
@@ -1,6 +1,28 @@
...
 WORKDIR /app
-RUN yarn install && yarn build
+
+# Install dependencies for production
+COPY package.json yarn.lock postinstall.js /app/
+ENV NODE_ENV=production
+RUN yarn install
+
+# Copy application build to image
+COPY --from=builder /app/.next /app/.next
```
ziyangczi, line 25: I am not strictly against this approach, but logically would be better if we can still do one image but *remove* the unnecessary intermediate build files instead of copying the necessary files to a new image.

client/postinstall.js
```diff
@@ -0,0 +1,9 @@
+const { execSync } = require('child_process');
+const { resolve } = require('path');
+
+if (process.env.NODE_ENV !== 'production') {
```
ziyangczi, line 4: out of curiosity what is this part doing, trying to skip something unless run in github actions?

ziyangczi (approved):
Generally LGTM, some suggestions and question for you to consider

client/Dockerfile
codemonkey800: I'm curious if there are any drawbacks to this approach since multi-stage builds were created specifically to solve this problem 🤔  

Any suggestions? I think another way we could do it is have a big `RUN` layer that installs the dependencies, builds the application, and then removes the dependencies, but then it wouldn't be taking advantage of docker's layering.

client/postinstall.js
codemonkey800: `husky` is a development dependency that runs after running `yarn install`. Since it's not included in the final image, this script skips running `husky`

client/Dockerfile
ziyangczi: the only concern that I had was the complexity of the logic to build the container, that people new to the project may be confused as to why there are 2 identical base images used in the process. I have no problem with the approach here, for now, let's merge and maybe iterate on this later

PR #37, opened 2021-05-11
revisions based on UXR findings

neuromusic:
some revisions based on @LCObus 's UXR work & discussion amongst the team last week:

**What to revise from filter/sort?**

- License: Switch to “open source” boolean
- Author: Drop as filter and move to search
- Development Stage: switch to “stable” boolean
- Python versions: Switch to tags of python versions

**Default list order**

- Keep "most recent" for now. Future iteration consider other prioritization

**How to make sure that Project Site isn’t just the code repo?**

- Recognize that devs could use this field for either the homepage or the source code.
- Revised logic:
  - If the Python project_url field is a github or gitlab URL, then...
    - do not populate “project site” with this field
    - populate Source Code if not already specified

### DIFF
--- a/docs/customizing-plugin-listing.md
+++ b/docs/customizing-plugin-listing.md
@@ -78,7 +78,7 @@ For each of the fields in a plugin's listing, we outline below how the field is 
 | Support              |     ✅      |     ⛔      |     ⛔      |      ⛔       |    ⛔    |    <img src="https://github.com/favicon.ico" height="20"> (<img src="https://pypi.org/static/images/logo-small.svg" height="20">) |
 | Report Issues        |     ✅      |     ⛔      |     ⛔      |      ⛔       |    ⛔    |    <img src="https://github.com/favicon.ico" height="20"> (<img src="https://pypi.org/static/images/logo-small.svg" height="20">) |
 | Twitter              |     ✅      |     ⛔      |     ⛔      |      ⛔       |    ⛔    |    <img src="https://github.com/favicon.ico" height="20"> (<img src="https://pypi.org/static/images/logo-small.svg" height="20">) |
-| Source Code      |     ✅      |     ⛔      |     ⛔      |      ⛔       |    ⛔    |    <img src="https://pypi.org/static/images/logo-small.svg" height="20"> |
+| Source Code         |     ✅      |     ⛔      |     ⛔      |      ⛔       |    ⛔    |    <img src="https://pypi.org/static/images/logo-small.svg" height="20"> |
 | Release Date         |     ✅      |     ✅      |     ⛔      |      ✅       |    ⛔    |    <img src="https://pypi.org/static/images/logo-small.svg" height="20"> |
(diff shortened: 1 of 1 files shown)


justinelarsen (approved):

docs/customizing-plugin-listing.md
```diff
@@ -78,7 +78,7 @@ For each of the fields in a plugin's listing, we outline below how the field is
 | Support              |     ✅      |     ⛔      |     ⛔      |      ⛔       |    ⛔    |    <img src="https://github.com/favicon.ico" height="20"> (<img src="https://pypi.org/static/images/logo-small.svg" height="20">) |
 | Report Issues        |     ✅      |     ⛔      |     ⛔      |      ⛔       |    ⛔    |    <img src="https://github.com/favicon.ico" height="20"> (<img src="https://pypi.org/static/images/logo-small.svg" height="20">) |
 | Twitter              |     ✅      |     ⛔      |     ⛔      |      ⛔       |    ⛔    |    <img src="https://github.com/favicon.ico" height="20"> (<img src="https://pypi.org/static/images/logo-small.svg" height="20">) |
-| Code Repository      |     ✅      |     ⛔      |     ⛔      |      ⛔       |    ⛔    |    <img src="https://pypi.org/static/images/logo-small.svg" height="20"> |
+| Source Code         |     ✅      |     ⛔      |     ⛔      |      ⛔       |    ⛔    |    <img src="https://pypi.org/static/images/logo-small.svg" height="20"> |
```
justinelarsen, line 81:
```suggestion
| Source Code          |     ✅      |     ⛔      |     ⛔      |      ⛔       |    ⛔    |    <img src="https://pypi.org/static/images/logo-small.svg" height="20"> |
```

justinelarsen: Looks like I still missed a space! 😆  Difficult to get the alignment right

justinelarsen (approved):

codemonkey800 (approved):
Thanks for updating this, LGTM 😄

docs/customizing-plugin-listing.md
```diff
@@ -244,7 +248,9 @@ See the [Python Packaging User Guide](https://packaging.python.org/guides/distri
 
 This is the development status of your plugin.
 
-We display this on the detailed plugin page and the plugin listings. We support filtering plugins based on this value.
+We display this on the detailed plugin page and the plugin listings. 
+
+We support filtering plugins that are "stable" based on this value. Plugins that are labelled as "5" or "6" will be considered "stable".
```
kne42, line 253: Maybe this will be different because people will be uploading their individual plugins but I would like to note that it's not uncommon for stable projects to be marked as still being in alpha/beta. See for example how [scikit-image is still labeled as in beta](https://pypi.org/project/scikit-image/). Perhaps in our documentation for plugin developers, we should specify milestones at which it's appropriate to use each tag.

kne42 (approved):
LGTM. minor non-blocking comment

docs/customizing-plugin-listing.md
neuromusic: this is a really good point. this will only be of value if plugin devs use these development status classifiers to convey useful information to users.

PR #42, opened 2021-05-13
Minor client refactors

codemonkey800:
## Description

This PR has a few changes for the column layout, app bar, and plugin page. It includes refactors to the column layout for 2-columns, a few updates from Lia's design changes, and simplifies the styling for some components.

## Changes

- Updates tailwind config:
  - Removed `napari-3-col-fr` because [it already exists as `grid-cols-3`](https://tailwindcss.com/docs/grid-template-columns)
  - Changed the order of `napari-2-col` to be `225px 775px`
  - Renamed `napari-app-bar-2-col` to `napari-2-col-reverse` and changed order to `775px 225px`
- Updated ColumnLayout
  - Added `component` for changing the root component. This is so the AppBar can use `nav` as the root component
  - Added `reverseTwoColumn` to render the 2-column layout with `napari-2-col-reverse`
- Add / Update usages of ColumnLayout in AppBar and PluginDetails
- Lia's design updates
  - Move app bar links to the header 
  - Add CTA button to bottom of page

### App Bar Links

image

### 2nd CTA Button at Bottom of Plugin Page

image

image

### DIFF
--- a/client/src/components/AppBar/AppBar.module.scss
+++ b/client/src/components/AppBar/AppBar.module.scss
@@ -1,5 +1,0 @@
-.links {
-  :not(:first-child) {
-    @apply ml-6;
-  }
-}
--- a/client/src/components/AppBar/AppBar.module.scss.d.ts
+++ b/client/src/components/AppBar/AppBar.module.scss.d.ts
@@ -1,9 +1,0 @@
-export type Styles = {
-  links: string;
-};
-
-export type ClassNames = keyof Styles;
-
-declare const styles: Styles;
-
-export default styles;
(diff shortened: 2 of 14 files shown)


kne42 (approved):
LGTM but was wondering when we would need to refactor the whole 3/2 columns thing to 5/3 columns as per the design. I suppose since you have that nice column component it should be relatively easy to change down the line

codemonkey800:
> wondering when we would need to refactor the whole 3/2 columns thing to 5/3 columns as per the design

I was thinking about that too lol It's pretty tricky right now because some elements will span multiple columns. For example on the landing page designs, the divider for each plugin result spans two columns.

image

On the plugin page, the plugin details span 3 columns, but the bottom CTA button spans 1 column:

image

I can see this becoming super tricky to implement purely in CSS Grids. So far implementation wise it's been easier to have the 3 column layout, and then use sub-layouts for alignment in the center column.

---

It might be possible though using `span` in [`grid-column`](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-column) and [`row-column`](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-row), but we'll need to refactor the plugin details page and app bar

PR #47, opened 2021-05-17
add unit test for lambda

ziyangczi:
added unit test for lambda

the rest of the code in lambda is heavy IO, so adding mocks as unit tests for them doesn't make much sense

### DIFF
--- a/backend/_tests/test_lambda.py
+++ b/backend/_tests/test_lambda.py
@@ -1,0 +1,129 @@
+from unittest import mock
+import importlib
+import requests
+from requests.exceptions import HTTPError
+
+api = importlib.import_module("backend.lambda")
+
+
+class FakeResponse:
+    def __init__(self, *, data: str):
+        self.text = data
+        self.status_code = requests.codes.ok
+
+    @property
+    def status_code(self):
+        status_code = self._status_code
+        self.status_code = requests.codes.ok + 100
+        return status_code
+
+    @status_code.setter
+    def status_code(self, status_code):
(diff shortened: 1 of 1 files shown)


backend/_tests/test_lambda.py
```diff
@@ -0,0 +1,129 @@
+from unittest import mock
+import importlib
+import requests
+from requests.exceptions import HTTPError
+
+api = importlib.import_module("backend.lambda")
```
codemonkey800, line 6: Is there a reason why you're using `importlib.import_module()` instead of `import api from backend.lambda`?

codemonkey800 (approved):
LGTM! I did have one question about `importlib`

backend/_tests/test_lambda.py
```diff
@@ -0,0 +1,129 @@
...
+
+class FakeResponse:
+    def __init__(self, *, data: str):
+        self.text = data
+        self.status_code = requests.codes.ok
+
+    @property
+    def status_code(self):
+        status_code = self._status_code
+        self.status_code = requests.codes.ok + 100
```
justinelarsen, line 17: requests.codes.ok + 100? Interesting way of doing this

justinelarsen (approved):
Looks good to me too!

backend/_tests/test_lambda.py
ziyangczi: nice catch! lambda is a keyword in python, so you can't really import from a file or module named lambda, so I went around with the string import method

PR #49, opened 2021-05-19
Add new hub breakpoints

codemonkey800:
## Description

This PR adds the new breakpoints to the config. This also adds the new and existing breakpoints under new names using the name format `screen-{width in pixels}`.

This format is an improvement over the original format because:

1. **It's more scalable**. With the old format, adding a new breakpoint in between breakpoints meant the other breakpoints would have to be renamed in both the config and the code. Now it's just the screen and width number.
2. **It's more readable**. When looking at class names, it's easier to understand what screen size the breakpoint is for. Compare `screen-875:flex` to `xl:flex`. You would need to look at `breakpoints.js` to know what `xl` is if you didn't memorize it.

I've left the old breakpoint names so we can refactor existing usages in future PRs.

### DIFF
--- a/client/breakpoints.js
+++ b/client/breakpoints.js
@@ -6,6 +6,19 @@
 module.exports = {
   // Special screen size for fresnel when using `lessThan` queries.
   zero: 0,
+  'screen-300': 300,
+  'screen-375': 375,
+  'screen-495': 495,
+  'screen-550': 550,
+  'screen-560': 560,
+  'screen-600': 600,
+  'screen-655': 655,
+  'screen-725': 725,
+  'screen-875': 875,
+  'screen-1150': 1150,
+  'screen-1425': 1425,
+
+  // TODO Remove when breakpoint existing usages are renamed.
   xs: 300,
   sm: 375,
   md: 495,


codemonkey800:
@liaprins-czi I added you as a reviewer in case you want to take a look at the breakpoints and verify I got everything 😄

kne42:
Honestly idk about this one. While on one hand it's easier to add new breakpoints, on the other, it makes it significantly more difficult to adjust breakpoint values. For example, on napari.dev, we realized that one of the tablet breakpoints should be moved up to kick in at smaller widths since at some point the buttons started looking too wide.

If we were to rename these to be a bit less confusing, I would group them based upon what device we're targeting (at least for <1150px), e.g. `phone-sm`, `tablet-md`, etc. This naming would be based on how the breakpoints align with commonly used devices' screen sizes, e.g. an iPhone XR would fall under the `phone-md` size in portrait mode and `phone-xl` or `tablet-xs` in landscape mode. I don't think that being unable to see the pixel widths associated with each breakpoint is that big a deal, since you can always check it by hovering over the class name in your IDE.

codemonkey800:
Thank you for the feedback @kne42! This one definitely had me thinking 😆 Naming things really is the hardest to do in CS.

> it makes it significantly more difficult to adjust breakpoint values.

Thank you for calling this out. I did a comparison and it seems both approaches have costs. It depends on what kind of changes we're making:

## 1. Adding a new breakpoint for a specific component

### `screen-<size>`

This approach is better because all it takes is adding a new screen to the breakpoint config and adding it to the component.

Example:
- Commit adding breakpoint and usage: https://github.com/chanzuckerberg/napari-hub/commit/2a1102f552131b65f6bfb90809c4c6832445e662

### `<device>-<size>`

This approach isn't great because it may require us to shift existing breakpoints and also update previous usages to those shifted breakpoints.

Example:
- Commit to add breakpoint. Notice how we have to shift the other ones: https://github.com/chanzuckerberg/napari-hub/commit/5bb78afc2a81b8f78e707979daf65ea0f80e5607
- Commit to fix previous usages. After shifting, we now have need to update all the previous usages: https://github.com/chanzuckerberg/napari-hub/commit/b747004480e2d30d6be3c5e4cbc0d83e4d467a49

## 2. Modifying an existing breakpoint

### `screen-<size>`

This approach isn't great because you need to update all the usages to the different screen size, much like the other approach when adding a new breakpoint:

Example:
- Commit for modifying existing breakpoints to a new screen: https://github.com/chanzuckerberg/napari-hub/commit/3f37db36765b6833ab10f218b8bf9a1f6f8eac7a

### `<device>-<size>`

This approach is better because it's a one line change:

Example:
- https://github.com/chanzuckerberg/napari-hub/commit/22575058f087bc1e5f27136a8fc4ab674e9784a2

---

> I would group them based upon what device we're targeting 

I'm not too keen on naming things based on devices because [breakpoints should be based on content](https://medium.com/simple-human/stop-using-device-breakpoints-b11a87e2625c). Although I know there are device names in the Figma, the non-standard device breakpoints lead me to believe that it was designed based on the content.

>  I don't think that being unable to see the pixel widths associated with each breakpoint is that big a deal, since you can always check it by hovering over the class name in your IDE.

Editors without Tailwind plugins and PR reviewers won't be able to hover to see the breakpoint value. Although, we should probably encourage folks to use VSCode when contributing for the best dev experience.

codemonkey800:
Ultimately this appears to be a decision based on which approach will be doing more as we add more content to the website: will be adding new breakpoints, or will we be modifying existing breakpoints?

So far we've only added new breakpoints, so I've tried optimizing for the former. But based on what you're saying, we may have to do the latter. But we can't do both unfortunately I'm missing something 😢 

> For example, on napari.dev, we realized that one of the tablet breakpoints should be moved up to kick in at smaller widths since at some point the buttons started looking too wide.

Also keep in mind that with this approach, modifying a single breakpoint may have unintentional ripple effects. It may fix the buttons when it's looking too wide, but what about other components using the same breakpoint? What if it breaks another part of the UI? Then again, this isn't necessarily a bad thing either if those ripple effects are accounted for.

codemonkey800:
Also how would you would feel about staying consistent with [Tailwind's approach](https://tailwindcss.com/docs/font-size) and just incrementing the `xl` number? lol:

```js
module.exports = {
  // Special screen size for fresnel when using `lessThan` queries.
  zero: 0,
  xs: 300,
  sm: 375,
  md: 400,
  lg: 495,
  xl: 550,
  '2xl': 560,
  '3xl': 600,
  '4xl': 655,
  '5xl': 725,
  '6xl': 875,
  '7xl': 1150,
  '8xl': 1425,
};
```

kne42:
> Also how would you would feel about staying consistent with [Tailwind's approach](https://tailwindcss.com/docs/font-size) and just incrementing the `xl` number? lol:
> 
> ```js
> module.exports = {
>   // Special screen size for fresnel when using `lessThan` queries.
>   zero: 0,
>   xs: 300,
>   sm: 375,
>   md: 400,
>   lg: 495,
>   xl: 550,
>   '2xl': 560,
>   '3xl': 600,
>   '4xl': 655,
>   '5xl': 725,
>   '6xl': 875,
>   '7xl': 1150,
>   '8xl': 1425,
> };
> ```

Not a fan of this whatsoever; using an `{n}xl` naming convention with this many values is just an eyesore (a confusing one at that). I think that the name should actually be representative of how big something's going to be: `xl` to `2xl` is only a 10px difference and yet it might make you think it's 100+.

codemonkey800:
also @liaprins-czi it would be great to get your opinion on this:

> Ultimately this appears to be a decision based on which approach will be doing more as we add more content to the website: will be adding new breakpoints, or will we be modifying existing breakpoints?

kne42 (approved):
That article you linked about tying things to devices and the other arguments/examples you've made are very convincing, so I've changed my mind and am now in favor of your `screen-{width}` approach.

justinelarsen (approved):
I like the screen approach better than the device approach as well. Thanks for linking that article.

codemonkey800:
> That article you linked about tying things to devices and the other arguments/examples you've made are very convincing, so I've changed my mind and am now in favor of your `screen-{width}` approach.
                                                                                                                                                                                                                                
Thank you! If this bites us back somehow in the future, pls point all your fingers at me 👉 🧑‍💻 👈

liaprins-czi:
> I'm not too keen on naming things based on devices because breakpoints should be based on content. Although I know there are device names in the Figma, the non-standard device breakpoints lead me to believe that it was designed based on the content.

Yes I agree with this in terms of the fact that the breakpoints are determined by the content: I just show what the design would look like on three of the most-differentiated screen sizes when I am iterating / exploring design ideas, and then show them in those sizes for the final designs. 

However, if it easier for you all to have the naming conventions for the breakpoints be based on devices, that is fine with me and totally up to you! One thing to keep in mind is that the breakpoints that the landing page has that the plugin page does not have (currently [560px](https://docs.google.com/document/d/13yWxYvrdQY2I9KxBfPR5j13mHCzQeB3cwx8Uj9etf8s/edit#heading=h.tsph0qq3bhpj), [655px](https://docs.google.com/document/d/13yWxYvrdQY2I9KxBfPR5j13mHCzQeB3cwx8Uj9etf8s/edit#heading=h.2gu8bmttiq6s), and [725px](https://docs.google.com/document/d/13yWxYvrdQY2I9KxBfPR5j13mHCzQeB3cwx8Uj9etf8s/edit#heading=h.c1cz934n4jil)) were added solely to account for the way the headline + illustration + bulleted text responds to screen width changing. This means a couple of things:

- You could consider naming them for their function, i.e. `title-3-lines` for the [655px](https://docs.google.com/document/d/13yWxYvrdQY2I9KxBfPR5j13mHCzQeB3cwx8Uj9etf8s/edit#heading=h.2gu8bmttiq6s) breakpoint, since that is the point of that breakpoint existing: when the headline goes to 3 lines that is when the styles for that breakpoint should enact. Again, i have no skin in the game in terms of how the breakpoints are named since I will not be referencing them or using them like you, so this is just an idea, in case it is helpful!
- Very relatedly: 
>Ultimately this appears to be a decision based on which approach we will be doing more as we add more content to the website: will be adding new breakpoints, or will we be modifying existing breakpoints?
At this point *for v0* the designs should not be changing at the point of adding new breakpoints, but like Kira said there might be small changes especially / likely exclusively for those breakpoints that occur only on the landing page (currently [560px](https://docs.google.com/document/d/13yWxYvrdQY2I9KxBfPR5j13mHCzQeB3cwx8Uj9etf8s/edit#heading=h.tsph0qq3bhpj), [655px](https://docs.google.com/document/d/13yWxYvrdQY2I9KxBfPR5j13mHCzQeB3cwx8Uj9etf8s/edit#heading=h.2gu8bmttiq6s), and [725px](https://docs.google.com/document/d/13yWxYvrdQY2I9KxBfPR5j13mHCzQeB3cwx8Uj9etf8s/edit#heading=h.c1cz934n4jil)). These might change because they are set depending upon when certain functional thresholds are met (i.e. [when the number of lines per bullet point change](https://docs.google.com/document/d/13yWxYvrdQY2I9KxBfPR5j13mHCzQeB3cwx8Uj9etf8s/edit#heading=h.c1cz934n4jil)), and it might be that this is at a slightly different pixel-width in the HTML / the way it renders than it was on my Figma when I was doing the resizing there. 

One other thing: Lucy just [updated the text of the middle bullet](https://docs.google.com/document/d/1HIGr10UMVDN0XqXR6pDQIbrSwipgyBhQQYDotmyc6s8/edit?ts=60a597f5) to be shorter, now it is **Install into napari with confidence**.  I have updated the designs, and for now not changing the breakpoint widths! But if this makes it look really weird at certain screenwidth, we may want to adjust the point at which the (currently) [725px breakpoint](https://docs.google.com/document/d/13yWxYvrdQY2I9KxBfPR5j13mHCzQeB3cwx8Uj9etf8s/edit#heading=h.c1cz934n4jil) occurs. This will not mean adding any new breakpoints, though!!!

Hopefully that helps! Please let me know if any questions or concerns

(And also this is assuming the font we are using loads properly, if it falls back to a different font then this would technically happen at a different screen width, but that is fine! They will still be able to read it even in different font and different / slightly wonkier layout.)

codemonkey800:
> You could consider naming them for their function

I like this idea, I'm open to revisiting how we name breakpoints in the future when we're not feature churning. It would be great if we can establish a global set of breakpoints and then name them using the standard (xs, sm, md, etc.). And then breakpoints for specific features can be named based on function as you mentioned.

> there might be small changes

Then it'll definitely take a bit more tedious effort to rename everything if we, for example, change a breakpoint to `screen-725` to something like `screen-700`. Until we figure out a naming scheme, let's stick with this approach for now

kne42:
+1 to the whole idea of mixing standard naming with functional naming. I definitely found myself confused as to why we would have a breakpoint at e.g. 550px and then one only 10px away at 560px. Some comments could also help this in the meantime.

PR #50, opened 2021-05-19
Landing page part 1

codemonkey800:
## Description

Depends on #49

This PR is part 1 of 2 for the landing page. This includes miscellaneous additions and refactors accumulated while working on the landing page.

## Changes

- ColumnLayout classes API
  - Added a prop `classes` for overriding styles
  - Inspired by [Material UI's classes API](https://material-ui.com/customization/components/#overriding-styles-with-classes) 
- Replace remaining icons with React components
- Refactor AppBarHeader to AppBarLinks
  - This extracts it to a separate module so that the landing page app bar can use the same component

## Demos

Demos are included in #51

PR #51, opened 2021-05-19
Landing page part 2

codemonkey800:
## Description

Depends on #50

This PR is part 2 of 2 for the landing page. This includes the full implementation and updated tests.

## Changes

- Large variant of search bar for landing page
- Fix search bar behavior
  - Search only on form submission instead when the user types
  - Render search / close icon button
- New Components
  - Hub: SVG icon for napari hub
  - AppBarLanding: Landing page variant of the app bar
  - PluginSearch: The landing and search page component

## Demos

Demo Link: http://imaging-dev-napari-hub-449596503.us-west-2.elb.amazonaws.com/

### Plugin Search

https://user-images.githubusercontent.com/2176050/118753263-74e83400-b819-11eb-9794-942c25cf1f9c.mp4

### Plugin Search from Other Pages

https://user-images.githubusercontent.com/2176050/118753265-77e32480-b819-11eb-8f34-600bbce2f3ec.mp4

### Responsiveness

https://user-images.githubusercontent.com/2176050/118753268-79145180-b819-11eb-92a2-42ab22d2eb21.mp4

https://user-images.githubusercontent.com/2176050/118753259-731e7080-b819-11eb-9ffa-42137ca70077.mp4


codemonkey800:
@liaprins-czi I added you as a reviewer. It would be awesome if you could check out the demo link and give feedback here if any 😄

client/src/components/SearchBar/SearchBar.tsx
```diff
@@ -78,17 +117,33 @@ export function SearchBar() {
...
+        // whether the user clicked on the button or not.
+        type="button"
+      >
+        {/* Render close button if the user submitted a query. */}
+        {query ? (
+          <Close className={clsx(iconClassName, styles.closeIcon)} />
+        ) : (
+          <Search className={iconClassName} />
+        )}
+      </button>
```
ehoops, line 146: This button should have `aria-label` as well.

ehoops (commented):
Looks good.  Left some comments, mostly around accessibility.

justinelarsen:
General comment on the flow of the design: I like that when you search, the search bar moves to the top of the page with the results below. This happens when the list is normal size and I think this is how the design was intended - very cool! But, when there is only one result and the page is short, the search bar does not move to the top of the page and you can see some of the landing page banner content. I think this won't be an issue once we add the sidebar filters, but I think we should set a minimum height of the search results content area so that the search bar moves to the top of the page despite the number of search results. See screenshots with many results vs just one.
Screen Shot 2021-05-19 at 2 23 47 PM
Screen Shot 2021-05-19 at 2 23 33 PM

codemonkey800:
@ehoops thank you for all the accessibility feedback! I'll get these added as soon as I can

codemonkey800:
> when there is only one result and the page is short, the search bar does not move to the top of the page and you can see some of the landing page banner content. I think this won't be an issue once we add the sidebar filters, but I think we should set a minimum height of the search results content area so that the search bar moves to the top of the page despite the number of search results.

@justinelarsen agreed! i'll look into a fix for this

client/src/components/AppBar/AppBarLanding.tsx
```diff
@@ -0,0 +1,95 @@
...
+      {children}
+    </ul>
+  );
+}
+
+/**
+ * Component that renders the landing page variant of the AppBar.
+ */
+export function AppBarLanding() {
+  // On smaller layouts, the last list item is rendered below the icon and 2
```
justinelarsen, line 35: Good comment here, otherwise I would be asking _why_?

client/src/components/PluginSearch/PluginSearchBar.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { PLUGIN_SEARCH_ID } from './PluginSearch.constants';
+
+/**
+ * Component that renders the landing page search bar.
+ */
+export function PluginSearchBar() {
+  const activeQuery = useActiveQueryParameter();
+
+  // Scroll to search container when the search changes.
+  useEffect(() => {
```
justinelarsen, line 16: Love this effect!

client/src/components/PluginSearch/PluginSearchResult.tsx
```diff
@@ -0,0 +1,129 @@
...
+    {
+      label: 'release date',
+      value: formatDate(plugin.release_date),
+    },
+    {
+      label: 'license',
+      value: plugin.license,
+    },
+    {
+      label: 'Python version',
```
justinelarsen, line 58: Why is it that "Python" is the only label to be capitalized? Maybe a question for @liaprins-czi if it's part of the design.

justinelarsen (approved):
LGTM! +1 to all Erin's accessibility comments

justinelarsen:
Another general comment on the design. For long month names in the search results, it looks weird to me that the date wraps onto two lines. @liaprins-czi Should we abbreviate the month name instead so all months fit the same?
Screen Shot 2021-05-19 at 2 41 37 PM

client/src/components/AppBar/AppBarLanding.tsx
```diff
@@ -0,0 +1,95 @@
...
+      <h1
+        className={clsx(
+          'font-bold col-span-full',
+          '-mt-12 screen-1425:m-0',
+          'screen-1425:col-start-2 screen-1425:col-span-3',
+          styles.heading,
+        )}
+      >
+        Discover, install, and share napari plugins
+      </h1>
```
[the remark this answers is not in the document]
codemonkey800: Sounds good! I'll fix the `napari hub` heading in #50 and I'll change the `TODO search filters` to something smaller

client/src/components/AppBar/AppBarLanding.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+      <Hub
+        className={clsx(
+          'h-full',
+          'self-center',
+          'screen-1425:row-start-2 screen-1425:col-start-2',
+          styles.logo,
```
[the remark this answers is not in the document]
codemonkey800: It gets rendered as inline SVG, so I added an `aria-hidden="true"` to the `Hub.tsx` file instead based on this article:
https://css-tricks.com/accessible-svgs/#example-2-standalone-icon-decorative

liaprins-czi:
> Another general comment on the design. For long month names in the search results, it looks weird to me that the date wraps onto two lines. @liaprins-czi Should we abbreviate the month name instead so all months fit the same?
> Screen Shot 2021-05-19 at 2 41 37 PM

This is a good catch but it should not be happening (I tested it with September which is the longest month name); I think this happening because it is currently not using the correct font: it should be using [Barlow](https://fonts.google.com/specimen/Barlow?query=barlo). I called this out (the typeface) in the review I left for PR #51 yesterday, too

liaprins-czi (commented):
Overall this looks really great! 🥳 There are a few things that don't match up with the design that should be changed before launch, but it doesn't matter to me if that happens now or in the bug bash time period, but I will leave them here for reference:

- The typeface should be [Barlow](https://fonts.google.com/specimen/Barlow?query=barlo); for me in my browser it is rendering as a generic sans-ui (and in the Inspector shows that as first font in the stack right now)
- After I search for a term, then click the "X", the search term is cleared from teh search bar, but the plugin list is not reset, and the magnifier icon does not take the place fo the "X". I have to click the "X" a second time for these to happen.
- At screen sizes less than 875px, the bullets should be just 50px horizontal spacing from the right edge fo the illustration ([see breakpoint design example screen](https://www.figma.com/file/pDKU0CdSzpSWGaoXXQVJLz/v0-hub-final-designs?node-id=17%3A3397))
- "Browse plugins" and "Search for a plugin by keyword or author" should be the same font size
- Baseline of "Version" (first item in metadata list on right side of plugin cards) should be aligned with baseline of plugin name (in left side of plugin cards)

Let me know if you have any questions or concerns! Also happy to hop on a call anytime.

codemonkey800:
Thank you for the feedback @justinelarsen and @liaprins-czi!

> The typeface should be Barlow; for me in my browser it is rendering as a generic sans-ui (and in the Inspector shows that as first font in the stack right now)

This has been added in  #52! I've merged main so now we should have the Barlow font. There are performance considerations we need to think of because it's a heavy-ish font, but it shouldn't be a problem for us for v0.

I noticed that the text for the metadata was the wrong font size. Updating that + having the Barlow font now fixes the issue:

image

> At screen sizes less than 875px, the bullets should be just 50px horizontal spacing from the right edge fo the illustration (see breakpoint design example screen)

Needed to use `justify-self: end`. Fixed!

> "Browse plugins" and "Search for a plugin by keyword or author" should be the same font size

Used the wrong text class. Fixed!

> Baseline of "Version" (first item in metadata list on right side of plugin cards) should be aligned with baseline of plugin name (in left side of plugin cards)

Forgot to remove a CSS class when hitting that breakpoint. Fixed!

PR #52, opened 2021-05-19
Material UI setup

codemonkey800:
## Description

Research: [Material UI Integration](https://docs.google.com/document/d/1Ah8eWaXj16UWkc-GGfe9lo6j3HRcRVzjFaHpLYIu8VU/edit?usp=sharing)

This PR sets up Material UI in napari hub with SSR enabled. This library adds a lot of value to napari hub because:

1. It improves accessibility in napari hub and is easier to build an accessible site using MUI components.
2. It provides a library of components that are commonly used in napari hub designs
3. It'll allow us to eventually migrate over to the [Science Design System](https://github.com/chanzuckerberg/sci-components)

## Changes

- Moves colors to separate module so that it can be used by the client and in the Tailwind config
- Uses transform-imports plugin to [transpile Material UI imports](https://material-ui.com/guides/minimizing-bundle-size/#option-2)
- Server Side Rendering of Material UI components based on [official Material UI example](https://github.com/mui-org/material-ui/tree/7df94e4235fc6fc98ec7c13710be45bdbbd86dbe/examples/nextjs)
- Adds Barlow font


client/src/pages/_app.tsx
```diff
@@ -50,14 +52,39 @@ export default function App({ Component, pageProps }: AppProps) {
   let page: ReactNode = <Component {...pageProps} />;
   page = getLayout?.(page) ?? <Layout>{page}</Layout>;
 
+  useEffect(() => {
+    // Remove the server-side injected CSS.
+    const jssStyles = document.querySelector('#jss-server-side');
```
kne42, line 57: is this CSS added by material ui or something? I don't quite understand why you want to remove it

client/src/global.scss
```diff
@@ -34,6 +34,9 @@ body {
     should only happen in rare cases where the viewport is < 300px.
   */
   @apply overflow-x-auto;
+
+  // Apply Barlow font to entire site.
+  @apply font-barlow;
```
kne42, line 39: with this in place, will we still need to use `tailwind-typography`?

client/.babelrc.js
```diff
@@ -1,3 +1,45 @@
...
+function getMUIImportsPlugin(name) {
+  let path = '${member}';
+
+  // Prefix with `esm` path to use ESM version of MUI.
+  if (name === 'client') {
+    path = `esm/${path}`;
+  }
+
+  return [
+    'babel-plugin-transform-imports',
```
tihuan, line 19: I haven't tested MUI's claim yet, but they seem to say that treeshaking should work out of the box without using a import transform plugin!

https://material-ui.com/guides/minimizing-bundle-size/#when-and-how-to-use-tree-shaking

client/src/pages/_app.tsx
tihuan: Yes! It's added by the FE server when rendering MUI's style on the server side, and we're removing it on the client side, because the browser is gonna generate its own stylesheet once JavaScript is up and running 🏃 

https://material-ui.com/guides/server-rendering/

client/src/pages/_app.tsx
```diff
@@ -50,14 +52,39 @@ export default function App({ Component, pageProps }: AppProps) {
...
+          rel="stylesheet"
+        />
       </Head>
 
       <ReactQueryProvider dehydratedState={pageProps.dehydratedState}>
-        <MediaContextProvider>{page}</MediaContextProvider>
+        <MediaContextProvider>
+          <ThemeProvider theme={theme}>
+            <StylesProvider
+              // By default, Material UI will inject styles at the bottom of the
```
tihuan, line 77: awesome comment!

tihuan (approved):
Looks super awesome! I'm transitioning Aspen to Next.js now too, so your MUI SSR setup is gonna come in handy 😆 Thanks for sharing, Jeremy!!  🎉  🎉  🎉

tihuan:
Unrelated to MUI, but when Napari Hub needs to add HSTS and CSP for security before launch, cellxgene has code for reference! https://github.com/chanzuckerberg/corpora-data-portal/blob/main/frontend/next.config.js

client/.babelrc.js
codemonkey800: Ahh somehow I missed this and instead focused on the line below 😆 :

> Development bundles can contain the full library which can lead to slower startup times

I'll double check and see if there's a difference in startup time. If not, then we can remove the plugin.

codemonkey800:
> Unrelated to MUI, but when Napari Hub needs to add HSTS and CSP for security before launch, cellxgene has code for reference! https://github.com/chanzuckerberg/corpora-data-portal/blob/main/frontend/next.config.js

Thanks for letting us know Timmy! I haven't set up HSTS or CSP from scratch before, so this will be super helpful 😻

client/src/theme.ts
```diff
@@ -0,0 +1,20 @@
...
+      main: colors.primary,
+      light: colors.light,
+    },
+  },
+
+  typography: {
+    fontFamily: 'Barlow',
+    button: {
+      // Remove uppercase styling from buttons
+      textTransform: 'none',
```
justinelarsen, line 17: Is this where you could also remove the rounded corners of the buttons? I know in your research doc comparison examples, the Material UI buttons had rounded corners whereas the hub design does not.

client/tailwind.config.js
```diff
@@ -25,15 +26,21 @@ module.exports = {
   theme: {
```
justinelarsen, line 26: How does Tailwind work with Material UI? It seems you had to create a Material UI theme with "createMuiTheme" and configure the font and colors which looks redundant to the Tailwind theme here.

client/.babelrc.js
codemonkey800: Just checked and there's actually a significant difference in initial build time for a page in development. Without the plugin, it takes about 10s more to build a page 😱

client/src/theme.ts
codemonkey800: Yeah it's possible to extend the theme to disable the border radius. I'll add that to the theme

client/tailwind.config.js
codemonkey800: The Material UI theme is used for styling Material UI components internally while Tailwind's theme is for the Tailwind utility classes. While we can use Material UI's styling system, [it's not actually a requirement](https://material-ui.com/guides/interoperability/). To eliminate redundancy, I created a single theme module that Tailwind and Material UI can share to keep things consistent

client/src/global.scss
codemonkey800: Yes because we use `tailwind-typography` for adjusting the font sizes of Markdown content

client/.babelrc.js
tihuan: DANG!! Thanks for testing that out 😆 Didn't know it was such a huge difference. They weren't kidding when they said it's roughly 6x slower without proper import in dev mode 😂 I added transform import in Aspen now too! Thanks a lot 🚀

PR #55, opened 2021-05-24
refactor signup form & footer

kne42:
- refactors signup form to use Material UI for the button and email input
- adjusts the scaling factor to be mathematically computed (was `3.5vw`, is now equivalent to `3.67vw`)
- resolves conflicts between prettier and eslint by adding prettier as a plugin to it

Note: Material UI adds some default animations to its components. If they're unwanted, they can be removed later. The signup form also changes size <495px when an error message pops up so that the spacing looks less weird (video of both below)

https://user-images.githubusercontent.com/29165011/119401257-1cd58580-bc90-11eb-8a40-7d0f2655551f.mov


client/src/global.scss
```diff
@@ -16,7 +19,11 @@ html {
 
     https://blog.logrocket.com/the-elements-of-responsive-typography
   */
-  font-size: clamp(11px, 3.25vw, 100%);
+  font-size: clamp(
+    #{$min-font-size},
+    #{$min-font-size / $min-content * 100vw},
+    100%
+  );
```
codemonkey800, line 26: still seems like magic to me 😆 

would love to hav a comment to describe what this is doing

codemonkey800 (approved):
left one comment but it LGTM! awesome work 🍜

client/src/components/SignupForm/SignupForm.tsx
```diff
@@ -49,59 +50,67 @@ export function SignupForm({ onSubmit }: Props) {
...
           action={MC_URL.href}
           method="post"
           target="_blank"
+          className={clsx(
+            // grid
+            'grid grid-cols-2 screen-875:grid-cols-napari-3',
+            // spacing
+            'screen-495:gap-x-12',
+            // make room for error message in vertical layout
+            error ? 'gap-y-8' : 'gap-y-3.5',
```
justinelarsen, line 70: Nice!

justinelarsen (approved):
LGTM, Great work @kne42 !

PR #58, opened 2021-05-25
Use static values for Python Version and Operating System filters

codemonkey800:
## Description

This PR replaces the dynamically populated form state values with static values as discussed on the CZI slack. This is because napari supports a specific matrix of python versions and operating systems, so it's better to use our own values instead.

### DIFF
--- a/client/src/context/search/filter.types.ts
+++ b/client/src/context/search/filter.types.ts
@@ -13,9 +13,22 @@ export interface LicenseFormState {
 }
 
 /**
- * State used for holding checked state of multiple checkboxes.
+ * Form state for filtering on operating systems.
  */
-export type CheckboxFormState = Record<string, boolean>;
+export interface OperatingSystemState {
+  linux: boolean;
+  mac: boolean;
+  windows: boolean;
+}
+
+/**
+ * Form state for filtering on python versions.
+ */
+export interface PythonVersionsState {
+  3.7: boolean;
+  3.8: boolean;
+  3.9: boolean;
+}
 
 /**
  * Root state object for the filter form. Each state object is a
@@ -25,6 +38,6 @@ export type CheckboxFormState = Record<string, boolean>;
 export interface FilterFormState {
   developmentStatus: DevelopmentStatusFormState;
   license: LicenseFormState;
-  operatingSystems: CheckboxFormState;
-  pythonVersions: CheckboxFormState;
+  operatingSystems: OperatingSystemState;
+  pythonVersions: PythonVersionsState;
 }
--- a/client/src/context/search/filter.types.ts
+++ b/client/src/context/search/filter.types.ts
@@ -15,19 +15,10 @@ export interface LicenseFormState {
 /**
  * Form state for filtering on operating systems.
  */
-export interface OperatingSystemState {
+export interface OperatingSystemFormState {
   linux: boolean;
   mac: boolean;
   windows: boolean;
-}
-
-/**
- * Form state for filtering on python versions.
- */
-export interface PythonVersionsState {
-  3.7: boolean;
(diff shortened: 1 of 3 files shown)


justinelarsen:
@codemonkey800 I don't think the slack discussion is publicly available. Could you summarize it here?

client/src/context/search/filter.types.ts
```diff
@@ -25,6 +38,6 @@ export type CheckboxFormState = Record<string, boolean>;
 export interface FilterFormState {
   developmentStatus: DevelopmentStatusFormState;
   license: LicenseFormState;
-  operatingSystems: CheckboxFormState;
-  pythonVersions: CheckboxFormState;
+  operatingSystems: OperatingSystemState;
```
justinelarsen, line 41: Update name for consistency, above two include "FormState" whereas the new two are only "State".

client/src/context/search/filter.utils.ts
```diff
@@ -2,61 +2,35 @@ import { isEmpty, pickBy, reduce, set } from 'lodash';
...
       onlyStablePlugins: false,
     },
+
     license: {
       onlyOpenSourcePlugins: false,
     },
-    operatingSystems: getInitialOperatingSystemState(results),
-    pythonVersions: getInitialPythonVersionState(results),
+
+    operatingSystems: {
```
justinelarsen, line 23: Is there a way to store this data so that we're not repeating it multiple times? I'm thinking about the maintenance of this if we have to update like adding Python versions for example. It looks like it's repeated twice in this PR, here and also in client/src/context/search/filter.types.ts. Also, it may introduce bugs if there are discrepancies.

justinelarsen (changes requested):
I left some comments. Since the options are static, I'm wondering if we can extract them so that they are easier to make updates in the future. I saw that they were duplicated and may make it more difficult to maintain. Thanks for addressing!

client/src/context/search/filter.utils.ts
codemonkey800: Thankfully TypeScript would actually catch whatever discrepancies if we ever encounter any.

However, I agree it's better if we only have to update one object vs two. I updated the PR to store the versions in an array so we only have to update one place for a new Python version

client/src/context/search/filter.utils.ts
```diff
@@ -2,61 +2,36 @@ import { isEmpty, pickBy, reduce, set } from 'lodash';
...
-function getInitialOperatingSystemState(results: SearchResult[]) {
-  return results.reduce<CheckboxFormState>((state, { plugin }) => {
-    plugin.operating_system.forEach((os) => {
-      // eslint-disable-next-line no-param-reassign
-      state[formatOperatingSystem(os)] = false;
-    });
-    return state;
-  }, {});
-}
+const SUPPORTED_PYTHON_VERSIONS = ['3.7', '3.8', '3.9'];
```
kne42, line 7: since we were talking about how these should be tied to the python versions napari supports, maybe in the future this value can be pulled somehow from the napari pypi info so we don't need to update it manually. thoughts @justinelarsen @neuromusic?

codemonkey800: Was thinking the same thing too. Not sure about specific strategies, but maybe we can use the `python_requires` property on the napari repo: https://github.com/napari/napari/blob/46f295b599269bc79485d6d7febf1265a4fc7c65/setup.cfg#L38

Or maybe we can keep a JSON file with these versions somewhere on a GitHub repo that can be updated independently from the code.

justinelarsen: I agree, dynamic would be better. "python_requires" property sounds good to me. @neuromusic What do you think?

PR #68, opened 2021-06-03
add cicd pipeline changes

ziyangczi:
tweaked terraform configuration for cicd update
updated lambda to use cache for environment to speed up dev env loading


.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -90,9 +90,9 @@ module backend_lambda {
 
   environment = {
     "BUCKET" = local.data_bucket_name
-    "BUCKET_PATH" = local.custom_stack_name
```
mbarrien, line 93: BUCKET_PATH is important for the dev environments to not collide with one another, since they share a bucket. Even if they share the same file and directory structure now and in theory could share with one another, this is not guaranteed in the future especially if one branch is testing a new feature.

.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -38,18 +38,18 @@ locals {
...
 
   slack_url = try(local.secret["slack_url"], "")
+  zulip_credentials = try(local.secret["zulip_credentials"], "")
 
-  frontend_url = try(join("", ["https://", module.frontend_dns[0].dns_prefix, ".", local.external_dns]), var.frontend_url)
+  frontend_url = try(join("", ["https://", module.frontend_dns.dns_prefix, ".", local.external_dns]), var.frontend_url)
 }
 
 module frontend_dns {
-  count                 = var.require_okta ? 1 : 0
```
mbarrien, line 46: This is needed for the dev environments.

ziyangczi: may I ask why? we do not run okta in any of our env

.happy/terraform/modules/ecs-stack/main.tf
ziyangczi: I reinstated this because our dev env is having a cold start issue at the beginning, and it is confusing why the environment is not immediately usable due to data being reloaded. I think we may try something different later to separate dev environments better

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,33 @@
+name: Deploy Changes to Cluster
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - "v*"
```
codemonkey800, line 8: With this enabled, it seems like pushing to an old `v*` tag will cause a production build. This is fine though if we never cherry pick fixes into older releases in favor of updating the version. Maybe like `v1.0.0` to `v1.0.1` for fixes? 

Curious on your thoughts for this

.happy/config.json
```diff
@@ -10,6 +10,12 @@
             "secret_arn": "happy/env-dev-config",
             "terraform_directory": ".happy/terraform/envs/dev",
             "log_group_prefix": "/napari-hub/dev"
+        },
+        "staging": {
```
codemonkey800, line 14: do we also need a `prod` config?

codemonkey800 (commented):
Overall LGTM! Left a few comments and it would be nice to have Michael's comments resolved too before approving

.happy/config.json
ziyangczi: I would want to test everything on staging first before we move to the prod, probably will do in a second PR

codemonkey800: sounds good to me!

.github/workflows/deploy.yml
ziyangczi: true, another thing we can do is use a static `prod` tag to note where the deployment is.

I think in either case the problem is that some tags being pushed out with unintended effects on prod environment, whatever approach we have would not eliminate that. Another thing we can do is always manually update prod environment so we do not have cicd doing this for us

ziyangczi:
@mbarrien I do want to ask one question about load balancing on the dev environment where there are multiple target groups, it seems the alb rule condition is the same for all them: "IF Path is/*", I wonder what is the magic behind the scene so that our urls gets directed to their own environment, like test1.com goes to test1 containers, and test2.com goes to test2 containers?

.happy/terraform/modules/service/main.tf
```diff
@@ -49,6 +49,10 @@ resource aws_ecs_task_definition task_definition {
         "name": "API_URL",
         "value": "${var.api_url}"
       },
+      {
+        "name": "ENV",
```
ziyangczi, line 53: @codemonkey800 added ENV to the container in case you need it

mbarrien:
> @mbarrien I do want to ask one question about load balancing on the dev environment where there are multiple target groups, it seems the alb rule condition is the same for all them: "IF Path is/*", I wonder what is the magic behind the scene so that our urls gets directed to their own environment, like test1.com goes to test1 containers, and test2.com goes to test2 containers?
> 
> ps. nvm. I found the missing host match part

Ya it was the combination of host_match along with the special uri with variables in it for the aws_api_gateway_integration

.happy/terraform/modules/ecs-stack/main.tf
mbarrien: Can just leave the variable there, but pass an empty BUCKET_PATH, which should restore previous behavior.

.happy/terraform/modules/ecs-stack/main.tf
mbarrien: Right now the Oauth server is the only thing that is handling the wildcard DNS; we are not setting up separate DNS entries for every dev server. If that is the behavior we want, we'll have to modify the Terraform to create the Route 53 entries (and take the corresponding slow down in creation of a dev server, because Route 53 creation is slow), or we'll have to add a dummy nginx configured to directly proxy to the backends without going through oauth.

ziyangczi: I could be wrong but I think we are doing that as part of this PR, from what I saw the different stack url gets directed to the alb listener and forwarded to the container correctly. We only will be having one dev server so hopefully this wouldn't be an issue

.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -170,15 +171,3 @@ resource aws_iam_role_policy policy {
   role     = module.backend_lambda.role_name
   policy   = data.aws_iam_policy_document.backend_policy.json
 }
-
-resource aws_s3_bucket_object excluded_plugins {
```
ziyangczi, line 174: this will be managed by the team manually

.github/workflows/deploy.yml
codemonkey800: All good points, I'm actually fine with both manual and automatic prod deployments, but automated would be preferable. 

I think the approach you have now will be fine if we make sure we only push new tags and never to existing tags. So if we find a critical bug in `v1.0.0`, we create a new release with the bug fixed and tagging it `v1.0.1`, and we only keep incrementing from there. Tags are immutable anyway, so it makes sense to do it this way.

.happy/terraform/modules/service/main.tf
ziyangczi: yes

ziyangczi:
going to merge if no objection, I want to test our cicd out and make sure our staging is up to date with main branch

codemonkey800 (approved):
no objections to merging 💯

ziyangczi:
will address any other comment in subsequent PR, just need to test out the main branch ci

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,33 @@
...
+
+jobs:
+  prod:
+    name: deploy to production
+    runs-on: ubuntu-latest
+    if: github.repository == 'napari/napari' && startsWith(github.ref, 'refs/tags/v')
```
justinelarsen, line 14: Wrong repo name? Should be hub instead?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Checkout Repo
+        uses: actions/checkout@v2
+
+      - name: Deploy Changes
+        run: |
+          sudo scripts/happy --env prod update prod
+  staging:
+    name: deploy to staging
+    runs-on: ubuntu-latest
+    if: github.repository == 'napari/napari' && github.ref == 'refs/heads/main'
```
justinelarsen, line 25: Wrong repo name? Should be hub instead?

.github/workflows/deploy.yml
ziyangczi: corrected, nice catch

.github/workflows/deploy.yml
ziyangczi: corrected, nice catch

PR #69, opened 2021-06-04
Plugin Search styling fixes

codemonkey800:
## Description

This PR includes a few fixes for styling issues on the plugin search page.

## Changes

- Viewport clamping moved to its own function (thank you @kne42 for the initial code!)
- Fix issue with clamping on Safari: https://stackoverflow.com/a/65093383
- Add explicit height to logo to reduce unnecessary margins above and below the logo
- Fix list spacing for last item on landing page
- Fix grid layout weird grid layout glitch issue
- Adds plugin count to `Browse plugins` title

- Divider between filter and sort

## Demos

Demo Link: https://napari-hub-filter-ui-state-demo.vercel.app/

### Explicit Logo Height

The previous fix was adding `height: 100%`, but this has unintended issues on Safari.

#### Before

image

#### After

image

### Landing Page List Item Fix

The spacing for the last item was too long compared to the rest of the list items.

#### Before

image

#### After

image

#### Before

https://user-images.githubusercontent.com/2176050/120850221-fd5e1880-c52b-11eb-8b5f-099f7c9922c6.mp4

#### After

https://user-images.githubusercontent.com/2176050/120850224-fe8f4580-c52b-11eb-839a-4b5d5d35e3e4.mp4

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -7,7 +7,7 @@
     "dev": "nodemon -x 'run-p dev:*' -w tailwind.config.js",
     "dev:api": "node mock-server",
     "dev:next": "next dev -p 8080",
-    "dev:tsm": "tsm -w -e default 'src/**/*.module.scss'",
+    "dev:tsm": "tsm -w -e default -p.@ src 'src/**/*.module.scss'",
     "lint": "run-p lint:*",
     "lint:prettier": "prettier --check .",
     "lint:stylelint": "stylelint **/*.scss",
--- a/client/package.json
+++ b/client/package.json
@@ -104,6 +104,7 @@
     "playwright": "^1.10.0",
     "plop": "^2.7.4",
     "postcss": "^8.2.10",
+    "postcss-strip-units": "^2.0.1",
     "prettier": "^2.2.1",
     "sass": "^1.32.8",
     "stylelint": "^13.12.0",
--- a/client/src/global.scss
+++ b/client/src/global.scss
@@ -9,13 +9,12 @@ html {
   @apply overflow-y-scroll;
 
   /*
-    Use viewport width for fluid typography. This value is calculated based
-    on the ratio between the starting font size and the minimum content width to ensure proper scaling.
-    The minimum font size is from the design spec, and the maximum font size
-    is set to the user's preferred font size set by their browser settings.
+    Use viewport width for fluid typography. The minimum font size is from the
+    design spec, and the maximum font size is set to the user's preferred font
+    size set by their browser settings.
 
-    Texts, margins, and paddings should use the `text-<size>` Tailwind classes or rem / em units in
-    CSS to ensure the font size is more accessible.
+    Texts, margins, and paddings should use the `text-<size>` Tailwind classes
+    or rem / em units in CSS to ensure the font size is more accessible.
 
     https://blog.logrocket.com/the-elements-of-responsive-typography
   */
--- a/client/src/global.scss
+++ b/client/src/global.scss
@@ -1,5 +1,4 @@
-$min-font-size: 11px;
-$min-content: 300px; // smallest breakpoint in the theme
+@import '@/utils/styles';
 
 html {
   /*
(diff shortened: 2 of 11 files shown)


client/package.json
```diff
@@ -7,7 +7,7 @@
     "dev": "nodemon -x 'run-p dev:*' -w tailwind.config.js",
     "dev:api": "node mock-server",
     "dev:next": "next dev -p 8080",
-    "dev:tsm": "tsm -w -e default 'src/**/*.module.scss'",
+    "dev:tsm": "tsm -w -e default -p.@ src 'src/**/*.module.scss'",
```
codemonkey800, line 10: This lets us use `@/*` style imports in SCSS:
https://github.com/skovy/typed-scss-modules#--aliasprefixes--p

client/src/global.scss
```diff
@@ -10,21 +9,26 @@ html {
...
   */
-  font-size: clamp(
-    #{$min-font-size},
-    #{$min-font-size / $min-content * 100vw},
-    100%
-  );
+  font-size: viewport-clamp(11px, 100%);
+
+  /*
+    Awful hack to get clamp resizing working in Safari 14:
```
justinelarsen, line 24: Nice comments!

justinelarsen (approved):
LGTM, nice improvements!

client/src/utils/styles.scss
```diff
@@ -0,0 +1,23 @@
+/* stylelint-disable at-rule-no-unknown */
+
+$min-content: strip(theme('screens.screen-300'));
```
kne42, line 3: awesome! was also thinking that now that you have this fixed in PostCSS if it would be possible to detect the smallest breakpoint (besides 0) automatically

client/src/global.scss
```diff
@@ -1,5 +1,4 @@
-$min-font-size: 11px;
-$min-content: 300px; // smallest breakpoint in the theme
+@import '@/utils/styles';
 
 html {
```
kne42, line 3: just realized, shouldn't we be using `:root` here instead of `html`?

client/src/utils/styles.scss
```diff
@@ -0,0 +1,23 @@
...
+    #{$max-value}
+  );
+
+  @return $value;
+}
+/**
+ * Function for calculating the rem value for another value.
+ */
+@function calculate-rem($value) {
+  @return $value / 16px * 1rem;
```
kne42, line 22: nit: could you make `16px` a constant value with a name denoting that it's the default font size

also I think it might be slightly cleaner to write this as

```suggestion
  @return ($value / 16px) rem;
```

kne42 (approved):
awesome! LGTM!

codemonkey800:
ooh @kne42 my bad about merging it too soon. i'll address any changes in another PR 😅

client/src/utils/styles.scss
kne42: also the docstring should specify that this is for calculating pixels as rem, not any unit type

kne42:
no worries, it's my fault I took so long to review! I thought it was funny you merged at the same time I was reviewing LOL

client/src/utils/styles.scss
codemonkey800: that'd be awesome to have, but I'm not sure if a utility exists since it seems really specific lol

We could always look into making our own PostCSS plugin too 😄

client/src/global.scss
codemonkey800: We could, but there wouldn't be a difference since `:root` is referring to the top-most element which is `html` anyway. 

I mostly like using `:root` for declaring CSS variables. There are cases where you would want to use `:root`: https://stackoverflow.com/a/61409620

But I don't think those cases apply to us yet lol

kne42: Fair, it does have [higher specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/:root) though but I don't think that useful for us either.

client/src/utils/styles.scss
codemonkey800: But PEMDAS 😢  /s

Sounds good, I'll make a tiny PR to update this 😆

PR #72, opened 2021-06-07
Fetch GitHub Data

codemonkey800:
## Description

This PR sets up fetching stats for stars, forks, and issues + PRs on GitHub. GitHub data is fetched using the [@octokit/rest](https://www.npmjs.com/package/@octokit/rest) package.

### GitHub Stats

repo-stats

### GitHub API Errors

Any errors while accessing the GitHub API will render instead of the repo stats. 

<details>
<summary>Pervious error message</summary>

repo-error

</details>

image

## Follow ups

Implement stats for GitLab (and any other platforms we want to support in the future)


codemonkey800:
@liaprins-czi would be nice to get your feedback on the error message here and if there are better ways you'd want to display the error 😃

codemonkey800:
Need to defer merging this until we have an agreement with infra on how to setup authenticated access

client/src/components/PluginDetails/PluginMetadata.tsx
```diff
@@ -20,18 +20,19 @@ interface GithubMetadataItem {
...
-      count: 0,
+      count: repo.stars,
     },
     {
       title: 'Forks',
-      count: 0,
+      count: repo.forks,
     },
     {
       title: 'Issues + PRs',
```
justinelarsen, line 34: Issues and PRs have to be grouped together? This seems weird to me as they are quite different.

kne42 (approved):
LGTM!

One thing is I don't really like the error message provided, I don't think it looks very clean to users. I think it would look nicer if you showed some generic error message like "We're having trouble loading your GitHub stats" with a help icon next to it that provides the specific error and instructions on how to submit a bug report when hovered over.

client/src/components/PluginDetails/PluginMetadata.tsx
kne42: Not sure if there's a way to differentiate between them in the API, but for some reason GitHub seems to consider PRs to be issues

codemonkey800: I think it's because issues and PRs share the same number state on GitHub. If you create an issue and then a PR, the issue will be #1 and the PR will be #2. Also on the GitHub API, the `open_issues_count` value is a sum of the number of PRs and issues. There are ways to get them as separate counts though: https://stackoverflow.com/a/25069349

codemonkey800:
> One thing is I don't really like the error message provided, I don't think it looks very clean to users. I think it would look nicer if you showed some generic error message like "We're having trouble loading your GitHub stats" with a help icon next to it that provides the specific error and instructions on how to submit a bug report when hovered over.

Agree about the error message, it's really a placeholder for whatever we decide is a better way to show the error. I haven't added any UI for this because I wanted feedback on the UX

client/src/components/PluginDetails/PluginMetadata.tsx
liaprins-czi: > Issues and PRs have to be grouped together? This seems weird to me as they are quite different.

This is how they do it in PyPI, so it is the metadata @neuromusic  had outlined to include in the hub. If it is possible + easy to split them out (@codemonkey800 do you know how much extra effort this would be?), I am not sure if we'd just want to show issues then? Or would users also be interested in PRs sepearately?

liaprins-czi (commented):
I added a comment to this PR about combining a more generic message with the more specific error code (but copying it here for posterity): 
I like the idea of a more generic error message, but I think we can keep it pretty simple for this first version and say "We're having trouble loading the GitHub stats: [insert error code]" so the error code can be part of the message and not need to be a tooltip or anything (this way it will be more accessible for all device types, too)

client/src/components/PluginDetails/PluginMetadata.tsx
codemonkey800: It would be extra effort to get the counts separately because GitHub do not provide these numbers directly through their API, so we would need to calculate it ourselves. Users can always access the repo directly and look at the issue and PR count themselves.

Maybe it's something we can do in the future if users really want the counts separated?

codemonkey800: @justinelarsen Actually it's not as bad as I had imagined 😅 . I read somewhere that we would have to do some fancy pagination logic, but looking at the data, it would be simpler than that.

Here's a query to fetch all open PRs for [napari/napari](https://github.com/napari/napari):
https://api.github.com/search/issues?q=repo:napari/napari%20is:open%20type:pr&per_page=1

It looks like it includes a `total_count` value, so we can use that to subtract from the issues + PR count to get separated counts.

With that in mind,  should we update the UI to show something different? @liaprins-czi @justinelarsen

justinelarsen: Oh, nice! Good find 👍 

I'll leave the design decision up to @liaprins-czi but I think it makes sense to at least show the Issue count. I'm not sure how relevant the PR count is for the user.

PR #78, opened 2021-06-11
relax github url matching to also use other names than "source code"

ziyangczi:
relaxed the matching pattern for GitHub links to not only use "Souce Code" but also other links, in case people do not register GitHub link in that tag.

old behavior: https://staging.napari-hub.org/plugins/affinder
new behavior: https://zliu-frontend.dev.imaging.cziscience.com/plugins/affinder

### DIFF
--- a/backend/_tests/test_lambda.py
+++ b/backend/_tests/test_lambda.py
@@ -134,8 +134,11 @@ def test_github_get_url():
     plugins = {"info": {"project_urls": {"Source Code": "test1"}}}
     assert("test1" == get_download_url(plugins))
 
-    plugins = {"info": {"project_urls": {"Random": "https://"}}}
+    plugins = {"info": {"project_urls": {"Random": "https://random.com"}}}
     assert(get_download_url(plugins) is None)
 
-    plugins = {"info": {"project_urls": {"Random": "https://github.com/napari/napari/issues"}}}
-    assert("https://github.com/napari/napari" == get_download_url(plugins))
+    plugins = {"info": {"project_urls": {"Random": "https://github.com/org"}}}
+    assert(get_download_url(plugins) is None)
+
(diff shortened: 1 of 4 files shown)


backend/napari.py
```diff
@@ -160,6 +160,25 @@ def get_extra_metadata(download_url: str) -> dict:
...
+    """
+    project_urls = get_attribute(plugin, ["info", "project_urls"])
+    if project_urls:
+        source_code_url = get_attribute(project_urls, ["Source Code"])
+        if source_code_url:
+            return source_code_url
+        elif isinstance(project_urls, dict):
+            for key, url in project_urls.items():
+                if url.startswith("https://github.com"):
+                    return url
```
codemonkey800, line 178: not sure if this is overkill but is there a way you can modify the URL to link directly to the repo? I'm imagining a case where the plugin author submits a URL that is a subpath of the GitHub repo. For example if they provide Report Issues or Bug Tracker, it might link to `https://github.com/napari/napari/issues`

ziyangczi: addressed, see updated re matching

backend/_tests/test_lambda.py
```diff
@@ -98,7 +99,7 @@ def raise_for_status(self):
     'requests.get', return_value=FakeResponse(data=plugin_list)
 )
 def test_get_plugins(mock_get):
-    result = api.get_plugins(None)
```
ziyangczi, line 101: this was broken from earlier infra migration, fixed in this PR

codemonkey800 (approved):
awesome work!

neuromusic:
what's the logic if (say) the plugin dev links to one repo for the source code and to another for issues and a third for the project URL?

ziyangczi:
> what's the logic if (say) the plugin dev links to one repo for the source code and to another for issues and a third for the project URL?

we would use the "Source Code" link if there is one, and if there isn't, we will go through the list and try to find the first match

PR #80, opened 2021-06-15
add simple content pages

kne42:
add the about, privacy, and contact pages. note that the content is not final yet so they will need to be updated as needed

also adds a custom plugin to use variables in our links so that if we need to change something it only has to be done in the `constants.js` file

### DIFF
--- a/client/src/constants.js
+++ b/client/src/constants.js
@@ -1,7 +1,4 @@
-import { GitHub } from '@/components/common/icons';
-import { LinkInfo } from '@/types';
-
-export const LINKS: Record<string, LinkInfo> = {
+const LINKS = {
   HOME: {
     title: 'Home',
     link: '/',
@@ -26,7 +23,6 @@ export const LINKS: Record<string, LinkInfo> = {
     title: 'GitHub repo',
     link: 'https://github.com/chanzuckerberg/napari-hub',
     newTab: true,
-    icon: GitHub,
   },
   NAPARI_REPO: {
     title: 'GitHub repo',
@@ -35,7 +31,7 @@ export const LINKS: Record<string, LinkInfo> = {
   },
   NAPARI_GOVERNANCE: {
     title: 'napari governance model',
-    link: 'https://napari.org/docs/dev/developers/GOVERNANCE.html',
+    link: 'https://napari.org/community/governance.html',
     newTab: true,
   },
   IMAGESC: {
@@ -64,3 +60,5 @@ export const LINKS: Record<string, LinkInfo> = {
     newTab: true,
   },
 };
+
+module.exports = { LINKS };
--- a/client/src/pages/about.mdx
+++ b/client/src/pages/about.mdx
@@ -1,7 +1,5 @@
 import { LayoutMDX } from '@/components';
 
-<!-- TODO Create a layout for simple pages when designs are avaialble -->
-
 <LayoutMDX toc>
 
 ## About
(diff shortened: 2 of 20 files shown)


codemonkey800 (approved):
great work! the URL substitution in MDX is awesome 😄

client/src/constants.js
```diff
@@ -0,0 +1,64 @@
...
+    newTab: true,
+  },
+  IMAGESC: {
+    title: 'napari image.sc forum',
+    link: 'https://forum.image.sc/tag/napari',
+    newTab: true,
+  },
+  PLAUSIBLE: {
+    title: 'napari.dev analytics dashboard',
+    link: 'https://plausible.io/napari.dev',
```
justinelarsen, line 44: wrong plausible URL?

client/src/constants.js
```diff
@@ -0,0 +1,64 @@
+const LINKS = {
```
justinelarsen, line 1: I love having the link constants all in one place! ❤️

client/src/pages/about.mdx
```diff
@@ -1,43 +1,41 @@
...
+The CZI Imaging Team seeks to remove barriers in the analysis of imaging and microscopy data
+and make it easier for biologists to access emerging methods for bioimage analysis
+that leverage machine learning. CZI sees promise in napari and seeks to assist
+its development by providing resources not always available to the open source community,
+including dedicated user research, design, and engineering support.
+CZI is proud to collaborate with the science community to accelerate research
+and enable open science for all. In addition to napari and the napari hub,
+CZI imaging supports a number of grants on key areas such as visual proteomics,
+deep tissue imaging, and expanding global access.
+You can learn more about the imaging program and grant resources by visiting <https://czi.co/Imaging>.
```
justinelarsen, line 32: Since when do we have czi.co?? 😆  I didn't know this!

client/src/pages/contact.mdx
```diff
@@ -0,0 +1,17 @@
...
+
+For help with or ideas for the napari hub, please [check out the discussion board]({HUB_REPO}/discussions).
+
+For bugs or features requests, please [submit an issue on our GitHub repository]({HUB_REPO}/issues).
+
+For issues related to napari, please visit its [image.sc forum]({IMAGESC}) or [GitHub]({NAPARI_REPO}/issues) site.
+
+If you believe you have found a security issue with the napari hub, please responsibly disclose by contacting the CZI security team at <mailto:security@chanzuckerberg.com>.
+
+For all other inquiries, email the hub team at <mailto:tbd@tbd.tbd>.
```
justinelarsen, line 15: I think we know what email to use now - @neuromusic?

justinelarsen (commented):
@kne42 Awesome work! I left a couple of comments for updating some of the content, otherwise, it looks good!

client/src/constants.js
justinelarsen: After reading this, it looks like this is supposed to be for the napari.dev plausible site. Sorry! Can you rename this to be more specific in case we add the hub plausible too?

@neuromusic Do we have the intention of adding the plausible link to the hub plausible as well?

kne42: I don't think so I think it's just a placeholder from the privacy page

kne42:
@justinelarsen since your comments are content-related, I'm going to just merge this as-is so we have the infrastructure in place and then we can update the content once the copy is finalized

justinelarsen:
> @justinelarsen since your comments are content-related, I'm going to just merge this as-is so we have the infrastructure in place and then we can update the content once the copy is finalized

Great, thanks!

PR #83, opened 2021-06-16
add FAQ page

kne42:
again, we're still waiting on the copy to be updated for the content

Screen Shot 2021-06-16 at 9 21 27 AM

special thanks to @codemonkey800 for help debugging material ui!

### DIFF
--- a/client/src/components/common/Accordion/Accordian.module.scss
+++ b/client/src/components/common/Accordion/Accordian.module.scss
@@ -1,0 +1,18 @@
+.expandIcon {
+  &.expanded {
+    transform: rotate(45deg);
+  }
+}
+
+.expanded {
+}
+
+.accordion {
+  &.expanded::before {
+    opacity: 1;
+  }
+
+  &::before {
+    background-color: #000;
+  }
+}
--- a/client/src/components/common/Accordion/Accordian.module.scss
+++ b/client/src/components/common/Accordion/Accordian.module.scss
@@ -4,9 +4,6 @@
   }
 }
 
-.expanded {
-}
-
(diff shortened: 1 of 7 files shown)


client/src/components/common/Accordion/Accordian.module.scss
```diff
@@ -0,0 +1,18 @@
+.expandIcon {
+  &.expanded {
+    transform: rotate(45deg);
+  }
+}
+
+.expanded {
```
justinelarsen, line 7: Remove if not used?

justinelarsen (approved):
LGTM - nice work!

codemonkey800 (commented):
Overall LGTM, but it does look narrower than the designs. I believe it's because prose sets a max width. You can see it being removed here in the Markdown component:

https://github.com/chanzuckerberg/napari-hub/blob/c65f9741cf3118550f974a758d3047cbc8859907/client/src/components/common/Markdown/Markdown.tsx#L55-L63

client/src/components/common/Accordion/Accordian.module.scss
codemonkey800: The CSS class is actually required to get the cascading right when overriding Material UI styles 😮 

@kne42 It would be helpful to have comments here describing what the CSS does, but do let me know if you need help since we pair-programmed on this.

codemonkey800: Actually @kne42 you might be able to remove this class because you technically define it on lines `2` and `11`. Try remove this class and see if it still works correctly 😅

PR #85, opened 2021-06-16
Fix Plausible not working on staging / prod deployments

codemonkey800:
## Description

This should fix Plausible not working in staging / prod. The `ENV` variable is required at build time for the frontend, so we need to pass it in the deployment workflow.


ziyangczi:
May I ask why? the container that is build have these, curious why the building process need this

ziyangczi (approved):

PR #86, opened 2021-06-16
ensure python versions are sorted

0x00b1:
We may want to add a build step to automatically populate `SUPPORTED_PYTHON_VERSIONS`. 

We will need to provide a custom comparison function if we want to non-uniform version numbers (e.g. `3` and `3.9.5`).

### DIFF
--- a/client/src/context/search/filter.utils.ts
+++ b/client/src/context/search/filter.utils.ts
@@ -27,7 +27,7 @@ export function getDefaultState(): FilterFormState {
       windows: false,
     },
 
-    pythonVersions: SUPPORTED_PYTHON_VERSIONS.reduce(
+    pythonVersions: SUPPORTED_PYTHON_VERSIONS.sort().reduce(
       (state, version) => set(state, [version], false),
       {} as FilterFormState['pythonVersions'],
     ),


0x00b1:
Could someone link the issue? I want to reproduce to ensure the issue is actually fixed. 😆

codemonkey800:
@0x00b1 I think the order is messed up because we're using [`Object.entries()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries) in the React component. The array `SUPPORTED_PYTHON_VERSIONS` is technically already sorted, but the result of `Object.entries()` is not guaranteed:

https://github.com/chanzuckerberg/napari-hub/blob/75e867e90878e48cec4b2d0382cb8c9acfa76e17/client/src/components/PluginSearch/PluginFilterByForm.tsx#L67-L74

justinelarsen (approved):
Nice! Thanks for fixing :)

justinelarsen (changes requested):
@0x00b1  This PR needs some work. See @codemonkey800 's comment https://github.com/chanzuckerberg/napari-hub/pull/86#issuecomment-862754679

neuromusic:
hey @0x00b1 do you think you can land this or should @codemonkey800 or @kne42 pick it up?

klai95:
Previous Behavior:
The order of the array returned by Object.entries() does not depend on how the object is defined.
Current Behavior:
The ordering of the properties is the same as that given by looping over the property values of the object manually.

PR Updating the Docs:
https://github.com/mdn/content/pull/6937

PR #87, opened 2021-06-16
GitHub flavored Markdown

0x00b1:


codemonkey800 (commented):
`renderDescription()` is only used for converting the markdown to plain text so that the search engine doesn't index on markdown syntax.

The component responsible for actually rendering Markdown is `Markdown.tsx`, and it already has this plugin:

https://github.com/chanzuckerberg/napari-hub/blob/75e867e90878e48cec4b2d0382cb8c9acfa76e17/client/src/components/common/Markdown/Markdown.tsx#L23-L29

0x00b1:
Thank you @codemonkey800!

0x00b1:
@codemonkey800 Can I close this since we likely don't need it for search?

codemonkey800:
> @codemonkey800 Can I close this since we likely don't need it for search?

sounds good, I had a similar hunch but wanted to see if you could find a use case 😄

PR #88, opened 2021-06-16
Render newline delimited breaks properly

0x00b1:
Now we can
break without the spaces!
Nice job, Al!

PR #90, opened 2021-06-17
Fix missing support info

codemonkey800:
## Description 

Fixes the missing support info for screens < 875px. The issue only happens on initial load of the page, is due to the weird bug with placing adjacent `MediaFragment` components. Replacing them with normal `Media` components fixes the issue.

### Fixes
- https://airtable.com/shrw63y4RRus2DUB5/tblNxANme1LbVQqq1/viwISga6FERUatih1/recmUlaRvUVOa573z

## Demo

https://napari-hub-bug-bash-frontend.dev.imaging.cziscience.com/

image


justinelarsen:
@codemonkey800 The shields are now stacked vertically. I'm not sure if it's related. I think they're supposed to be aligned horizontally as they were before.
Screen Shot 2021-06-21 at 3 18 14 PM

liaprins-czi (approved):
This seems to fix the bug it says it does, on my browsers (Chrome + Safari)!

codemonkey800:
> @codemonkey800 The shields are now stacked vertically. I'm not sure if it's related. I think they're supposed to be aligned horizontally as they were before.
> Screen Shot 2021-06-21 at 3 18 14 PM

Just noticed that too but it's unrelated to this bug. It's because of unintended effects introduced in #88. It looks like the [remark-breaks](https://github.com/remarkjs/remark-breaks) package adds newlines at times when we don't want it to 😢

codemonkey800:
The newline issue has been fixed in #96

PR #91, opened 2021-06-18
Update search engine settings

codemonkey800:
## Description

Updates the search threshold and weightings for indexed data. 

### Fixes

- https://airtable.com/shrw63y4RRus2DUB5/tblNxANme1LbVQqq1/viwISga6FERUatih1/recTznzrow3ZX65Ns

### Demo

https://napari-hub-bug-bash-frontend.dev.imaging.cziscience.com/

### DIFF
--- a/client/src/components/PluginSearch/PluginSearch.test.tsx
+++ b/client/src/components/PluginSearch/PluginSearch.test.tsx
@@ -80,7 +80,7 @@ describe('<PluginSearch />', () => {
   });
 
   it('should support fuzzy matching', () => {
-    mockSearch('neuron');
+    mockSearch('neuro');
 
     const [author] = screen.getAllByTestId('searchResultSummary');
     expect(author.innerHTML.toLowerCase()).toContain('neuro');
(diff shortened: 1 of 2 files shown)


client/src/components/PluginSearch/PluginSearch.test.tsx
```diff
@@ -80,7 +80,7 @@ describe('<PluginSearch />', () => {
   });
 
   it('should support fuzzy matching', () => {
```
justinelarsen, line 82: Now that neuron isn't supported since we changed the threshold, can we still have a test for fuzzy search? Maybe something like searching for "animate" should return "animation"? It looks like one plugin was returned when I made that search using the new settings.

justinelarsen (approved):
LGTM! I made one comment on the fuzzy search test case.

client/src/components/PluginSearch/PluginSearch.test.tsx
codemonkey800: Sounds good, I added the test case you mentioned 👍

PR #96, opened 2021-06-22
Fix markdown line breaks

codemonkey800:
## Description

Fixes line breaks in Markdown introduced in #88. Issues with the previous approach:

1. [remark-breaks](https://github.com/remarkjs/remark-breaks) adds `<br>` tags automatically for newlines. This is correct for markdown for issues and PRs, **but not for READMEs**
2. The `remark-breaks` plugin didn't actually add a newline for `<br>` tags in READMEs

This fixes the above issues by using the [rehype-raw](https://github.com/rehypejs/rehype-raw) plugin for parsing HTML within markdown, and the [rehype-sanitize](https://github.com/rehypejs/rehype-sanitize) plugin to guard against XSS attacks.

### Before

https://staging.napari-hub.org/plugins/napari-aicsimageio

#### Line breaks on `<br>`

before

#### Badges

This was one of the unintended effects of the `remark-breaks` package.

image

### After

https://napari-hub-bug-bash-frontend.dev.imaging.cziscience.com/plugins/napari-aicsimageio

#### Line breaks on `<br>`

after

#### Badges

image

PR #98, opened 2021-06-23
fix footer spacing

kne42:
fix footer spacing on iOS devices since <14.6 does not support flex gap. we can revert to flex wrap + flex gap in maybe a year or so once 14.6 isn't so new

also fixes the footer's right margin size

includes some change to yarn.lock, not really sure why but git insisted on it

[relevant issue](https://airtable.com/tblNxANme1LbVQqq1/viwISga6FERUatih1/recJEqrx8ycySH9v5?blocks=hide)

https://user-images.githubusercontent.com/29165011/123176181-da8e9800-d437-11eb-8e0a-f57f4fdf4a6e.mov

### DIFF
--- a/client/yarn.lock
+++ b/client/yarn.lock
@@ -9481,13 +9481,6 @@ release-zalgo@^1.0.0:
   dependencies:
     es6-error "^4.0.1"
 
-remark-breaks@^2.0.2:
-  version "2.0.2"
-  resolved "https://registry.yarnpkg.com/remark-breaks/-/remark-breaks-2.0.2.tgz#55fdec6c7da84f659aa7fdb1aa95b632870cee8d"
-  integrity sha512-LsQnPPQ7Fzp9RTjj4IwdEmjPOr9bxe9zYKWhs9ZQOg9hMg8rOfeeqQ410cvVdIK87Famqza1CKRxNkepp2EvUA==
-  dependencies:
-    unist-util-visit "^2.0.0"
-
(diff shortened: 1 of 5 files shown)


liaprins-czi (commented):
This looks good to me. It looks like the logos have space to hop up to the same line as the links a little sooner (that is, a little narrower), but I assume you are instead re-using an existing breakpoint that is near to what these need, instead of adding a new one just for these?

justinelarsen (approved):
LGTM, thanks for the video!

kne42:
> I assume you are instead re-using an existing breakpoint that is near to what these need, instead of adding a new one just for these?

precisely! if you'd like to I'd be more than happy to add in another breakpoint at your discretion

liaprins-czi:
> precisely! if you'd like to I'd be more than happy to add in another breakpoint at your discretion

I don't think that is necessary, since this breakpoint is close enough, and since like you said this is likely a shortish term solution (until more ppl updgrade their OS)! 🙂

client/yarn.lock
```diff
@@ -9481,13 +9481,6 @@ release-zalgo@^1.0.0:
   dependencies:
     es6-error "^4.0.1"
 
-remark-breaks@^2.0.2:
```
codemonkey800, line 9484: thanks for removing this, I must have forgotten to run `yarn` when I worked on #96

codemonkey800 (approved):
looks good to me! 🍜

PR #106, opened 2021-06-24
make accordion divider not disappear on expansion

kne42:
previously, the divider of the next accordion would disappear when the previous accordion was expanded

special thanks to @codemonkey800 for deep diving into the material ui codebase because this was really unintuitive to solve :)

Screen Shot 2021-06-24 at 3 44 29 PM
Screen Shot 2021-06-24 at 3 44 38 PM


codemonkey800 (approved):
amazing 🍣

PR #107, opened 2021-06-25
Fix sort type switching to relevance on initial load

codemonkey800:
## Description

Fixes bug where the sort type switches to `relevance` on initial load. The issue is that we have some code that sets the sort type to `Relevance` when a the user enters a search query, but it doesn't take into account of the sort parameter being set on initial load. This can happen in three scenarios:

1. User searches for `video`, sets the sort type to `Plugin name`, and then opens the `napari-video` plugin page, then they navigate back to the search page. 
1. User modifies URL directly and sets `?sort=pluginName`.
1. User shares search URL with sort type set to `Plugin name`

In these cases, we want to respect the selected sort type on initial load. The only exception is if the URL has `?query` but not `?sort`, then we need to set the sort type to `Relevance`.

After initial load, all changes to `query` will now update the sort type.

## Fixes

- [Search criteria switches between page refreshes](https://airtable.com/shrw63y4RRus2DUB5/tblNxANme1LbVQqq1/viwISga6FERUatih1/recWIPLUyS9SF8geD)

## Demo

https://napari-hub-bug-bash-frontend.dev.imaging.cziscience.com

https://user-images.githubusercontent.com/2176050/123352033-54e01a80-d513-11eb-9339-00611f0c4c4f.mp4

### DIFF
--- a/client/src/context/search/search.hooks.test.ts
+++ b/client/src/context/search/search.hooks.test.ts
@@ -6,8 +6,14 @@ import { act } from 'react-dom/test-utils';
 
 import pluginIndex from '@/fixtures/index.json';
 
-import { useSearch } from './search.hooks';
+import {
+  DEFAULT_SORT_TYPE,
+  SearchQueryParams,
+  SearchSortType,
+} from './constants';
+import { useSearch, useSearchSetSortType } from './search.hooks';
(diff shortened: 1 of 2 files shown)


justinelarsen (approved):
LGTM I tested different scenarios and it seems to be working as expected.

PR #108, opened 2021-06-25
clarify sort labels

kne42:
First release -> Newest
Release date -> Recently updated

([relevant issue](https://airtable.com/tblNxANme1LbVQqq1/viwISga6FERUatih1/recd9jyNu4zagTzvf?blocks=hide))

### DIFF
--- a/client/src/components/PluginSearch/PluginSortByForm.tsx
+++ b/client/src/components/PluginSearch/PluginSortByForm.tsx
@@ -20,8 +20,8 @@ const DEFAULT_SORT_BY_RADIO_ORDER: SearchSortType[] = [
 
 const SORT_BY_LABELS: Record<SearchSortType, string> = {
   [SearchSortType.Relevance]: 'Relevance',
-  [SearchSortType.FirstReleased]: 'First released',
-  [SearchSortType.ReleaseDate]: 'Release date',
+  [SearchSortType.FirstReleased]: 'Created',
+  [SearchSortType.ReleaseDate]: 'Last updated',
   [SearchSortType.PluginName]: 'Plugin name',
 };
 
--- a/client/src/components/PluginSearch/PluginSortByForm.tsx
+++ b/client/src/components/PluginSearch/PluginSortByForm.tsx
@@ -20,8 +20,8 @@ const DEFAULT_SORT_BY_RADIO_ORDER: SearchSortType[] = [
 
 const SORT_BY_LABELS: Record<SearchSortType, string> = {
   [SearchSortType.Relevance]: 'Relevance',
-  [SearchSortType.FirstReleased]: 'Created',
-  [SearchSortType.ReleaseDate]: 'Last updated',
+  [SearchSortType.FirstReleased]: 'Newest',
+  [SearchSortType.ReleaseDate]: 'Recently updated',
   [SearchSortType.PluginName]: 'Plugin name',
 };
 
--- a/client/src/context/search/sorters.test.ts
+++ b/client/src/context/search/sorters.test.ts
@@ -1,0 +1,65 @@
+import { compareDates } from './sorters';
+
+function shuffle<T>(array: Array<T>): Array<T> {
+  // https://flaviocopes.com/how-to-shuffle-array-javascript/
+  const shuffled = array.slice();
+
+  do {
+    shuffled.sort(() => Math.random() - 0.5);
+  } while (shuffled === array);
+
+  return shuffled;
+}
+
+function sortNewToOld(dates: string[]): string[] {
+  return dates.slice().sort(compareDates);
+}
+
+describe('compareDates()', () => {
+  it('should return 0 when dateA and dateB are the same', () => {
+    expect(compareDates('1995-12-17 03:24:00', '1995-12-17T03:24:00')).toBe(0);
+  });
+
+  it('should return a positive number when dateA is newer than dateB', () => {
+    expect(
+      compareDates('2020-01-01 00:00:00', '2019-01-01 00:00:00'),
+    ).toBeGreaterThan(0);
+  });
+
+  it('should return a negative number when dateB is newer than dateA', () => {
+    expect(
+      compareDates('2020-01-01 00:00:00', '2020-01-01 00:00:01'),
+    ).toBeLessThan(0);
+  });
+
+  it('should sort dates from newest to oldest (trivial)', () => {
+    const datesSorted = [
+      '2019-01-01 00:00:00',
+      '2019-01-01 00:00:01',
+      '2019-01-01 00:01:00',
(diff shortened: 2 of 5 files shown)


codemonkey800 (approved):
LGTM! 🌮

client/src/components/PluginSearch/PluginSortByForm.tsx
```diff
@@ -20,8 +20,8 @@ const DEFAULT_SORT_BY_RADIO_ORDER: SearchSortType[] = [
 
 const SORT_BY_LABELS: Record<SearchSortType, string> = {
   [SearchSortType.Relevance]: 'Relevance',
-  [SearchSortType.FirstReleased]: 'First released',
-  [SearchSortType.ReleaseDate]: 'Release date',
+  [SearchSortType.FirstReleased]: 'Created',
```
neuromusic, line 23:
```suggestion
  [SearchSortType.FirstReleased]: 'Newest',
```

client/src/components/PluginSearch/PluginSortByForm.tsx
```diff
@@ -20,8 +20,8 @@ const DEFAULT_SORT_BY_RADIO_ORDER: SearchSortType[] = [
...
+  [SearchSortType.ReleaseDate]: 'Last updated',
```
neuromusic, line 24:
```suggestion
  [SearchSortType.ReleaseDate]: 'Recently updated',
```

neuromusic (changes requested):
the airtable notes included a request for different wording for the search filters:

> sort by "Recently updated" and "Newest".

also, can you confirm that this PR fixes the sort order to reflect that the top results are the most recently updated and newest for each of the selections?

client/src/components/PluginSearch/PluginSortByForm.tsx
neuromusic: b/c "Newest" makes it clear what the sort order is

codemonkey800:
> also, can you confirm that this PR fixes the sort order to reflect that the top results are the most recently updated and newest for each of the selections?

ah i forgot about this (although the airtable didn't say anything about it 🤔 ). @kne42 the implementations for the sorters are here:
https://github.com/chanzuckerberg/napari-hub/blob/912f18b1979911315521c3ad298ea4cc1fca84e2/client/src/context/search/sorters.ts#L17-L37

neuromusic:
>the airtable didn't say anything about it

sorry, I thought it was implied w/ the "Recently updated" and "Newest" labels

kne42:
@neuromusic Ah, apologies, it seems that the notes are hidden for me by default. Could you clarify how exactly the functionality needs to change? Also it wasn't clear to me that anything more than the text needed to change since it's labeled as a "Copy" issue.

codemonkey800:
> @neuromusic Ah, apologies, it seems that the notes are hidden for me by default. Could you clarify how exactly the functionality needs to change? Also it wasn't clear to me that anything more than the text needed to change since it's labeled as a "Copy" issue.

that's what I was thinking too, is there a separate issue regarding the sort? maybe we can have that implemented in a separate PR 🤔

neuromusic:
I'll let you two decide if you want to land the changes in one PR or two, but the functionality should be...

- sort by "Release date" becomes sort by "Recently updated". sorts by "release date" metadata, from high to low (newest to oldest)
- sort by "First release" becomes sort by "Newest". sorts by "First release" metadata, from high to low (newest to oldest)

kne42:
Ok, thanks for the clarifications @neuromusic! I think this change is simple enough that I can add it on to this PR. I'll include some tests as well to verify that things behave as we expect.

kne42:
Ok, I've added tests that verify the search results are being ordered from newest to oldest. You can take a look at the tests I wrote to double check that this is the behavior we expect.

client/src/context/search/sorters.test.ts
```diff
@@ -0,0 +1,71 @@
...
+function shuffle<T>(array: Array<T>): Array<T> {
+  // https://flaviocopes.com/how-to-shuffle-array-javascript/
+  const shuffled = array.slice();
+
+  do {
+    shuffled.sort(() => Math.random() - 0.5);
+  } while (shuffled === array);
+
+  return shuffled;
+}
```
codemonkey800, line 12: nit: this is a really cool implementation, but I would prefer if we use `lodash` functions whenever possible. It looks like there's a [shuffle()](https://lodash.com/docs/4.17.15#shuffle) function that does the same thing.

client/src/context/search/sorters.ts
```diff
@@ -3,14 +3,16 @@ import { SearchResult } from './search.types';
...
+ *
+ * See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#description
  *
  * @param dateA The first date
  * @param dateB The second date
- * @returns The difference between the dates in milliseconds
+ * @returns >0 if dateA is newer than dateB, <0 if dateB is newer than dateA, and 0 if equal.
  */
-function compareDates(dateA: string, dateB: string) {
+export function compareDates(dateA: string, dateB: string): number {
```
codemonkey800, line 14: This is intentionally an internal function used primarily by the sorters (which are also internal functions). Generally I've been testing the things that are actually exported (aka the public API). In this case, the only export for this module is `sortResults()`, so I expect the tests in `sorters.test.ts` to test that function.

I prefer testing only publicly exported functions, but I realize it's a pretty polarizing topic:
- https://philipwalton.com/articles/why-i-test-private-functions-in-javascript/
- https://kentcdodds.com/blog/testing-implementation-details
- https://stackoverflow.com/questions/105007/should-i-test-private-methods-or-only-public-ones

Ultimately, I'll leave the decision up to you on how you want test this, but it would be nice to have tests for the `sortResults()` plugin.

client/src/context/search/sorters.ts
```diff
@@ -3,14 +3,16 @@ import { SearchResult } from './search.types';
...
+  // time in ms makes newer dates have higher values
   return new Date(dateA).getTime() - new Date(dateB).getTime();
```
codemonkey800, line 16: You might need to reverse this calculation? I checked out your branch and I see the older dates still showing up on the top:

image

kne42:
@codemonkey800 and i agreed to add more tests for each sort function later, but it seems to be working now

client/src/context/search/sorters.ts
codemonkey800: For everyone else's context, we agreed to leave the test for the internal implementation and to add tests for `sortResults()` in the future

codemonkey800 (approved):

neuromusic (approved):
LGTM

PR #112, opened 2021-06-25
center signup form on home page

kne42:
([relevant issue](https://airtable.com/tblNxANme1LbVQqq1/viwISga6FERUatih1/recj5Qm8KyA5hCXvv?blocks=hide))

Screen Shot 2021-06-25 at 1 58 25 PM

### DIFF
--- a/client/src/components/SignupForm/SignupForm.tsx
+++ b/client/src/components/SignupForm/SignupForm.tsx
@@ -14,16 +14,18 @@ MC_URL.search = new URLSearchParams({
 
 interface Props {
   onSubmit?: (event: React.FormEvent) => void;
+  variant?: 'default' | 'home';
 }
 
 /**
  * Signup form to capture email address and send it to MailChimp
  */
-export function SignupForm({ onSubmit }: Props) {
+export function SignupForm({ onSubmit, variant = 'default' }: Props) {
   const [email, setEmail] = useState('');
   const [error, setError] = useState('');
   const emailRef = useRef<HTMLInputElement | null>(null);
   const plausible = usePlausible();
+  const isHome = variant === 'home';
 
   const validate = () => {
     const validityState = emailRef.current?.validity;
@@ -53,7 +55,11 @@ export function SignupForm({ onSubmit }: Props) {
   };
 
   return (
-    <ColumnLayout className="bg-napari-light p-6 screen-495:p-12">
+    <ColumnLayout
+      className="bg-napari-light p-6 screen-495:p-12"
+      // Use 3-column layout instead of 4-column layout.
+      classes={isHome ? { fourColumn: 'screen-1150:grid-cols-napari-3' } : {}}
+    >
       <div className="col-span-2 screen-495:col-span-3 screen-1425:col-start-2">
         <h3 className="text-lg font-semibold mb-1 screen-495:mb-4">
           Sign up to receive updates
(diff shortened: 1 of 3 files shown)


codemonkey800 (approved):
LGTM 🥔

client/src/components/SignupForm/SignupForm.tsx
```diff
@@ -53,7 +55,11 @@ export function SignupForm({ onSubmit }: Props) {
   };
 
   return (
-    <ColumnLayout className="bg-napari-light p-6 screen-495:p-12">
+    <ColumnLayout
+      className="bg-napari-light p-6 screen-495:p-12"
+      // Use 3-column layout instead of 4-column layout.
+      classes={isHome ? { fourColumn: 'screen-1150:grid-cols-napari-3' } : {}}
```
codemonkey800, line 61: unrelated to the PR, but how do you feel about the ColumnLayout API? i feel iffy about it, but I haven't thought about cleaner looking override it 😆

kne42: it's a little bit confusing having both `className` and `classes` as options. at minimum I would rename `classes` to `overrides` or something like that

I also think the component may be too generic with this and would probably be fine with a `variant` version instead

client/src/components/SignupForm/SignupForm.tsx
```diff
@@ -53,7 +55,11 @@ export function SignupForm({ onSubmit }: Props) {
...
+    >
```
kne42, line 62: in terms of the actual API though I think the following appeals to me more:
```suggestion
<ColumnLayout className={clsx('bg-napari-light p-6 screen-495:p12', isHome && '~screen-1150:grid-cols-napari-4 screen-1150:grid-cols-napari-3')} />
```

basically `~` before a class name signifies that it should be removed

client/src/components/SignupForm/SignupForm.tsx
justinelarsen: I agree with @kne42 - having both with close names is confusing.

justinelarsen (approved):
Code looks good, but difficult to test without a UI preview link

client/src/components/SignupForm/SignupForm.tsx
codemonkey800: It's meant to be consistent with how Material UI handles overrides: https://material-ui.com/customization/components/

We can change it for this component, but the `classes` and `className` API isn't going away completely (at least for Material UI components)

kne42:
@justinelarsen you can now test it out at this url: https://kira-sandbox-frontend.dev.imaging.cziscience.com

client/src/components/SignupForm/SignupForm.tsx
codemonkey800: I can see this working, although if we're going to use a special operators, we might as well use the important operator `!`:

```tsx
<ColumnLayout className={clsx(
  'bg-napari-light p-6 screen-495:p12',
  isHome && 'screen-1150:!grid-cols-napari-3'
)} />
```

But using `!important` is generally not recommended. Either way, this is good feedback and is something we can possibly look into during the quality sprint.

justinelarsen:
Thanks @kne42 ! LGTM!

liaprins-czi (approved):
Looks great!

PR #118, opened 2021-06-29
Replace install plugin images

codemonkey800:
## Description

Closes #114 

## Demo

https://fix-plugin-installer-image-frontend.dev.imaging.cziscience.com/plugins/napari-btrack-reader

image

PR #121, opened 2021-06-29
Refactor install modal to Material UI

codemonkey800:
## Description

I had some time to refactor the install modal to use the [Material UI Dialog component](https://material-ui.com/components/dialogs/#dialog) after working on #118.  Using the Dialog component also fixes a couple of issues with the current implementation:

1. Scrolling is no longer possible when the modal is open
2. Fixes scrollbar being visible on certain screens

## Demos

https://refactor-install-modal-to-mui-frontend.dev.imaging.cziscience.com/plugins/napari-btrack-reader

### Install Modal Responsiveness

https://user-images.githubusercontent.com/2176050/123724463-3cd50780-d841-11eb-8f27-ebd04cb17c4d.mp4

#### Before

https://user-images.githubusercontent.com/2176050/123724451-38105380-d841-11eb-8f60-92ad7c5160b0.mp4

#### After

https://user-images.githubusercontent.com/2176050/123724444-32b30900-d841-11eb-93c1-4672ed0a6756.mp4

### DIFF
--- a/client/src/components/PluginDetails/InstallModal.module.scss
+++ b/client/src/components/PluginDetails/InstallModal.module.scss
@@ -1,6 +1,17 @@
+@import '@/utils/styles';
+
 // Make SVG stroke color black
 .closeIcon {
   path {
     @apply stroke-current text-black;
   }
+}
+
+.modal {
+  $min-width: 270px;
+  $max-width: 775px;
+
+  max-width: $max-width;
+  min-width: $min-width;
+  width: viewport-clamp($min-width, $max-width);
 }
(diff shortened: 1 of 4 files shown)


codemonkey800:
@liaprins-czi since we're changing the underlying component, can you check for any design inconsistencies please? 😄

liaprins-czi (approved):
The only things I noticed (and I think this was the case before this PR but Ididn't notice) is that:
- The images, and also multi-line text (on smaller screens, when a line of text overflows to the next line) should be indented / aligned with the text above it (instead of the numeric bullet) so that the numeric bullet is sticking out (this makes it easier for users to scan from one step to the next)
- It looks like the line height is a little too high? It should be 1.5

However, these things are not important to be fixed for this PR adn can be added to low priority with the bug bash items!

kne42 (approved):
LGTM! 🎉

justinelarsen (approved):
LGTM!

PR #123, opened 2021-06-29
Right nav stickiness

codemonkey800:
## Description

Fixes the stickiness for the right nav in the plugin and markdown pages so that the right nav sticks only when its `50px` from the top of the screen after scrolling.

### Plugin Page

https://right-nav-stickiness-frontend.dev.imaging.cziscience.com/plugins/napari-btrack-reader

https://user-images.githubusercontent.com/2176050/123850261-a64f2780-d8ce-11eb-9244-c44dcee5bae2.mp4

### Markdown Pages

https://right-nav-stickiness-frontend.dev.imaging.cziscience.com/about

https://user-images.githubusercontent.com/2176050/123850254-a3543700-d8ce-11eb-9209-086edf7f5a9f.mp4

### DIFF
--- a/client/src/components/PluginDetails/PluginDetails.tsx
+++ b/client/src/components/PluginDetails/PluginDetails.tsx
@@ -87,8 +87,8 @@ function PluginRightColumn() {
 
   return (
     <Media greaterThanOrEqual="2xl">
+      {/*  Keep CTA button and TOC on screen when scrolling on 2xl. */}
       <div className="sticky top-12">
-        {/*  Keep button on screen when scrolling on 2xl. */}
         <CallToActionButton />
 
         <Markdown.TOC
--- a/client/src/components/PluginDetails/PluginDetails.tsx
+++ b/client/src/components/PluginDetails/PluginDetails.tsx
@@ -87,19 +87,22 @@ function PluginRightColumn() {
 
   return (
     <Media greaterThanOrEqual="2xl">
-      {/*  Keep button on screen when scrolling on 2xl. */}
-      <CallToActionButton className="fixed" />
+      <div className="sticky top-12">
+        {/*  Keep button on screen when scrolling on 2xl. */}
+        <CallToActionButton />
 
-      <Markdown.TOC
-        className="mt-24"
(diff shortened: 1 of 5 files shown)


client/src/global.scss
```diff
@@ -1,13 +1,6 @@
 @import '@/utils/styles';
 
 html {
-  /*
-    Enable scrollbar for the entire page by default. This is to prevent
-    layout shift when switching between pages that have content long enough /
-    not long enough for a scrollbar.
-  */
-  @apply overflow-y-scroll;
```
codemonkey800, line 9: Removing this is required for `position: sticky` to work. I think all of our pages have enough content to be scrollable, so I don't think we have to worry about this

liaprins-czi (approved):
Awesome work! Looks great to me!

client/src/components/PluginDetails/PluginDetails.tsx
```diff
@@ -87,19 +87,22 @@ function PluginRightColumn() {
...
-        markdown={plugin.description}
-        onClick={(section) => {
-          plausible('Description Nav', {
-            section,
-            plugin: plugin.name,
-          });
-        }}
-      />
+      <div className="sticky top-12">
+        {/*  Keep button on screen when scrolling on 2xl. */}
```
justinelarsen, line 91: Update comment to keep TOC sticky instead?

client/src/global.scss
justinelarsen: What about the Contact page? That one is pretty short.

justinelarsen (commented):
Awesome work! I just added a couple comments.

client/src/global.scss
codemonkey800: Good point, the page isn't scrollable as you mention. Still, I don't think we have to worry too much about. It's a super minor thing that only a handful of people will experience, and even then, they might not even notice it that much.

justinelarsen: Sounds good. Let's keep it in mind.

justinelarsen (approved):

PR #144, opened 2021-07-08
Local/remote dev setup

ziyangczi:
Documentation taken from michael's branch
adjusted the workflow to the team discussion:
- continue with current deployment flow and maintain staging - main branch, prod - 2 weeks behind staging
- add remote dev deployment with prefix dev-


.github/workflows/push-rdev.yml
```diff
@@ -0,0 +1,69 @@
+name: Push Remote Dev
```
ziyangczi, line 1: This workflow enables automatic stack creation from branch dev-*, which is only available to CZI employees since we don't give write permission to others in repo

scripts/happy-deploy.py
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env python3
```
codemonkey800, line 1: what is this script for? is it for starting a deployment workflow on GitHub?

it would be nice to have some documentation about this, but it can be in another PR if you're working on docs in another branch 😄

.github/workflows/push-rdev.yml
```diff
@@ -0,0 +1,69 @@
...
+
+on:
+  push:
+    branches:
+      - 'dev-*'
```
codemonkey800, line 6: it would be nice if we could have some kind of cleanup when the branch gets deleted. There is a way to listen for [delete events](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#delete).

codemonkey800 (approved):
LGTM 🍜 

I left a few comments but they don't have to be addressed in this PR. I'm just excited to get the dev branches 😄

.github/workflows/push-rdev.yml
ziyangczi: added but untested, will see how it works

scripts/happy-deploy.py
ziyangczi: this is what infra is pushing for future deployment flows, just going to leave it in here until we want to use this

ziyangczi: just to be clear it is not being referenced right now

PR #202, opened 2021-07-14
update value prop bullet

neuromusic:
new text: "Understand how to install into napari"


sofroniewn (approved):
Thanks for these changes, they look good to me!!

LCObus:
Understand isn't the most active verb (and i'm splitting hairs but it implies the user will understand, which we can't guarantee)
Could I suggest a more active verb of "**learn how to install into napari**"?

neuromusic:
thoughts on @LCObus's rec, @sofroniewn?

sofroniewn:
> "learn how to install into napari"?

also fine with me

sofroniewn:
> i'm splitting hairs but it implies the user will understand, which we can't guarantee

love the attention to detail 🚀

PR #203, opened 2021-07-14
add cert dependency for validation so adding cert to lb won't fail

ziyangczi:
This is to fix certificate linkage failure before the cert is validated that Jeremy was seeing on dev environments occasionally

There was a race condition where the cert was not validated before trying to add to load balancer when aws would complain about such addition


codemonkey800 (approved):
wonderful 🍜

PR #205, opened 2021-07-14
Render markdown on backend

codemonkey800:
## Description

This PR moves the description rendering functionality to the backend to [speed up initial load on the frontend](https://docs.google.com/document/d/1ykAnPTPLWf5Hxfisnlg8eehyTpb0xY9BHqaAQ5tFWeo/edit?usp=sharing). 

This works by rendering the Markdown to HTML, and then using an HTML parser to extract only the textual content from the HTML.

Moving rendering to the backend improves SSR response time from ~1203 millisecondsto ~608 milliseconds, a nearly 2x improvement.

## Demo

Without backend rendering: https://perf-main-frontend.dev.imaging.cziscience.com/
With backend rendering: https://speed-up-hub-frontend.dev.imaging.cziscience.com


backend/napari.py
```diff
@@ -223,10 +233,14 @@ def format_plugin(plugin: dict) -> dict:
         extra_metadata = get_extra_metadata(download_url)
         project_urls = extra_metadata.get('project_urls', {})
 
+    description = extra_metadata.get('description', f'{get_attribute(plugin, ["info", "description"])}')
+    description_text = render_description(description)
```
ziyangczi, line 237: maybe just include this in the extra metadata set

codemonkey800: I looked into it and I'm not sure if I can because `description` can come from the `extra_metadata` or `plugin.info`. Or did you want me to add it to the set after creating it?

```py
extra_metadata['description_text'] = render_description(description)
```

ziyangczi, line 237: I meant: 
```suggestion
    extra_metadata['description_text'] = render_description(description)
```

backend/napari.py
```diff
@@ -223,10 +233,14 @@ def format_plugin(plugin: dict) -> dict:
...
+
     return {
         "name": get_attribute(plugin, ["info", "name"]),
         "summary": extra_metadata.get('summary', get_attribute(plugin, ["info", "summary"])),
-        "description": extra_metadata.get('description', f'{get_attribute(plugin, ["info", "description"])}'),
+        "description": description,
+        "description_text": description_text,
```
ziyangczi, line 243:
```suggestion
        "description_text": extra_metadata.get('description_text', ''),
```

ziyangczi (changes requested):
added suggestion to package description text

ziyangczi (changes requested):
would be nice to package the description text in extra metadata. sorry I can't make it one commit, I tried to add these change to a batch not not sure why they are now showing as a batch. but the idea is to add some clarity where these metadata is coming from

backend/napari.py
ziyangczi: hmm somehow I can't batch these changes... but they should go in together

backend/napari.py
ziyangczi: so you can also undo the change above. though minor but this allows us to consistently see where everything is coming from

frontend/src/components/PluginSearch/PluginSearchResult.tsx
```diff
@@ -171,12 +171,15 @@ export function PluginSearchResult({ className, matches, plugin }: Props) {
           </ul>
 
           {/* Search preview of plugin description. */}
-          {isSearching && matches.description && (
+          {isSearching && matches.description_text && (
```
ziyangczi, line 174: if we are not using the raw description anywhere in the frontend code, we should just drop description from the json, thoughts?

ziyangczi:
a general question, is there any identified difference between the beautifulsoup parsing solution to the current approach? particularly if anything bs4 can't handle/translate to HTML?

frontend/src/components/PluginSearch/PluginSearchResult.tsx
codemonkey800: Yeah we can remove from the `index_subset`, but we still need it for the plugin page 😅

codemonkey800:
> a general question, is there any identified difference between the beautifulsoup parsing solution to the current approach? particularly if anything bs4 can't handle/translate to HTML?

BeautifulSoup is a full HTML parser, so I can't imagine there being a case where it can't handle HTML. The current approach actually could use BeautifulSoup to parse the pypi plugin page and query the data.

For my use case, the main benefit of using BeautifulSoup is the convenience of using `soup.get_text()` to get the textual content for the entire HTML document. If we wanted to parse for text using the current approach, we would need to write a lot of custom logic because we would need to recursively look at each HTML node and determine if it has text content.

ziyangczi (approved):

PR #206, opened 2021-07-14
Client-side skeleton loader

codemonkey800:
## Description

This PR adds [skeleton loaders](https://material-ui.com/components/skeleton/) for the plugin and search pages when navigating on the client-side. This will improve perceived performance when loading uncached pages because the user will now see a loading page almost immediately when loading the plugin or search page.

However, because the loader is only visible during client-side transitions, the performance of initial load is unaffected. I originally worked on a dynamic skeleton loader that would [fetch data on the client-side](https://github.com/chanzuckerberg/napari-hub/tree/jeremy/improve-initial-load-skel-loader), reducing the initial load to less than 100ms. But I realized later that this would [have an impact on SEO](https://www.if-so.com/dynamic-content-impact-on-seo/), so I'm deferring that improvement until we do the SEO and performance audit.

## Scrolling Behavior

Now that we're rendering a skeleton loader for the plugin and search page, the browser doesn't know where to scroll back once the content is loaded. To fix this, I needed to implement custom scroll behavior in [usePageTransitions()](https://github.com/chanzuckerberg/napari-hub/blob/jeremy/client-skel-loader/frontend/src/hooks/usePageTransitions.ts).

### Remove Global Smooth Scroll

The first thing I did was remove the CSS for global smooth scroll. I did this because smooth scrolling everything felt awkward while using the hub (and [others agree](https://css-tricks.com/downsides-of-smooth-scrolling/)). I think it makes more sense to make smooth scroll opt-in so that we use it when it makes sense for UX. Right now, the only place where I have smooth scrolling is when the user submits a search query.

### Automatic Scrolling on Load

When the page starts loading, the app will automatically scroll the user to a certain location depending on the page:

1. For the plugin page, the browser will scroll to the top of the page and stay there until it finished loading.
1. For the search page, the browser will:
    1. Scroll to the last location the user scrolled to if navigating back.
    1. Scroll to the top if clicking on the home link or when using the search bar.

### Cancel Automatic Scroll on User Scroll

If the user scrolls at any time while the page is loading, the page will cancel automatic scrolling so that the user isn't taken away from their current scroll location.

## Changes

- Added `<SkeletonLoader />`: Component for rendering a skeleton loader in place for a component while loading.
- Context for global loading state. This is used to provide shared load state for an entire component tree.
- Removes global smooth scroll. There are cases where it's better to have instant scroll vs smooth scroll, so now smooth scroll has been made opt-in
- Renamed `useSearchSetSortType()` to `useSearchEffects()` so that the hook is generalized for any type of effect that runs when the search query changes. 
- Enable client-side page transitions using [Next.js router events](https://nextjs.org/docs/api-reference/next/router#routerevents)
- Moved "scroll on query" from [\<PluginSearchBar /\>](https://github.com/chanzuckerberg/napari-hub/commit/c0534323e292dfdf4dc913d7d45b17a5145b17b5#diff-0e579ca2de3fe8a6ae41e6dbaf96f8bf0bb07bac2e1072e9a866d4d96aa3e7e9) to [useSearchEffects()`](https://github.com/chanzuckerberg/napari-hub/commit/c0534323e292dfdf4dc913d7d45b17a5145b17b5#diff-f93663b98e08abc1abea4cd93bbfe6bc57fd880b8f34b5dd11bb1f072413e2fe):
  - This fits better architecturally since it is a side effect for the search query state.
  - Fixes issue where search query scroll was overriding the page transition scroll.

## Demos

https://speed-up-hub-frontend.dev.imaging.cziscience.com

### Automatic Scrolling on Load

https://user-images.githubusercontent.com/2176050/125697829-c9efad8d-7d8d-4dc4-8440-c275c55389a5.mp4

### Cancel Automatic Scroll on Load

https://user-images.githubusercontent.com/2176050/125697841-fd461845-2e95-4cce-956d-dfe8f1fd3368.mp4

## Follow Ups

- I don't know how the CDN will behave with the skeleton loader. Since the CDN isn't enabled for dev environments, I'll have to run a smoke test on https://staging.napari-hub.org.

### DIFF
--- a/frontend/src/hooks/usePageTransitions.ts
+++ b/frontend/src/hooks/usePageTransitions.ts
@@ -55,17 +55,17 @@ export function usePageTransitions() {
     scrollYRef.current = window.scrollY;
   }, []);
 
-  function addScrollHandler() {
-    scrollYRef.current = 0;
-    shouldScrollRef.current = true;
-    document.addEventListener('scroll', loadingScrollHandler);
-  }
-
-  function removeScrollHandler() {
-    document.removeEventListener('scroll', loadingScrollHandler);
-  }
-
   useEffect(() => {
+    function addScrollHandler() {
+      scrollYRef.current = 0;
+      shouldScrollRef.current = true;
+      document.addEventListener('scroll', loadingScrollHandler);
+    }
+
+    function removeScrollHandler() {
+      document.removeEventListener('scroll', loadingScrollHandler);
+    }
+
     function onLoading(url: string, { shallow }: RouteEvent) {
       if (shallow) return;
 
--- a/frontend/src/hooks/usePageTransitions.ts
+++ b/frontend/src/hooks/usePageTransitions.ts
@@ -89,13 +89,7 @@ export function usePageTransitions() {
         //
         // https://mzl.la/36x9rzG
         requestAnimationFrame(() => {
-          // This function gets the distance using `offsetTop`, which will cause
-          // a reflow. While this is usually bad for animations, it's fine here
-          // because we're using `requestAnimationFrame()` for scheduling
-          // instead of animation, so the function is only being called once per
-          // page load.
-          const distanceFromTop = getSearchBarDistanceFromTop();
-          window.scroll(0, Math.min(scrollY, distanceFromTop));
+          window.scroll(0, scrollY);
 
           // Schedule scroll handler registration for next frame so that above
           // scroll doesn't trigger a scroll event.
@@ -119,19 +113,7 @@ export function usePageTransitions() {
       removeScrollHandler();
 
       if (isSearchPage(url)) {
-        setTimeout(() => {
-          const scrollY = getSearchScrollY();
-
-          if (window.scrollY !== scrollY && shouldScrollRef.current) {
-            // Schedule scroll on macrotask queue for execution later. This is
-            // required because at runtime, the DOM hasn't finished loading yet
-            // for the current page.
-            window.scroll({
-              top: scrollY,
-              behavior: 'smooth',
-            });
-          }
-        });
+        console.log('finished loading search page');
       } else if (isPluginPage(url)) {
         // Scroll to current scroll position while plugin page was loading. If
         // the user didn't scroll at all, this will be 0.
--- a/frontend/src/hooks/usePageTransitions.ts
+++ b/frontend/src/hooks/usePageTransitions.ts
@@ -155,7 +155,7 @@ export function usePageTransitions() {
       router.events.off('routeChangeComplete', onFinishLoading);
       router.events.off('routeChangeError', onError);
     };
-  }, [addScrollHandler, removeScrollHandler, router]);
+  }, [loadingScrollHandler, router]);
 
   return { loading, nextUrl };
 }
--- a/frontend/src/hooks/usePageTransitions.ts
+++ b/frontend/src/hooks/usePageTransitions.ts
@@ -1,12 +1,7 @@
 import { useRouter } from 'next/router';
 import { useCallback, useEffect, useRef, useState } from 'react';
 
-import {
-  getSearchBarDistanceFromTop,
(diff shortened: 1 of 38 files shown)


codemonkey800:
@liaprins-czi would love your feedback on the scroll behavior and if there's any modifications I can make to it.

liaprins-czi (approved):
This looks great to me, just how I imagined it would be! 🥳

The only thing I might change from stylistic perspective (and these are very small not important things) might be:
• Have all the elements the same color (right now the line separators are still black so they kind of stick out too much)
• try the light grey color we use for the background of the metdata at top of page on plugin page (#F7F7F7); this is a very light grey so it might actually be too light (if you do end up trying it let me know and we can see what we think); but usually I think really subdued  / subtle skeletons work really well

justinelarsen:
This looks amazing! I love it! Great work @codemonkey800 !! 🚀 

One question and possible suggestion is on the search bar. I noticed that when the page loads, that's greyed out like the skeleton content. Was that intentional? If so, I would suggest we just leave the search bar as is unless you're afraid that people will try interacting with it? If so, maybe a more subtle design would be better like lightening the text box area and search icon, but I'll leave that up to Lia if that's the intention.

codemonkey800:
> This looks amazing! I love it! Great work @codemonkey800 !! 🚀
> 
> One question and possible suggestion is on the search bar. I noticed that when the page loads, that's greyed out like the skeleton content. Was that intentional? If so, I would suggest we just leave the search bar as is unless you're afraid that people will try interacting with it? If so, maybe a more subtle design would be better like lightening the text box area and search icon, but I'll leave that up to Lia if that's the intention.

Yes this was intentional. There's an issue where the search and filters flash from without to with data, so I opted to hide those with skeleton loaders too. I would like for the static controls like the filters and search bar to be visible, but there's also a state refactor that I want to work on in the future that would make it easier to work on this.

But I am also fine with a more subtle design as you suggest 😸

liaprins-czi:
> If so, maybe a more subtle design would be better like lightening the text box area and search icon, but I'll leave that up to Lia if that's the intention.

Good catch about the search bar Justine. And that is how I would do it: keep the thin line we have as the search bar and the icon also, but change their color to the grey color. Let me know @codemonkey800 if you want me to mock it up!

frontend/src/hooks/usePageTransitions.ts
```diff
@@ -0,0 +1,136 @@
...
+        // ready, so layout hasn't been calculated for the page. Because of
+        // this, the distance from top will be `0`.
+        //
+        // The fix is to schedule for later execution using
+        // `requestAnimationFrame()`. This will call the function before
+        // painting and layout, allowing us to set scroll the before the first
+        // render. As a result, this avoids flickering between the top of the
+        // page and the `scrollY` location.
+        //
+        // https://mzl.la/36x9rzG
```
justinelarsen, line 85: Link is broken. I got a page not found error. Can you replace with a working link please?

frontend/src/hooks/usePageTransitions.ts
```diff
@@ -0,0 +1,136 @@
...
+          // Schedule scroll handler registration for next frame so that above
+          // scroll doesn't trigger a scroll event.
+          requestAnimationFrame(addScrollHandler);
+        });
+      } else if (isPluginPage(url)) {
+        // Scroll to top of the plugin page loader.
+        window.scroll(0, 0);
+
+        // Schedule as macrotask so above scroll doesn't trigger a scroll event.
+        setTimeout(addScrollHandler);
```
justinelarsen, line 98: This part looks interesting... Comments are super helpful!

justinelarsen (approved):
Looks good to me! I just left a couple of comments.

frontend/src/hooks/usePageTransitions.ts
codemonkey800: Weird, the page got removed prior to your review 😅

PR #211, opened 2021-07-19
Redirect non-www to www

### LINKED ISSUE #141: hub is rendering at both www.napari-hub.org and napari-hub.org
For SEO purposes, we should make sure that the site only renders at either www.napari-hub.org or napari-hub.org, but not both.

My understanding is that www.napari-hub.org is preferable, as it will better future-proof us if we start to need other subdomains (e.g. api.napari-hub.org for #133)

Whichever we choose, we should ensure that the other has a permanent redirect (e.g. redirect napari-hub.org to www.napari-hub.org)

codemonkey800:
## Descriptions

Closes #141

Sets up root domain redirection to `www` subdomain. This works by adding nginx as a reverse proxy for the Next.js application in the production Dockerfile. This allows us to set up the redirection independent of the application code and to maintain [automatic static optimization](https://nextjs.org/docs/advanced-features/automatic-static-optimization).

## Demo

**TODO Needs dev deployment**

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -17,6 +17,9 @@ FROM node:16-alpine
 
 WORKDIR /app
 
+# Install supervisord for process management and ngninx for the web server.
+RUN apk update && apk add --no-cache supervisor nginx && mkdir /run/nginx
+
 # Install dependencies for production
 COPY package.json yarn.lock postinstall.js /app/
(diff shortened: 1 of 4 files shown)


codemonkey800:
Closing because an infra PR has been opened to handle the redirection with an AWS ALB

PR #212, opened 2021-07-20
Fix search param encoding

### LINKED ISSUE #138: spaces are improperly serialized/urlencoded when searching from the details page
#### Description
If I do a search with more than one term from a plugin details page, the space(s) are not serialized properly, returning no results.

#### Steps/Code to Reproduce

1. start from any [details page](https://napari-hub.org/plugins/napari-pyclesperanto-assistant)
2. at the top bar, search for "cell counting"

#### Expected Results

image

#### Actual Results

image

Note URL has the search term as `cell%2520counting` when it should be `cell%20counting`. It appears to be getting serialized for the URL twice

codemonkey800:
## Description

Closes #138

The Next.js router will [automatically encode's query parameters, but it doesn't decode them](https://stackoverflow.com/a/65421830). This explains why the string `cell counting` was being double-encoded to `cell%2%520detection`.

## Demo

1. Go to https://fix-spaces-frontend.dev.imaging.cziscience.com/plugins/affinder
1. Enter `cell counting` into the search bar and press `Enter`
1.  There should be one search result, the value for `?search` should be `cell%20counting`, and the search bar should have `cell counting`.

https://user-images.githubusercontent.com/2176050/126376768-8700b67e-60f9-45cc-81f1-b994a696c0bc.mp4

### DIFF
--- a/frontend/src/components/SearchBar/SearchBar.tsx
+++ b/frontend/src/components/SearchBar/SearchBar.tsx
@@ -19,19 +19,6 @@ interface Props extends HTMLProps<HTMLFormElement> {
 }
 
 /**
- * Creates a new URL with the search query added.
- *
- * @param query The query string.
- * @returns The URL object.
- */
-function getURLWithSearchParam(query: string): URL {
-  const url = new URL(SEARCH_PAGE, window.location.origin);
-  url.searchParams.set(SearchQueryParams.Search, encodeURIComponent(query));
-
-  return url;
-}
-
-/**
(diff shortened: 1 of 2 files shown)


kne42 (approved):
LGTM. one question: might we want to use `+` to denote spaces like google does?

codemonkey800:
> LGTM. one question: might we want to use `+` to denote spaces like google does?

That sounds like a good idea. We need to overhaul the encoding for all query parameters, so it's something we can consider when it's time for that work 😄

Issue #140, opened 2021-07-06
broken markdown & nav for napari-properties-viewer


neuromusic:
#### Description

Rendering the description for napari-properties-viewer appears to be broken.

- right nav doesn't work
- missing image
- "image" in h2

#### Steps/Code to Reproduce

visit https://www.napari-hub.org/plugins/napari-properties-viewer

#### Expected Results

A cleanly rendered description :D

#### Actual Results

![image](https://user-images.githubusercontent.com/1245615/124640536-66de8900-de42-11eb-901d-36931dd0af08.png)

It seems like the markdown parser might be mis-interpreting what is intended to be a horizontal line as designating the image (which can't render b/c it is referenced as a relative link) as a heading?

Markdown here: https://raw.githubusercontent.com/kevinyamauchi/napari-properties-viewer/main/README.md

kne42:
Okay so there are three aspects that you've identified to this issue:
1. markdown nav breaks when there are invalid headers – bug; fixed by #220
2. missing image due to it being a relative link – do you desire to add support for resolving relative links as a new feature?
3. image appearing in h2 due to no newline between it and a separator – however [this is how the syntax works](https://www.markdownguide.org/basic-syntax/#alternate-syntax) and I believe that we shouldn't fix users' faulty markdown

How would you like to approach (2) and (3) @neuromusic?

neuromusic:
for 2, no... let's not try to add support for this (yet). instead, let's add an entry to the FAQ "Why aren't my images rendering on my plugin description?" and clarify that we only support absolute links.

for 3, that's interesting. Github-flavored markdown must not support the alternate syntax, because it renders as a separator on the Github repo: https://github.com/kevinyamauchi/napari-properties-viewer 

I think folks will be surprised if we are rendering markdown differently from Github... should we consider switching to GFM? is this easily supported by our markdown parser? e.g. with `marked` its just a single [option](https://marked.js.org/using_advanced#options) to switch to the GFM spec

codemonkey800:
It looks like like (2) and (3) are both caused by the fact we don't support relative images. It should be easy to support provided we follow a few heuristics:

1. If the image is an absolute URL, then use the URL directly
    - Example: `https://example.com/image.png`
2. If the image is a relative URL but does not start with a prefix, then append to repo's raw URL.
    - Example: `resources/image.png` -> `https://raw.githubusercontent.com/<owner>/<repo>/main/resources/image.png`
[...]

PR #220, opened 2021-07-23
disregard invalid headers in markdown ToC

kne42:
sometimes markdown can generate invalid headers, such as in issue #140. this excludes any headers missing an id from the ToC parser


codemonkey800 (approved):
awesome!

PR #222, opened 2021-07-29
Add sitemap.xml and robots.txt

codemonkey800:
## Description

Adds dynamically generated `sitemap.xml` and `robots.txt` routes. 

### `sitemap.xml`

The `sitemap.xml` file is responsible for [listing the most important pages for Google to crawl and index](https://yoast.com/what-is-an-xml-sitemap-and-why-should-you-have-one/). The generated `sitemap.xml` includes all pages (including plugins) using the Next.js build manifest and hub `/plugins` API.

### `robots.txt`

The `robots.txt` file is used for [controlling which URLs should be crawled by a search engine](https://developers.google.com/search/docs/advanced/robots/intro).

### Indexing

We block indexing for non-production environments by using the [robots meta tag](https://developers.google.com/search/docs/advanced/crawling/block-indexing). This is so that non-production environments aren't indexed by Google Search. 

## Demos

- [/sitemap.xml](https://sitemap-frontend.dev.imaging.cziscience.com/sitemap.xml)
- [/robots.txt](https://sitemap-frontend.dev.imaging.cziscience.com/robots.txt)


frontend/src/pages/sitemap.xml.ts
```diff
@@ -0,0 +1,109 @@
...
+    | 'yearly'
+    | 'never';
+  priority?: number;
+}
+
+function getEntry(url: string): SitemapEntry {
+  return {
+    url,
+    changefreq: 'daily',
+    priority: 0.7,
```
klai95, line 29: Can you explain the range of values `priority` can take on so I can understand the relative significance of its value of 0.7?

codemonkey800: Thanks for calling this out, I did a bit more reading and I'm actually going to remove this property. Generally this value is used for [increasing the likelihood of important pages being indexed](https://www.sitemaps.org/protocol.html#xmlTagDefinitions).

However, I just found out that [Google doesn't use these values](https://developers.google.com/search/docs/advanced/sitemaps/build-sitemap#general-guidelines):
image

I updated the code to instead use `lastmod`, which is basically the last time the page content was modified. I only did it for the plugin page entries though because that information isn't available for the other pages.

klai95 (approved):
LGTM!

kne42 (approved):
LGTM

PR #223, opened 2021-08-02
easy citation reading

ziyangczi:
add the citation.cff indexing on the backend without transformation.

Without knowing frontend plan on what we plan to display we can't decide on what format of data we should be storing, it is possible for the frontend to transform the data into a different citation format.

Also worth noting that the official citation [python converter ](https://github.com/citation-file-format/cff-converter-python) release is quite old and required some libraries (click==0.6.*) which are not compatible with what we use, so for now I didn't go into it further to make it work. But if we need to add transformation on the backend we could either install the library without dependency or add transformational logic on our own to make apa or bibtex which is currently provided by github


codemonkey800 (approved):
LGTM!! 🍜

PR #224, opened 2021-08-02
Fix contrast ratio for napari gray color

codemonkey800:
## Description

Fixes the [contrast ratio issue](https://web.dev/color-contrast/) reported by Lighthouse for the napari hub by darkening the napari gray color.


liaprins-czi (approved):
Looks good, this new hex color passes a11y for all the background colors we have it on!

PR #227, opened 2021-08-05
Return plugin specific data for /plugins/:name for mock server

codemonkey800:
## Description

Updates the mock server to return plugin specific data instead of returning `napari-compressed-labels-io` all the time.

### DIFF
--- a/frontend/mock-server.js
+++ b/frontend/mock-server.js
@@ -4,6 +4,7 @@ if (process.env.MOCK_SERVER === 'false') {
 }
 
 const express = require('express');
+const { set } = require('lodash');
 
 const napariPlugin = require('./src/fixtures/napari.json');
 const pluginIndex = require('./src/fixtures/index.json');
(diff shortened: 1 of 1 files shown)


frontend/mock-server.js
```diff
@@ -4,22 +4,37 @@ if (process.env.MOCK_SERVER === 'false') {
...
 
 const napariPlugin = require('./src/fixtures/napari.json');
 const pluginIndex = require('./src/fixtures/index.json');
 
 const app = express();
 
 app.get('/plugins', async (_, res) => {
-  res.json({ 'napari-compressed-labels-io': '0.0.0' });
+  const versions = pluginIndex.reduce(
+    (result, { name }) => set(result, name, '0.0.0'),
```
klai95, line 16: Clarifying questions: Why is the version set to '0.0.0'?

codemonkey800: Technically the version doesn't really matter for mock data. But then again, the plugin version is available in the index and it's not too much work to add it 😄

klai95 (approved):
Looks good to me after the feedback!

ziyangczi (approved):

PR #229, opened 2021-08-06
add citation parsing for different formats

ziyangczi:
Added parsing of citation into RIS and BibTex format, pending on APA format on discussion here: https://github.com/citation-file-format/cff-converter-python/issues/157

### DIFF
--- a/backend/napari.py
+++ b/backend/napari.py
@@ -167,12 +167,12 @@ def get_extra_metadata(download_url: str) -> dict:
 
     citation_file = get_file(download_url, "CITATION.cff")
     if citation_file is not None:
-        extra_metadata['citations'] = (get_citation(citation_file))
+        extra_metadata['citations'] = get_citations(citation_file)
 
     return extra_metadata
 
 
-def get_citation(citation_str: str) -> dict:
+def get_citations(citation_str: str) -> dict:
     """
     Get citation information from the string.
     :param citation_str: citation string to parse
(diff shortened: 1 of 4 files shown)


codemonkey800 (approved):
LGTM 🔥

backend/napari.py
```diff
@@ -164,13 +165,33 @@ def get_extra_metadata(download_url: str) -> dict:
         config = yaml.safe_load(yaml_file)
         extra_metadata.update(config)
 
-    citation = get_file(download_url, "CITATION.cff")
-    if citation is not None:
-        extra_metadata['citation'] = citation
+    citation_file = get_file(download_url, "CITATION.cff")
+    if citation_file is not None:
+        extra_metadata['citations'] = (get_citation(citation_file))
```
kne42, line 170: why is the `get_citation` call wrapped in parentheses?

backend/napari.py
```diff
@@ -164,13 +165,33 @@ def get_extra_metadata(download_url: str) -> dict:
...
 
     return extra_metadata
 
 
+def get_citation(citation_str: str) -> dict:
```
kne42, line 175: I find it weird that this is named `get_citation` (singular) as opposed to `get_citations` (plural) since it is referred to in plural for everywhere else

kne42 (approved):
LGTM~ left some code style questions, feel free to ignore

backend/napari.py
ziyangczi: nice catch!

backend/napari.py
ziyangczi: nice catch! it should have been renamed to accommodate the dictionary

PR #253, opened 2021-08-25
Plugin visibility feature

### LINKED ISSUE #247: hidden plugins
_Discussed in https://github.com/chanzuckerberg/napari-hub/discussions/134#discussioncomment-953855_

### Summary

There are times where plugin developers want their plugin on PyPI, but don't want it to show up on the napari hub. We will let plugin developers control the visibility of their plugin on the napari hub.

### Intended outcome

Plugin developers will be able to "disable" or "hide" their plugin listing by configuring parameters in their Github repo

### How will it work?

Plugin developers will be able to specify a `visibility` parameter in `.napari/config.yml` that accepts 3 levels of "visibility" on the hub:

- `public`: fully visible. (this is the default value)
- `disabled`: no plugin page created, does not show up in search listings. (equivavent to being on the exclusion list)
- `hidden` plugin page existss, but the doesn't show up in search listings

klai95:
Closes #247

Clubhouse Ticket: https://app.clubhouse.io/imaging/story/157409/plugin-visibility-feature

Visibility Flag Values:
- Public (default, fully visible)
- Hidden (listing page exists, but doesn't show up in search. useful for pre-release development)
- Disabled (no listing at all)


backend/napari.py
```diff
@@ -396,19 +413,21 @@ def cache_available(key: str, ttl: [timedelta, None]) -> bool:
         return False
 
 
-def filter_excluded_plugin(packages: dict) -> dict:
+def filter_excluded_plugin(packages: dict, white_list: set = {}) -> dict:
     """
     Filter excluded plugins from the plugins list
     :param packages: all plugins list
+    :param white_list: whitelisted plugins
```
ziyangczi, line 420: this is whitelisted keyword, maybe update the parameter name to reflect this as well

backend/napari.py
```diff
@@ -117,6 +117,7 @@ def update_index() -> dict:
                    for k, v in get_plugins().items()]
     for future in concurrent.futures.as_completed(futures):
         results.append(future.result())
+    s3_client.upload_file("s3", endpoint_url=endpoint_url)
```
ziyangczi, line 120: you should use cache() function since s3 upload are all done here

klai95: The return statement of `update_index()` is `return jsonify(cache(results, index_key))` and thus it already uses the cache method right now. I will figure out what is missing in the method.

backend/napari.py
```diff
@@ -144,12 +145,13 @@ def get_file(download_url: str, file: str) -> [dict, None]:
...
     """
-    Extract extra metadata from the github download url
+    Extract extra metadata from the github download url.
 
-    :param download_url: github url to download from
+    :param download_url: github repo url to download from
     :return: extra metadata dictionary
     """
     extra_metadata = {}
+    visibility_set = {'public', 'disabled', 'hidden'}
```
justinelarsen, line 154: Can you add some comments describing what each of the values is used for?
public: fully visible (default)
disabled: no plugin page created, does not show up in search listings
hidden: plugin page exists, but doesn't show up in search listings

backend/napari.py
```diff
@@ -303,6 +310,16 @@ def format_plugin(plugin: dict) -> dict:
...
 
 @app.route('/plugins')
+def get_plugins_filtered() -> dict:
+    """
+    Get all filtered plugins list such that only public plugins are returned
+
+    :param context: context for the run to raise alerts
+    :return: json of filtered plugins and their version
+    """
+    return filter_excluded_plugin(get_plugins())
```
justinelarsen, line 320: I'm not sure that this new method does anything more than the "get_plugins" method since the "get_plugins" method is already calling "filter_excluded_plugin". Is it necessary?

backend/napari.py
```diff
@@ -350,6 +367,7 @@ def get_plugin(plugin: str, version: str = None) -> dict:
     :return: plugin metadata dictionary
     """
     plugins = get_plugins()
+    plugins = filter_excluded_plugin(plugins, {'hidden'})
```
justinelarsen, line 370: Is this filtering out the hidden plugins? If so, doesn't that mean that we won't return any data for the plugin detail page for a particular plugin that is set to hidden? From the product requirements, it seems like we want to return data for plugins even if they are set to hidden which is useful for users to see their meta data without it being published on the search page.

backend/napari.py
```diff
@@ -396,19 +414,21 @@ def cache_available(key: str, ttl: [timedelta, None]) -> bool:
...
     Filter excluded plugins from the plugins list
     :param packages: all plugins list
+    :param whitelisted_keyword: keyword reflects plugins to include
     :return: only plugins not in the filtered list
     """
-    filtered = packages.copy()
-    exclusions = get_exclusion_list()
-    for exclusion, versions in exclusions.items():
-        if exclusion in packages and \
-                (versions is None or packages[exclusion] in versions):
```
justinelarsen, line 409: Is this version check code not needed anymore?

justinelarsen (commented):
@klai95 Nice work! I left a few comments if you could review and fix/clarify. Thanks!

backend/napari.py
ziyangczi: we are changing the format of the exclusion file so we no longer check for version in our exclusion logic. The new flow would treat exclusion information at plugin level, instead of version level

backend/napari.py
ziyangczi: There are some flow changes that are yet to implement here, I have talked to Kevin about this, and we will update the logic here soon

ziyangczi: (I think the PR should have been opened as a draft and not completely ready. I have talked to Kevin about draft PRs for future reference)

justinelarsen: Sounds good. Thanks!

backend/napari.py
justinelarsen: Got it!

backend/napari.py
klai95: Will do, thank you for the comment!

backend/napari.py
klai95: Yeah, as Ziyang said, I was not aware of the draft feature for PR, but moving forward I will me sure to tag it!

backend/napari.py
klai95: This is including the hidden plugins. The second parameter of `filter_excluded_plugin` is the keyword that reflects plugins to include. At default without specification of the second parameter, only public plugins are returned.

backend/napari.py
```diff
@@ -101,7 +101,7 @@ def get_index() -> dict:
     if cache_available(index_key, cache_ttl):
```
ziyangczi, line 101: we can just skip the cache_available check and always return the cache here

backend/napari.py
```diff
@@ -117,6 +117,21 @@ def update_index() -> dict:
                    for k, v in get_plugins().items()]
```
ziyangczi, line 117: The flow needs to be reconsidered here:
The list of plugins is fetched via get_plugins() call, and later fetched again using get_cache call. It should be fetched once via get_plugins() and used in both places

ziyangczi: also the app.route('/plugins') below should be refactored so that the API only reads the cached version

backend/napari.py
```diff
@@ -117,6 +117,21 @@ def update_index() -> dict:
...
     for future in concurrent.futures.as_completed(futures):
         results.append(future.result())
+    # read cache version of plugins.json
+    plugins = get_cache(plugins_key)
+    # include public and whitelisted plugins in updated_plugins
+    updated_plugins = filter_excluded_plugin(plugins, {'hidden'})
+    # read cache version of excluded_plugins.json
+    excluded_plugins = get_exclusion_list()
```
ziyangczi, line 125: the get_exclusion_list is called both in this function and the filter_excluded_plugin function, ideally it should only be called once to do the update

backend/napari.py
```diff
@@ -331,7 +355,7 @@ def get_plugins() -> dict:
     packages = query_analysis_dump()
 
     if packages:
-        return cache(filter_excluded_plugin(packages), index_key)
+        return get_cache(filter_excluded_plugin(packages), index_key)
```
ziyangczi, line 358: this seems wrong, this function (get_plugins) shouldn't be changed, it should be kept as is, (the `return cache` here would cache the results we found from analysis dump and then return it)

instead, the changes should reroute the '/plugins' API to use another function that only returns the get_cache

backend/napari.py
```diff
@@ -564,4 +588,4 @@ def cache(content: [dict, list], key: str) -> dict:
         fp.write(json.dumps(content))
         fp.flush()
         s3_client.upload_file(fp.name, bucket, os.path.join(bucket_path, key))
-    return content
+    return content
```
ziyangczi, line 591: nit picking: missing empty line at the end

backend/napari.py
klai95: Why was the if statement needed before?

ziyangczi: it was used to check whether the cache is up to date, this part of responsibility has been moved to update_index

backend/napari.py
```diff
@@ -122,11 +119,33 @@ def update_index() -> dict:
     :return: json for index page metadata
     """
     results = []
+    plugins = get_plugins()
```
ziyangczi, line 122: I think this is the last remaining issue:
you are calling get_plugins here, which always read the cache, that is not the correct logic here. For this method, the flow should look like this:
1. query pypi for latest list of plugins: plugins
2. collect all metadata for these plugins: plugins_metadata
3. read the exclusion list: excluded_plugins
4. update exclusion list based on the plugins_metadata and cache it
5. remove excluded plugins from the plugin and plugins_metadata
6. read the cached list of plugins, this is what we got from the last run
7. if the plugins list is not empty, compare the difference and notify zulip via notify_new_packages, cache the plugins list and plugins_metadata
8. continue on with analysis dump

ziyangczi: you may want to compare your current flow against this and see which part is missing/having issue, we can go through them in person later as well

backend/napari.py
```diff
@@ -417,19 +426,20 @@ def cache_available(key: str, ttl: [timedelta, None]) -> bool:
...
     """
-    filtered = packages.copy()
-    exclusions = get_exclusion_list()
-    for exclusion, versions in exclusions.items():
-        if exclusion in packages and \
-                (versions is None or packages[exclusion] in versions):
-            filtered.pop(exclusion)
-    return filtered
+    valid_plugins = packages.copy()
+    exclusion = excluded_plugins
```
ziyangczi, line 438: nit picking: this assignment is not necessary

backend/napari.py
```diff
@@ -417,19 +426,20 @@ def cache_available(key: str, ttl: [timedelta, None]) -> bool:
         return False
 
 
-def filter_excluded_plugin(packages: dict) -> dict:
+def filter_excluded_plugin(packages: dict, excluded_plugins: dict = None, whitelisted_keyword: set = {}) -> dict:
```
ziyangczi, line 429: nit picking: None as default variable is not very useful

backend/napari.py
```diff
@@ -324,29 +352,10 @@ def get_plugins() -> dict:
     :param context: context for the run to raise alerts
     :return: json of valid plugins and their version
     """
-    if cache_available(plugins_key, cache_ttl):
+    if cache_available(plugins_key, None):
```
ziyangczi, line 355: documentation for this method is not updated

backend/_tests/test_lambda.py
```diff
@@ -97,14 +97,14 @@ def raise_for_status(self):
   "upload_time_iso_8601":"2020-04-20T15:28:53.386281Z",
   "url":"","yanked":false,"yanked_reason":null}]}}"""
 
+
 @mock.patch(
-    'requests.get', return_value=FakeResponse(data=plugin_list)
+    'backend.napari.get_plugins', return_value={'test2': '0.0.2'}
 )
 def test_get_plugins(mock_get):
```
ziyangczi, line 104: if you are testing get_plugins(), you should be mocking the function calls it makes, instead of the function call itself, the philosophy for unit test is that "if the modules used in the function in question are functioning correctly, would the function in question behaves as expected", maybe rewrite some of the unit tests following this mindset

ziyangczi: in this particular case, you may want to mock the get_cache and if_cache_available methods

backend/napari.py
ziyangczi: hmm it should be plugin name and version, visibility shouldn't show up here

klai95: Apologies for the wrong comment. Allow me to ask a clarifying question here: I was under the presumption that `plugins.json` currently contains elements such as "napari-plugin-search": "0.1.0", but that we are in the process of changing that version value to visibility value, or was that for something else?

ziyangczi: that is for the exclusion list

PR #259, opened 2021-09-09
Update shield API to return a valid badge for plugins not listed

DragaDoncila:
Rather than returning blank JSON response when a plugin is not found, we now return a valid JSON schema for a `plugin not found` badge.

Also updated tests to reflect this change.


ziyangczi (approved):
LGTM

PR #260, opened 2021-09-09
Add round trip test through shields.io api for affinder badge

DragaDoncila:
Test ensures we can retrieve the correct badge using our shield API endpoint.

### DIFF
--- a/backend/_tests/test_lambda.py
+++ b/backend/_tests/test_lambda.py
@@ -8,6 +8,33 @@ from backend.napari import get_download_url
 from backend.napari import get_license
 from backend.napari import get_citations
 
+AFFINDER_SVG = ('<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"' 
+                  ' width="139" height="20" role="img" aria-label="napari hub: affinder">'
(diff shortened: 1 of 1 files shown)


backend/_tests/test_lambda.py
```diff
@@ -159,6 +186,13 @@ def test_get_shield(mock_plugins):
     result = get_shield('not-a-package')
     assert result == {}
 
+def test_shield_round_trip():
```
ziyangczi, line 189: Thanks for adding this test, though I think this test should be added somewhere else. The reason is that this file should be for unit tests and run frequently to verify changes of PRs, etc. where the real "round trip test" should happen infrequently maybe before deployment, and should be testing staging instead of production really. maybe for some scenario we would run a separate production test suite as part of production availability check.

We have an end-to-end test for frontend which I think we should also have a test for backend as well.

DragaDoncila:
After some discussion with @codemonkey800 I think we are best off closing this PR.

This test doesn't really *do* anything, because if it passes we can't be assured of it passing in future (e.g. our API or Shields API goes down a second after we run tests), and if it fails it's not necessarily indicative of any change/action being required on our part. In terms of end-to-end testing @codemonkey800's thoughts were that maybe this could be achieved by having the frontend end-to-end testing ping the staging or production API rather than mocking up the backend data.

PR #263, opened 2021-09-18
Fix code highlighting for single backticks in MDX files

DragaDoncila:
This PR fixes code highlighting for single backticks in MDX files e.g. the FAQ page - the `Article` element just needed the appropriate styles. @codemonkey800 please make sure I haven't done anything weird, it's my first foray into the frontend for the hub.

Before this PR:
image

After this PR:
image

### DIFF
--- a/frontend/src/components/LayoutMDX/LayoutMDX.tsx
+++ b/frontend/src/components/LayoutMDX/LayoutMDX.tsx
@@ -5,9 +5,8 @@ import { ReactNode } from 'react';
 import { ColumnLayout, TableOfContents, TOCHeader } from '@/components/common';
 import { Media } from '@/components/common/media';
 
-import { useHeaders } from './LayoutMDX.hooks';
-
 import styles from '../common/Markdown/Markdown.module.scss';
+import { useHeaders } from './LayoutMDX.hooks';
 
 interface Props {
   toc?: boolean;
--- a/frontend/src/components/LayoutMDX/LayoutMDX.tsx
+++ b/frontend/src/components/LayoutMDX/LayoutMDX.tsx
@@ -3,9 +3,9 @@ import Head from 'next/head';
 import { ReactNode } from 'react';
 
 import { ColumnLayout, TableOfContents, TOCHeader } from '@/components/common';
+import styles from '@/components/common/Markdown/Markdown.module.scss';
 import { Media } from '@/components/common/media';
 
-import styles from '../common/Markdown/Markdown.module.scss';
 import { useHeaders } from './LayoutMDX.hooks';
 
 interface Props {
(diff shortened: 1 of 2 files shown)


frontend/src/components/LayoutMDX/LayoutMDX.tsx
```diff
@@ -1,9 +1,11 @@
+import clsx from 'clsx';
 import Head from 'next/head';
 import { ReactNode } from 'react';
 
 import { ColumnLayout, TableOfContents, TOCHeader } from '@/components/common';
 import { Media } from '@/components/common/media';
 
+import styles from '../common/Markdown/Markdown.module.scss';
```
codemonkey800, line 8: Could you change this to an absolute import pls. Some of the pros of using absolute import paths are described here:
https://beqode.com/blog/absolute-vs-relative-import-paths-nodejs

codemonkey800 (commented):
This is awesome Draga! great first frontend PR 😄 I left a comment about the import syntax, but overall things are looking good!

frontend/src/components/LayoutMDX/LayoutMDX.tsx
DragaDoncila: Thanks for picking this up - better now?

codemonkey800: It should be:

```ts
import styles from '@/components/common/Markdown/Markdown.module.scss';
```

DragaDoncila: @codemonkey800 yeah I did try that initially, and it worked locally but the jest tests failed [here](https://github.com/chanzuckerberg/napari-hub/pull/263/checks?check_run_id=3647291590). Any idea why that is?

codemonkey800: Ahh so the issue is that the test configuration is resolving the actual SCSS file before the SCSS mock file. 

Can you update the configuration file please: https://github.com/chanzuckerberg/napari-hub/blob/2abd74cba28033ea99b96ef3f320004712405b00/frontend/jest/test.config.js#L6-L15

You'll need to change the order of the entries:

```ts
{
  moduleNameMapper: {
    /*
      `identity-obj-proxy` returns a string for whatever key you use, so we use
      it for SCSS modules since components use the exports to assign class
      names.
    */
    '^.+\\.module\\.scss$': 'identity-obj-proxy',

    '^@/(.*)$': '<rootDir>/src/$1',
  },
}
```

And this should make the tests pass 😅

DragaDoncila: Alright, tests are looking good now.

codemonkey800 (approved):
awesome, LGTM 💯

PR #268, opened 2021-09-23
Dev plugin visibility

ziyangczi:
Warning: this PR has some major refactors of code structure to support future feature development.

(Q: Why refactor now? A: because of the plugin visibility feature change, we are reworking some of the underlying flow which resulted in complexity increase and fundamental data filter/manipulations, the main block grow past hundreds line of code without breaking down, and made it difficult to develop for future.)

In order to test the refactored code, I have been adding more unit tests along the way and try to regression test against
https://plugin-visibility-frontend.dev.imaging.cziscience.com/
It seems the environment have been working consistently comparing to prod site so far. (other than the newly introduced visibility flag, which is being tested now with napari-demo set to 'hidden' and it seems to be working correctly)

Just to be safe, I plan to leave this dev branch up a few days on regression testing and make sure the zulip messaging is also consistent with prod behavior

ps. did I mention cffconverter updated their library suddenly (8 hours ago when I was working on it) and supported APA? I added it in since it is a very small change


.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -129,7 +129,7 @@ resource "aws_cloudwatch_event_target" "update_target" {
     rule = aws_cloudwatch_event_rule.update_rule.name
     arn = module.backend_lambda.function_arn
     input_transformer {
-        input_template = jsonencode({path = "/plugins/index/update", httpMethod = "GET"})
+        input_template = jsonencode({path = "/update", httpMethod = "POST"})
```
klai95, line 132: Are we refactoring the `path` for the sake of simplicity?"

backend/zulip.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    Send message to zulip
+    :param username: username for the user to post message
+    :param key: api key for the user
+    :param topic: topic in zulip stream to send
+    :param message: message to send
+    """
+    try:
+        data = {
+            'type': 'stream',
```
klai95, line 59: For the `type` and 'to' variables, what are the possible values they can take on?

.happy/terraform/modules/ecs-stack/main.tf
ziyangczi: originally the update is only updating the index file, however right now the update is responsible for keeping content up to date across all resources, so this change is to would reflect the actual logical level of the endpoint

backend/zulip.py
ziyangczi: for zulip API, see https://zulip.com/api/send-message

backend/utils.py
```diff
@@ -0,0 +1,73 @@
...
+
+    :param obj: object to iterate on
+    :param path: list of string to get subpath within json
+    :return: the value if the path is accessible, empty string if not found
+    """
+    part = obj
+    for token in path:
+        if isinstance(part, dict) and token in part:
+            part = part[token]
+        elif isinstance(part, list) and token < len(part):
```
klai95, line 25: If `part` is assigned the value of `obj`, why are we checking isinstance(part, list)?

klai95: Could we refactor `part` as a variable name to a more descriptive one?

backend/s3.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def cache_available(key: str) -> bool:
+    """
+    Check if cache is available for the key.
+
+    :param key: key to check in s3
+    :return: True iff cache exists
+    """
+    if bucket is None:
```
klai95, line 29: We can return False also if key is an empty string?

backend/s3.py
```diff
@@ -0,0 +1,70 @@
...
+    :return: file content for the key if exists, None otherwise
+    """
+    if cache_available(key):
+        return json.loads(s3.Object(bucket, os.path.join(bucket_path, key)).get()['Body'].read())
+    else:
+        print(f"Not cached: {key}")
+        return None
+
+
+def cache(content: Union[dict, list], key: str) -> Union[dict, list]:
```
klai95, line 52: Since we are using type `Union` for the parameter `content`, when do we expect `content` to be of type `dict` vs type `list`?

backend/s3.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+    Cache the given content to the key location.
+
+    :param content: content to cache
+    :param key: key path in s3
```
klai95, line 57: Should we do an error check for the `key` parameter or an empty string check?

backend/s3.py
ziyangczi: true, though there is no use case for that yet

backend/s3.py
ziyangczi: technically it doesn't care, it would cache the given content regardless of type, just that the hints are saying these are supported ones

backend/s3.py
ziyangczi: we don't have anything calling it with empty string right now

backend/pypi.py
```diff
@@ -0,0 +1,101 @@
...
+    url = requote_uri(f"https://pypi.org/search/?q=&o=-created&c=Framework :: napari&page=")
+    while True:
+        try:
+            response = requests.get(f'{url}{page}')
+            if response.status_code != requests.codes.ok:
+                response.raise_for_status()
+            html = response.text
+            names = name_pattern.findall(html)
+            versions = version_pattern.findall(html)
+            assert (len(names) == len(versions))
```
klai95, line 33: What is the purpose of this assert statement?

backend/utils.py
ziyangczi: when iterating the obj, it could be assigned as a list, this util function iteratively try to fetch inner element from a nested dict/list

backend/pypi.py
ziyangczi: if the length doesn't match, the process would just terminate and throw an error, this is likely a case of corrupted input, we shouldn't be proceeding using the result

backend/pypi.py
```diff
@@ -0,0 +1,101 @@
...
+def format_plugin(plugin: dict) -> dict:
+    """
+    Format the plugin metadata to extra relevant information.
+
+    :param plugin: plugin dictionary from pypi
+    :return: formatted plugin dictionary
+    """
+    version = get_attribute(plugin, ["info", "version"])
+
+    return {
```
klai95, line 70: we should include visibility here right?

ziyangczi: nope, the visibility comes from github metadata, the logic is in model.py

backend/pypi.py
ziyangczi: going to return {} here to have caller deal with empty result

backend/utils.py
ziyangczi: renamed part to current_location

backend/model.py
```diff
@@ -0,0 +1,201 @@
...
+
+def get_plugin_metadata_async(plugins: Dict[str, str]) -> dict:
+    """
+    Query plugin metadata async.
+
+    :param plugins: plugin name and versions to query
+    :return: plugin metadata list
+    """
+    plugins_metadata = {}
+    with futures.ThreadPoolExecutor(max_workers=32) as executor:
```
klai95, line 195: How did we decide on `max_worker=32`?

backend/github.py
```diff
@@ -0,0 +1,149 @@
...
+
+    return None
+
+
+def get_license(url: str) -> [str, None]:
+    try:
+        api_url = url.replace("https://github.com/",
+                              "https://api.github.com/repos/")
+        auth = None
+        if github_client_id is not None and github_client_secret is not None:
```
klai95, line 49: Would `if None not in (github_client_id, github_client_secret):` be more concise?

backend/model.py
ziyangczi: it is more or less arbitrary, just trying to parallelize the work, it may not be the optimal setting but good enough so far

backend/github.py
ziyangczi: if we have more items I would be tempted to do so, but I personally found this more readable

.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -37,7 +37,7 @@ locals {
   frontend_alb_zone     = try(local.secret[local.alb_key]["frontend"]["zone_id"], "")
   frontend_alb_dns      = try(local.secret[local.alb_key]["frontend"]["dns_name"], "")
 
-  slack_url = try(local.secret["slack_url"], "")
+  slack_url = try(local.secret["slack"]["url"], "")
```
klai95, line 40: Can you explain the context for this code change?

ziyangczi: we updated the slack URL in the infra config, it does not allow set of string value in the secret so it was wrapped in a library

backend/github.py
```diff
@@ -0,0 +1,149 @@
...
+    citations = {}
+    try:
+        citation = Citation(cffstr=citation_str)
+        citations['citation'] = citation_str
+        citations['RIS'] = citation.as_ris()
+        citations['BibTex'] = citation.as_bibtex()
+        citations['APA'] = citation.as_apalike()
+    except ValueError:
+        # invalid CITATION.cff content
+        citations['citation'] = None
```
klai95, line 145: How about `citations = dict.fromkeys(['citation', 'RIS', 'BibTex', 'APA'], None)`?

backend/model.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+def update_cache():
+    """
+    Update existing caches to reflect new/updated plugins. Files updated:
+    - excluded_plugins.json (overwrite)
+    - cache/public-plugins.json (overwrite)
+    - cache/hidden-plugins.json (overwrite)
+    - cache/index.json (overwrite)
+    - cache/{plugin}/{version}.json (skip if exists)
```
klai95, line 133: Why do we skip if exists?

backend/model.py
```diff
@@ -0,0 +1,200 @@
...
+    for plugin, version in plugins.items():
+        visibility = plugins_metadata[plugin].get('visibility', 'public')
+        if visibility in visibility_plugins:
+            visibility_plugins[visibility][plugin] = version
+
+    for plugin, _ in excluded_plugins.items():
+        if plugin in plugins_metadata:
+            del (plugins_metadata[plugin])
+
+    if visibility_plugins['public']:
```
klai95, line 149: In the case where hidden plugins do exist, how come we are not calling `notify_new_packages` to notify if the conditions are true?

backend/model.py
```diff
@@ -0,0 +1,200 @@
...
+    public: fully visible (default)
+    hidden: plugin page exists, but doesn't show up in search listings
+    disabled: no plugin page created, does not show up in search listings
+
+    :param plugins_metadata: plugin metadata containing visibility information
+    :return: updated exclusion list
+    """
+    excluded_plugins = get_excluded_plugins()
+    for plugin, plugin_metadata in plugins_metadata.items():
+        if 'visibility' not in plugin_metadata:
```
klai95, line 175: if this condition is true, why do we not want to create the `visibility` key in `plugin_metadata`? Also, what would be the case in which this happens?

backend/model.py
ziyangczi: we don't reprocess metadata for specific plugin because pypi metadata does not change

backend/model.py
ziyangczi: we do not send notifications for hidden plugins

backend/model.py
ziyangczi: if visibility does not present we treat it as a public plugin

klai95 (approved):
LGTM!

Issue #245, opened 2021-08-20
Pagination


neuromusic:
### Summary

Adds pagination to improve the navigation

### Intended outcome

Users will only see the most relevant plugins and will be able to more easily access links in the footer (privacy policy, mailing list signup, etc)

### How will this work?

When users search or browse for plugins, they will only see the top 10-20 plugins

### Assets

- [Explorations](https://docs.google.com/document/d/1WLsbJdHNJBKKbT3Hw-Wz3H6ff4N7fgTPeNxoDGBnrJE/edit)
- [Designs](https://www.figma.com/file/b7jLWRhPTZcvZPJiZFijfF/Pagination?node-id=44%3A93)

PR #270, opened 2021-09-24
Search Result Pagination

codemonkey800:
## Description

Adds Pagination to the plugin search page. This adds the pagination state management, Pagination component, and unit + E2E tests.

## Demos

https://pagination-frontend.dev.imaging.cziscience.com

https://user-images.githubusercontent.com/2176050/134603445-31628342-f394-42a6-9528-72f1244c6b62.mp4


liaprins-czi (approved):
Looks perfect to me!

frontend/src/components/PluginSearch/PluginSearchResultList.tsx
```diff
@@ -2,42 +2,53 @@ import clsx from 'clsx';
...
-import { getSkeletonResultCount } from '@/utils';
 
 import { FilterChips } from './FilterChips';
 import { PluginSearchResult } from './PluginSearchResult';
 
+/**
+ * Buffer to add to search results in case the search results content is taller
+ * than the skeleton result height.
+ */
+const RESULTS_PER_PAGE_BUFFER = 5;
```
klai95, line 18: How did you decide that `RESULTS_PER_PAGE_BUFFER = 5` would be sufficient in the case the search results content is taller than the skeleton result height?

codemonkey800: Good call out lol It was a temporary fix for something I noticed during development.
I refactored the loading state to instead use the actual heights from the search result list for the skeleton results.

neuromusic:
this is 15 results per page? I know @liaprins-czi and I went back-and-forth on the number to display, but this _feels_ like a lot of scrolling before I hit the bottom of the page (incl the email signup and footer links).

I wonder if 10 results is better?

codemonkey800:
@neuromusic here's a version with the results changed to 10:
https://pagination-10-frontend.dev.imaging.cziscience.com/

let me know how it compares to the original 15. i'd also like to hear what @liaprins-czi thinks about it too 😄

neuromusic:
I like the 10, but would also like to hear @liaprins-czi's thoughts

neuromusic:
Also: this will close #248

neuromusic:
Also: this will close #245

liaprins-czi (commented):
Looks great @codemonkey800 , thanks for mocking up these options!

To me, the number of results per page depends what we are optimizing for:

If we are optimizing for the user's plugin-searching / -browsing experience, I prefer more plugins per page (15 seemed good to me!)

If we are optimizing for them being able to see the bottom of the page (and not minding that they will have to click more often to see mroe results, which is a barrier that will prevent soem from exploring the next page(s)), then less results per page is better.

To me, it seems like the first use case is teh primary reason a user is at the hub: to explore plugins. And the actions in the footer /subscribe area are secondary use cases. So from that perspective, I prefer having a few more results per page, and I don't think it is too much extra for them to scroll a little further to get to the bottom if they are there for one of the  secondary (I presume) use cases. But that is just my opinion 🙃

neuromusic:
that makes sense! let's stick w/ 15

potating-potato:
I can't review code change so I tried the new pagination and it is working great.

I do wonder if the pagination should be reset to first page if the filter criteria changes? say the user is on page 3 and then decided that the filter criteria is not great and modified it. Should the pagination be brought back to first page in this scenario?

codemonkey800:
> I can't review code change so I tried the new pagination and it is working great.
> 
> I do wonder if the pagination should be reset to first page if the filter criteria changes? say the user is on page 3 and then decided that the filter criteria is not great and modified it. Should the pagination be brought back to first page in this scenario?

I was thinking the same thing too for filters, as well as the sort and search functionality. @neuromusic @liaprins-czi (I think her PTO started today though) any opinions on resetting the page when the filter, sort, or search changes?

neuromusic:
> resetting the page when the filter, sort, or search changes?

ah, good catch. yes, I think @potating-potato is right & a quick glance at implementations like amazon.com confirm that this is common.

frontend/src/components/PluginSearch/PluginSearch.tsx
```diff
@@ -14,31 +16,62 @@ import { PluginSearchResultList } from './PluginSearchResultList';
 const SCROLL_HANDLER_THROTTLE_MS = 200;
```
klai95, line 16: How did we come up with this value of 200?

codemonkey800: It's an arbitrary value that allows the scroll handler to be called frequently enough for it to be "instant" without it being spammy enough. Examples of scroll throttling teeter around 100ms - 300ms, so I just went with 200ms. 

We can always do some experimentation with different values, but I doubt the ROI will be worth it

PR #273, opened 2021-10-05
Fix pagination issues

codemonkey800:
## Description

Fixes issues found during QA:

> 1. when user clicks the next/back page... page remains at the bottom when it should go up every time user navigates the pagination
> 1. when page 1 or last page if displayed, if user positions the mouse next to the pagination, there is a hover state that should not be showing up
> 1. if there is only 5 pages, and in the search bar user enters page 7 for example, first page is displayed, instead of last page, in this case. page 5

## Demos

https://pagination-frontend.dev.imaging.cziscience.com/

### Pagination scrolls to top when changing pages

Before the page would stay at the bottom when navigating.

https://user-images.githubusercontent.com/2176050/136067488-ea55a891-7be5-4ea3-ab15-4ca11de32487.mp4

#### Before

highlight-state-before

#### After

highlight-state-after

### Open closest extreme for out-of-range pages

Before, the page would always start at page 1 if the page was out of range. Now, the page will be set to the closest extreme:

1. If the page is < 1 or not a number, then the page is automatically set to first page.
2. If the page is > `totalPages`, the page is set to the last page.

page-invalid-values

### DIFF
--- a/frontend/src/store/search/queryParameters.ts
+++ b/frontend/src/store/search/queryParameters.ts
@@ -1,4 +1,4 @@
-import { inRange, set } from 'lodash';
+import { set } from 'lodash';
 import { subscribe } from 'valtio';
 
 import { BEGINNING_PAGE, RESULTS_PER_PAGE } from '@/constants/search';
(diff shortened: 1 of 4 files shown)


frontend/src/store/search/queryParameters.ts
```diff
@@ -84,10 +84,15 @@ function initStateFromQueryParameters() {
       searchFormStore.search.index.length / RESULTS_PER_PAGE,
     );
 
-    searchFormStore.page =
-      Number.isNaN(page) || !inRange(page, BEGINNING_PAGE, totalPages + 1)
-        ? BEGINNING_PAGE
-        : page;
+    // If the page number is not a number or not in range, then set it to the
+    // beginning or ending page depending on which is closer.
+    if (Number.isNaN(page) || page < BEGINNING_PAGE) {
```
klai95, line 89: What if page > RESULTS_PER_PAGE - 1?

codemonkey800: I'm curious how `RESULTS_PER_PAGE` is related 🤔  The constant `RESULTS_PER_PAGE` refers to how many search results are on a page.

codemonkey800: Yes it should be the condition below `page > totalPages` for checking the right side of the range 😄

PR #274, opened 2021-10-05
Preview page infra setup

codemonkey800:
## Description

This refactors the [POC branch](https://github.com/chanzuckerberg/napari-hub/tree/03e63e64e3a6a4d370524b707a7dd8915b23741e) to be production ready. This configures the Hub's Next.js configuration to support building the preview page when the preview page environment variables are set.

### Building a preview page

To build the preview page, the following environment variables need to be set:

| Variable               | Description                                                                                                                                                                                                          |
| ---------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `PREVIEW`              | Path to JSON file with plugin data. This should match the structure of the backend response data. An example of how the data should be look an be seen in the [index.json file](./frontend/src/fixtures/index.json). |
| `PREVIEW_PULL_REQUEST` | A link to the pull request that created the current preview page.                                                                                                                                                    |

When these are set, a special build command can be used to build and export the the preview page as static HTML:

```sh
yarn build:preview
```

This will create the preview page as static HTML in the `preview/` directory.

## Demo

A demo of working build can be found in <INSERT PR HERE>


codemonkey800:
@DragaDoncila FYI, the JSON file your Python script will create will serve as input to the preview page build using the `PREVIEW` environment variable

klai95 (approved):
LGTM!

DragaDoncila:
> @DragaDoncila FYI, the JSON file your Python script will create will serve as input to the preview page build using the `PREVIEW` environment variable

ok that makes sense, I'm in the process of adding this functionality now and can produce the JSON.

What I'm still confused about is what triggers the production of this JSON - an API call to `/plugins/<plugin>` or to a different endpoint specifically for the preview page?

codemonkey800:
> > @DragaDoncila FYI, the JSON file your Python script will create will serve as input to the preview page build using the `PREVIEW` environment variable
> 
> @codemonkey800 ok that makes sense, I'm in the process of adding this functionality now and can produce the JSON.
> 
> What I'm still confused about is what triggers the production of this JSON - an API call to `/plugins/<plugin>` or to a different endpoint specifically for the preview page?

It depends on the GitHub integration we'll be using, but the trigger would be one of the following:

1. GitHub action workflow that gets notified when the PR changes (similar to the GitHub checks for this PR)
2. GitHub app that listens to a PR webhook for modifications.

The above events will be responsible for running the script, produce the plugin JSON, and build the preview page 😄

PR #276, opened 2021-10-05
Preview page app bar

codemonkey800:
## Description

Implements the frontend for the preview page app bar and metadata panel. 

## Demo

`<ADD DEMOS HERE>`


codemonkey800:
Tagging @liaprins-czi for if she wants to look at the PR when gets back 😄

klai95 (approved):

potating-potato:
The code is created alongside other "real" components, would it be a risk that we need to move this code to somewhere else if we decide to have a different deployment for the preview feature? for example, if you decide to go with a GitHub app as the outcome for the deployment exploration, would we need to move this code out?

codemonkey800:
> The code is created alongside other "real" components, would it be a risk that we need to move this code to somewhere else if we decide to have a different deployment for the preview feature? for example, if you decide to go with a GitHub app as the outcome for the deployment exploration, would we need to move this code out?

Since the designs for the preview specific components are fairly tightly coupled with the base website, I don't think we'll ever need to move the code outside of napari hub.

If we do use a different deployment approach for previews, then more likely we'll need to figure out a different interface for the napari hub preview page from a GitHub app 🤔

codemonkey800:
> also we should drop `Project Site` and `Twitter` from the checklist... these are optional according to the [original metadata spec](https://docs.google.com/spreadsheets/d/1p1JwKAE3NRelYwnD1g1L9iCAOtMJ9x-AHUoYaej6PPw/edit#gid=0), though we should clarify this on https://github.com/chanzuckerberg/napari-hub/blob/main/docs/customizing-plugin-listing.md

I think Lia added it so that plugin developer would know that it's missing, but the designs don't account for whether missing data is optional or required. With that in mind, would you still prefer to remove it?

I suppose later on we can always add it back to the preview page but present it differently so that it doesn't seem like we're indicating the page is failing validation because it's missing optional data.

PR #279, opened 2021-10-07
Add utilities to parse preview metadata from GitHub URL

DragaDoncila:
Add functions to clone repo, build wheel and parse metadata of plugin given a GitHub URL.
Reuses the backend utilities for parsing additional metadata not retrieved from PyPI.

Notes/For Discussion:
- Since we were able to build wheels for all the plugins on the napari hub, I have not included code that alternatively tries to build an sdist, because parsing the metadata from sdists is more complex.
- When trying to build the distribution I check the system for the executing prefix and run that version of Python. I initially did this when running locally because otherwise bin Python was used which is usually 2.x and can't really build wheels for any of our plugins. This might be overkill if we're running in a dedicated runner with its own Python
- Right now the tests use the `example-plugin` which is currently hosted on my own personal GitHub. I'm not sure if this is the best approach, and we may want to maintain an `example-plugin` in the chanzuckerberg org anyway, to test out new features against? @neuromusic thoughts? I do think it's important to have an example plugin repo to be testing as closely as possible to the real workflow
- The two fields `release_date` and `first_released` can only really be retrieved from PyPI. I'm not sure if we want to try to retrieve these for the preview page or if we just want to say something like "not available in preview". Pinging @neuromusic and @liaprins-czi on this one.

Still to do:
- [x] test parity with expected format for napari hub
- [x] figure out difference between `description` and `description_text` and update the metadata accordingly
- [x] check whether the napari hub parses extra requirements and have the preview parsing do the same
- [x] add code that actually runs the preview function

### DIFF
--- a/backend/get_preview.py
+++ b/backend/get_preview.py
@@ -1,5 +1,5 @@
 import os
-from .preview import get_plugin_preview
+from preview import get_plugin_preview
 import argparse
 
 parser = argparse.ArgumentParser(description='Write preview metadata to JSON.')
--- a/backend/preview/__init__.py
+++ b/backend/preview/__init__.py
@@ -1,3 +1,3 @@
-from .preview import get_plugin_preview
+from preview.preview import get_plugin_preview
 
 __all__ = ['get_plugin_preview']
--- a/backend/preview/preview.py
+++ b/backend/preview/preview.py
@@ -8,7 +8,7 @@ import glob
 import os
 import json
 
-from ..utils.github import github_pattern, get_github_metadata, get_github_repo_url
+from utils.github import github_pattern, get_github_metadata, get_github_repo_url
 
 def get_plugin_preview(repo_pth: str, dest_dir: str, is_local: bool=False) -> dict:
     """Get plugin preview metadata of package at repo_pth.


potating-potato:
from a functionality point, this part of the code doesn't quite belong to the backend part (assuming I read right that this is supposedly used from command-line instead of as an API?) we can create a new folder (maybe CLI?) under the root directory to contain this. I want to keep the backend package exclusively for the API-related stuff. 

(If this is designed to be integrated as an API one day, I think there can be trouble when we deploy this. There is limited disk space in AWS lambda, so pulling the repo from git and then compile it may take up a lot, even if we clean it up afterward may introduce risks of multiple requests concurrently may take the API container down)

DragaDoncila:
Re. moving the preview code to a separate directory.

Because we need to import functions from the `backend` directory (or package), we're going to run into trouble with imports.
The underlying reason is that Python won't do relative imports from outside the package.

Let's say I make a new top level folder called `preview` for the new functionality. Our folder structure would be:

```
napari-hub/
  backend/
    __init__.py
    github.py
    ...
  preview/
    __init__.py
    get_preview.py # we run this
    preview.py
  ...
```

Because `napari-hub` itself is not a package, we can't import from `backend/` into `preview/`. I'm not sure what the best option is here, and to be honest I'm no expert on this side of things.

Best practice would be to add a top level directory which serves as the root of all our python code, which has an `__init__.py` (marking it as a package) - we could then run our code as a module - same as we currently do with backend tests for the same reason. This would look like:

 ```
napari-hub/
  __init__.py
  python_code/
      backend/
        __init__.py
        github.py
        ...
      preview/
        _tests/
         ...
        __init__.py
        get_preview.py # we run this
        preview.py
  frontend/
    ...
  ...
```

But I'm not entirely sure how this change would affect the rest of the imports/code we're running, and whether it would affect the frontend.

A dodgier fix is to add the parent directory to the path (e.g. [this stack answer](https://stackoverflow.com/questions/11536764/how-to-fix-attempted-relative-import-in-non-package-even-with-init-py/11536794#11536794)). I have not been able to get this to work without breaking relative imports as we're currently using them in `backend/`.

I think my preferred fix would be to move all Python code to a well formed Python package, but I'm not sure if there are side effects I haven't considered. We also just recently did a big refactor of the backend so maybe these concerns have already been considered? Curious on your thoughts @potating-potato @codemonkey800

**Comment by potating-potato** [comment]
Referring to this for better practice: https://docs.python-guide.org/writing/structure/ 
It did have specialized import for the test suite: https://docs.python-guide.org/writing/structure/#test-suite
Though one can argue this pattern is only good for testing, I would personally think this is fine for one-way dependency. 

Alternatively, we can create an "src" folder and make that a package with its own __init__.py, so that modules can use this to absolute import 

I gave it a try and it seems to work correctly (you do have to tweak based on how you are planning to run it, ie: how is python path is populated at all)

**Comment by DragaDoncila** [comment]
@potating-potato ok great, I'll update this to a dev branch this afternoon so we can see any effects on frontend workflows, and fix anything I can find, and I'll let you know when it looks ready for a final review.

**Comment by DragaDoncila** [comment]
@potating-potato thanks for updating the imports. I've updated them in the `preview` package as well. @codemonkey800 finally, the way you get your JSON is running the following command from `napari-hub/backend` (it won't work if you're not in backend).

```
python -m get_preview <github url> <destination directory>
```

OR

```
python -m get_preview <path to repo> <destination directory> --local
```
if the plugin repository has already been cloned.

DragaDoncila:
@neuromusic I like that idea and I think it covers most use cases. My only concern then is whether we need to flag to the user in any way that this information is not final/may change depending on when they decide to release to PyPI? Or do we just assume they're aware of their own release schedule 😄

PR #308, opened 2021-10-30
Preview page styling

codemonkey800:
## Description

Updates the preview page styling to match the Figma designs.

## Demo
https://napari-hub-preview-styling.vercel.app/

image


codemonkey800:
@liaprins-czi I was unable to add the "arrow that follows the tooltip anchor" effect for this PR:

image

However, @kne42 is planning on implementing this so that we can use it for both the preview page and categories 😄

liaprins-czi (changes requested):
Overall this is looking really great to me! There are a few small things I think should be updated:
- The vertical spacing between metadata with the orange flags is too tall, it should be the same as it is without the orange flags: [link to Figma](https://www.figma.com/file/6lXYXXNi7bDPnbv8vmHdNZ/Plugin-page-preview?node-id=97%3A12112)
How it is implemented now:
Screen Shot 2021-11-01 at 1 57 31 PM
How it should loook:
Screen Shot 2021-11-01 at 1 57 23 PM

- There should be a little more space between the "View pull request" text and the arrow: should be 10px spacing
- The overview of the missing metadata that lists all the fields should be open by default
- The grey hover state of the chevron to open / close the overview listing all metadata snhould be a square, but it is a narrow rectanble instead:
Screen Shot 2021-11-01 at 1 52 45 PM

- When I viewed it in mobile / narrow-screen view (>495px wide), sometimes the tooltip was centered with the screen, but for "Author" it was shifted over to the left a little (this one is not a very big deal at all)
Screen Shot 2021-11-01 at 1 52 19 PM

- I know you implemented the little "!" and checkmark symbol colored squares all in a row at the top at the right size according to my designs, but to me they look a little too small when I see them in the real context, is it possible to bump them all up to 10 px square? That'd add about 28px to the overall horizontal distance they take up together, and i think there's space for them to still fit... sorry for askign for this; it is my mistake!

codemonkey800:
> When I viewed it in mobile / narrow-screen view (>495px wide), sometimes the tooltip was centered with the screen, but for "Author" it was shifted over to the left a little (this one is not a very big deal at all)

Do the titles on the page and tooltip have to align? it will be a bit tricky to implement but possible doable using CSS grids. let's defer this to the a later implementation because @kne42 is taking over the full tooltip implementation which includes the sharkfin thing

---
@liaprins-czi The styling issues have been addressed (I hope!). Could you take a look again please? 😄

liaprins-czi (approved):
Everything looks really good to me! Thanks!

klai95 (approved):

PR #312, opened 2021-11-07
Add category API

potating-potato:
* Add API for category:
`/category`: get all category mappings
`/category/<category>`: match a particular category
`/category/<category>/version/<version>`: match a particular category under a specific version

* Parse GitHub config for category metadata

* Add category curation scripts to generate the mapping file from scratch.
* Add curated mapping file for edam alpha06

API for dev environment: 
plugins/index endpoint format: (won't show on index page itself because this plugin is hidden)
https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/category-api/plugins/napari-demo

category api format:
https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/category-api/categories
https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/category-api/categories/Manual%20segmentation
https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/category-api/categories/2D%20image/versions/EDAM-BIOIMAGING:alpha06


backend/api/app.py
```diff
@@ -61,11 +64,22 @@ def get_exclusion_list() -> Response:
...
+def get_category(category: str, version: str) -> Response:
+    return jsonify(get_category_mapping(category, version=version))
+
+
 @app.errorhandler(404)
 def handle_exception(e) -> Response:
     links = [rule.rule for rule in app.url_map.iter_rules() if 'GET' in rule.methods
-             and (rule.rule.startswith("/plugins") or rule.rule.startswith("/shields"))]
-    return app.make_response((f"Invalid Endpoint, valid endpoints are {links}", 404,
+             and (rule.rule.startswith("/plugins")
```
klai95, line 77: `if all( ['GET' in rule.methods, rule.rule.startswith("/plugins"] ) 
  or any(rule.rule.startswith("/shields"), rule.rule.startswith("/category"))`

backend/api/model.py
```diff
@@ -236,3 +251,54 @@ def move_artifact_to_s3(payload, client):
...
+    -------
+    match : dict if no category is given or list of dict if matched
+        list of mapped label, dimension and hierarchy, where hierarchy is from most abstract to most specific.
+        for example, Manual segmentation is mapped to the following list:
+        [
+            {
+                "label": "Image Segmentation",
+                "dimension": "Operation",
+                "hierarchy": [
+                    "Image segmentation",
```
klai95, line 280: Does this mean that, in this case, `Image Segmentation` is the top-level parent, and `Manual Segmentation` is a leaf node?

backend/api/model.py
```diff
@@ -236,3 +251,54 @@ def move_artifact_to_s3(payload, client):
             pull_request = client.pull_request(owner, repo, pull_request_number)
             pull_request.create_comment('Preview page for your plugin is ready here:\n'
                                         f'https://preview.napari-hub.org/{owner}/{repo}/{pull_request_number}')
+
+
+def get_category_mapping(category: str = None, mappings: dict = None,
+                         version="edam-alpha06") -> Union[Dict[str, List], List[Dict]]:
```
klai95, line 257: Is there a reason we have version set to `edam-alpha06' as a static value?

backend/api/model.py
```diff
@@ -236,3 +251,54 @@ def move_artifact_to_s3(payload, client):
...
+                    "Image annotation",
+                    "Dense image annotation",
+                    "Manual segmentation"
+                ]
+            }
+        ]
+    """
+    if not mappings:
+        mappings = get_cache(f'category/{version}-mappings.json')
+    if not mappings:
```
klai95, line 297: Why is the `if not mappings` condition repeated here?

klai95, line 297: Why is the `if not mappings` condition repeated here?

backend/category/edam.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+
+    Parameters
+    ----------
+    version : str
+        version of edam ontology to get.
+
+    Returns
+    -------
+    Dict[str, List[str]]
```
klai95, line 17: Can you give a brief example of what Dict[str, List[str]] looks like for context?

backend/api/model.py
potating-potato: That is correct

backend/api/model.py
potating-potato: the is currently the default version, we can remove it when we have more versions

backend/api/model.py
potating-potato: when an invalid version is specified, the mapping returned would be None, in this scenario, we would return a static empty list

backend/category/edam.py
potating-potato: probably better to add as documentation here, will do

klai95 (approved):

backend/api/model.py
neuromusic: we should be more explicit here with our naming, as there are multiple EDAM ontologies, including "EDAM" (which is more broadly for the biosciences), EDAM Bioimaging (which we are using) and EDAM-geo (for geo & earth sciences which might be a candidate if we extend to other domains in the future).

`edam-bioimaging-alpha06` might be more appropriate

backend/api/model.py
```diff
@@ -122,6 +124,19 @@ def build_plugin_metadata(plugin: str, version: str) -> Tuple[str, dict]:
         metadata = {**metadata, **get_github_metadata(github_repo_url)}
     if 'description' in metadata:
         metadata['description_text'] = render_description(metadata.get('description'))
+    if 'category' in metadata:
```
neuromusic, line 127: this schema doesn't match the one spec'd in the PRD, which lets the plugin dev specify the ontology:

```
labels:
  ontology: "EDAM-BIOIMAGING:alpha06"
  terms:
  - 3D Image                 # Data Shape
  - Cell segmentation        # Processing step
  - Model-based segmentation # Processing step
  - Fluorescence Microscopy  # Imaging Modality
```

neuromusic (changes requested):
two concerns inline, otherwise looks great.

endpoints look 💯 . love the multi-hierarchy support. intuitive enough that they should serve (partially) as documentation for plugin devs

neuromusic (approved):
my concerns have been addressed :D

backend/utils/github.py
```diff
@@ -122,17 +118,23 @@ def get_github_metadata(repo_url: str, branch: str = 'HEAD') -> dict:
...
-    if 'Twitter' in project_urls:
-        github_metadata['twitter'] = project_urls['Twitter']
+    yaml_file = get_file(repo_url, ".napari/config.yml", branch=branch)
+    if yaml_file:
+        config = yaml.safe_load(yaml_file)
+        hub_config = {key: config[key] for key in hub_config_keys if key in config}
+        github_metadata.update(hub_config)
+
+        project_urls = config.get('project_urls', {})
+        if 'Project Site' in project_urls:
```
klai95, line 128: What do you think of `github_metadata_parameters = ['project_site', 'documentation', 'support', 'report_issues', 'twitter']
project_urls_parameters = ['Project Site', 'Documentation', 'User Support', 'Report Issues', 'Twitter']
index_list = [i for i, project_url_parameter in enumerate(project_urls_parameters) if project_url_parameter in project_urls]
if len(index_list):
    github_metadata[github_metadata_parameters[index_list[0]]] = project_urls[project_urls_parameters[index_list[0]]]`?

klai95, line 128: What about 
Screen Shot 2021-11-08 at 6 22 57 PM?

neuromusic: alternatively...

``` Python
mapping = {
    "Project Site": "project_site",
    "Documentation": "documentation",
    "User Support": "support",
    "Report Issues": "report_issues",
    "Twitter": "twitter",   
}

for param, url in project_urls.items():
    github_metadata[mapping[param]] = url
```

klai95: Does that mean in each instance of `project_urls`, `param` is the key and `url` is the value such that project_urls[param] == url?

neuromusic: yeah, exactly

you can iterate over those key:value pairs in `project_urls` with `.items()`... this outputs a tuple of `(key,value)` which is assigned to `param` and `url`

neuromusic: and by iterating over `project_urls`, the assignment only happens for project_urls which are defined... no need to check if the keys are present

although, now that I think of it, this will bork if there are stray keys in `project_urls` that aren't in `mapping`... in which case, could make `mapping` a defaultdict or add some try/except logic to the assignment in the for loop

klai95: Would this work?
``` Python
mapping = {
    "Project Site": "project_site",
    "Documentation": "documentation",
    "User Support": "support",
    "Report Issues": "report_issues",
    "Twitter": "twitter",   
}

for param, url in project_urls.items():
    if (param in mapping):
        github_metadata[mapping[param]] = url
```

neuromusic: that would do it.

and good call re: defaultdict

justinelarsen:
Some high-level comments and questions:

1. Do we need both category and category_hierarchy on the plugin specific API? It seems like the category_hierarchy would be enough and the category is just redundant. https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/category-api/plugins/napari-demo
2. Should the plugin index API contain category info to build the search page listing? https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/category-api/plugins/index
3. Are these /category end points exposed for public use or are they just for internal consumption? I'm trying to understand why these are needed versus just adding the data returned to the existing /plugins end points. If they are necessary, I would rename "/category" to "/categories" to keep consistent with the "/plugins" API and because it's returning multiple.

klai95:
> Some high-level comments and questions:
> 
> 1. Do we need both category and category_hierarchy on the plugin specific API? It seems like the category_hierarchy would be enough and the category is just redundant. https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/category-api/plugins/napari-demo
> 2. Should the plugin index API contain category info to build the search page listing? https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/category-api/plugins/index
> 3. Are these /category end points exposed for public use or are they just for internal consumption? I'm trying to understand why these are needed versus just adding the data returned to the existing /plugins end points. If they are necessary, I would rename "/category" to "/categories" to keep consistent with the "/plugins" API and because it's returning multiple.

That is a good call out to refactor `/category` to `/categories`. Would it make sense for `@app.route('/category/<category>/version/<version>')` to be refactored to `@app.route('/categories/<category>/versions/<version>')`?

backend/api/model.py
```diff
@@ -236,3 +252,53 @@ def move_artifact_to_s3(payload, client):
...
+        name of the category to map
+    mappings: dict
+        mappings to use for lookups, if None, use cached mapping
+
+    Returns
+    -------
+    match : dict if no category is given or list of dict if matched
+        list of mapped label, dimension and hierarchy, where hierarchy is from most abstract to most specific.
+        for example, Manual segmentation is mapped to the following list:
+        [
```
justinelarsen, line 275: Thanks for adding return examples! It makes it a lot easier to visualize the real data.

backend/category/edam.py
```diff
@@ -0,0 +1,132 @@
...
+            "Semi-automatic segmentation": [
+                "Image segmentation"
+            ],
+            "Sparse image annotation": [
+                "Image annotation"
+            ]
+        }
+    """
+    if version:
+        url = f'https://edamontology.org/EDAM-bioimaging_{version}.csv'
```
justinelarsen, line 32: Consider using a constant and not repeating the URL for easier maintenance

justinelarsen (approved):

justinelarsen:
> > Some high-level comments and questions:
> > 
> > 1. Do we need both category and category_hierarchy on the plugin specific API? It seems like the category_hierarchy would be enough and the category is just redundant. https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/category-api/plugins/napari-demo
> > 2. Should the plugin index API contain category info to build the search page listing? https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/category-api/plugins/index
> > 3. Are these /category end points exposed for public use or are they just for internal consumption? I'm trying to understand why these are needed versus just adding the data returned to the existing /plugins end points. If they are necessary, I would rename "/category" to "/categories" to keep consistent with the "/plugins" API and because it's returning multiple.
> 
> That is a good call out to refactor `/category` to `/categories`! Would it make sense for `@app.route('/category/<category>/version/<version>')` to be refactored to `@app.route('/categories/<category>/versions/<version>')`?

We should probably be consistent with "/versions" as well then to match the plugins end point.
Proposal would then be:
/category -> /categories
/category/<category> -> /categories/<category> 
/category/<category>/version/<version> -> /categories/<category>/versions/<version>

To match existing plugin end points:
/plugins
/plugins/<plugin>
/plugins/<plugin>/versions/<version>

backend/category/edam.py
klai95: What about this approach? This approach essentially uses a ternary operator to yield the value to inject in %s, such that it is either an empty string or "_{version}". For example, if `version == "1.1"`, then `url = https://edamontology.org/EDAM-bioimaging_1.1.csv`. If `version == ""`, then `url = https://edamontology.org/EDAM-bioimaging.csv`

`url = "https://edamontology.org/EDAM-bioimaging%s.csv" % ("" if len(version) == 0 else f'_{version}')`

backend/api/model.py
```diff
@@ -236,3 +252,53 @@ def move_artifact_to_s3(payload, client):
...
+                "dimension": "Operation",
+                "hierarchy": [
+                    "Image annotation",
+                    "Dense image annotation",
+                    "Manual segmentation"
+                ]
+            }
+        ]
+    """
+    if not mappings:
```
klai95, line 295: Could you elaborate the logic of the series of `if` and `else` statements from `line 295` to `line 304`  for context?

potating-potato:
> 1. Do we need both category and category_hierarchy on the plugin specific API? It seems like the category_hierarchy would be enough and the category is just redundant. https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/category-api/plugins/napari-demo

The duplication is an attempt to make frontend easier by providing the exact data as required for the figma design. Yes we could drop the category and only use hierarchy, but frontend code would have to do one more level of querying when user tries to filter on the plugin. But if this is acceptable, I have no problem dropping the category (and/or rename the hierarchy)

> 2. Should the plugin index API contain category info to build the search page listing? https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/category-api/plugins/index

Yes it should, right now the napari-demo plugin is hidden so it does not show up in the search page, it would show up for other public plugins.

> 3. Are these /category end points exposed for public use or are they just for internal consumption? I'm trying to understand why these are needed versus just adding the data returned to the existing /plugins end points. If they are necessary, I would rename "/category" to "/categories" to keep consistent with the "/plugins" API and because it's returning multiple.

These endpoints would be public so that people can look at our internal mappings and query how ontology terms can be mapped to hub terms. The preview page serves some of the purpose, but this gives a more proactive approach to see how category works

backend/api/model.py
potating-potato: I think this method may have been overloaded too much in logic, I can change this

PR #316, opened 2021-11-09
Category UI state

codemonkey800:
Sets up the UI state necessary for storing category and category hierarchy data from the backend, and reading that data for filtering and rendering.

The TypeScript interfaces for the category data were taken from the `napari-demo` plugin:

https://api.staging.napari-hub.org/plugins/napari-demo

```json
  "category": {
    "Supported data": [
      "2D", 
      "3D"
    ], 
    "Workflow step": [
      "Image Segmentation", 
      "Image annotation"
    ]
  }, 
  "category_hierarchy": {
    "Supported data": [
      [
        "2D"
      ], 
      [
        "3D"
      ]
    ], 
    "Workflow step": [
      [
        "Image Segmentation", 
        "Manual segmentation"
      ], 
      [
        "Image annotation", 
        "Dense image annotation", 
        "Manual segmentation"
      ], 
      [
        "Image Segmentation", 
        "Semi-automatic segmentation"
      ]
    ]
  }, 
```

### DIFF
--- a/frontend/src/store/search/filters.test.ts
+++ b/frontend/src/store/search/filters.test.ts
@@ -2,7 +2,12 @@ import { defaultsDeep } from 'lodash';
 import { DeepPartial } from 'utility-types';
 
 import pluginIndex from '@/fixtures/index.json';
-import { DeriveGet, PluginIndexData } from '@/types';
+import {
+  DeriveGet,
+  PluginCategory,
+  PluginCategoryHierarchy,
+  PluginIndexData,
+} from '@/types';
 
 import { filterResults } from './filters';
 import { DEFAULT_STATE, SearchFormStore } from './form.store';
@@ -49,6 +54,23 @@ function getLicenseResults(...licenses: string[]): SearchResult[] {
   const plugins = licenses.map((license) => ({
     ...pluginIndex[0],
     license,
+  }));
+
+  return getResults(...plugins);
+}
+
+interface CategoryResultData {
+  category?: PluginCategory;
+  category_hierarchy?: PluginCategoryHierarchy;
+}
+
+function getCategoryResults(
+  ...categoryData: CategoryResultData[]
+): SearchResult[] {
+  const plugins = categoryData.map(({ category, category_hierarchy }) => ({
+    ...pluginIndex[0],
+    category,
+    category_hierarchy,
   }));
 
   return getResults(...plugins);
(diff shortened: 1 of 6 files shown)


frontend/src/store/search/filters.test.ts
```diff
@@ -54,6 +59,23 @@ function getLicenseResults(...licenses: string[]): SearchResult[] {
   return getResults(...plugins);
 }
 
+interface CategoryResultData {
+  category?: PluginCategory;
+  category_hierarchy?: PluginCategoryHierarchy;
```
klai95, line 64: What is the difference between `category` and `category_hierarchy` both in definition and how frontend will make use of them?

frontend/src/store/search/filters.test.ts
```diff
@@ -239,4 +261,115 @@ describe('filterResults()', () => {
       expect(filtered).toEqual(getLicenseResults('valid'));
     });
   });
+
+  describe('filter by workflow step', () => {
+    const results = getCategoryResults(
```
klai95, line 266: Very nice test case!

frontend/src/store/search/filters.test.ts
```diff
@@ -239,4 +261,115 @@ describe('filterResults()', () => {
...
+        results,
+      );
+      expect(filtered).toEqual(results.slice(0, 2));
+    });
+  });
+
+  describe('filter by image modality', () => {
+    const results = getCategoryResults(
+      {
+        category: {
```
klai95, line 305: Since we have duplicate code `        category: {
          'Workflow step': ['foo', 'bar'],
        },
      },
      {
        category: {
          'Workflow step': ['bar'],
        },
      },
      {
        category: {
          'Image modality': ['foo', 'bar'],
        },` can we declare that somewhere above these two tests and then these two tests can use them?

frontend/src/store/search/filters.test.ts
```diff
@@ -239,4 +261,115 @@ describe('filterResults()', () => {
...
+    it('should filter plugins with matching workflow steps', () => {
+      const filtered = filterResults(
+        createMockFilterGet({
+          imageModality: {
+            bar: true,
+          },
+        }),
+        results,
+      );
+      expect(filtered).toEqual(results.slice(2, 3));
```
klai95, line 335: is `results.slice(2, 3)` equivalent to `results[2]`?

frontend/src/store/search/form.store.ts
```diff
@@ -1,9 +1,10 @@
 import { over } from 'lodash';
+import { DeepPartial } from 'utility-types';
```
klai95, line 2: Is `DeepPartial` used in this file?

frontend/src/store/search/filters.test.ts
codemonkey800: Yeah! i'll simplify the code!

frontend/src/store/search/filters.test.ts
codemonkey800: I'm not sure what you mean in this context, but this interface is only used for testing so it's not really used by the frontend

frontend/src/store/search/filters.test.ts
codemonkey800: actually, it's supposed to be `[results[2]]`, but it's still shorter than the actual code

frontend/src/store/search/form.store.ts
codemonkey800: nope! but the linter found that for me already but thank you lol

frontend/src/types.ts
```diff
@@ -8,19 +8,32 @@ export interface PluginAuthor {
...
-
 export interface CitationData {
   citation: string;
   RIS: string;
   BibTex: string;
   APA: string;
 }
 
+export type HubDimension =
+  | 'Workflow step'
```
klai95, line 19: Just to clarify, `Workflow step` is the internal value we store, but `Workflow Step` is the displayed value on UI?

frontend/src/types.ts
```diff
@@ -8,19 +8,32 @@ export interface PluginAuthor {
 /**
  * Citation formats.
  * */
-
```
klai95, line 11: Is this line deleted by linter?

frontend/src/store/search/form.store.ts
```diff
@@ -115,6 +119,38 @@ export function initOsiApprovedLicenseSet(licenses: SpdxLicenseData[]): void {
...
+
+/**
+ * Initialization function that sets up the category filters using the provided
+ * data. This is necessary so that the filters only include a subset of terms
+ * that are included in the current plugin ecosystem.
+ *
+ * @param index The plugin index.
+ */
+export function initCategoryFilters(index: PluginIndexData[]): void {
+  for (const plugin of index) {
```
klai95, line 139: Would the triple for loop become a performance bottleneck? I am not sure if there is a way to get around it, but just wanted to get some clarity on this!

frontend/src/types.ts
codemonkey800: no i deleted it lol

frontend/src/types.ts
codemonkey800: No, it's also `Workflow step`

frontend/src/store/search/form.store.ts
codemonkey800: Great feedback! 🤩 

It could be if the amount of categories and category keys increase a lot. That would be basically `O(NMK)` 🤣  I can't think of a better way at the moment since:

1. We need to iterate the entire list of `N` plugins.
2. We need to iterate up to `M` categories.
3. We need to iterate over all `K` keys so we can populate the form state.

There's no way around this because we need to figure out the entire set of keys possible, and the only way to do that is by looking at every key for every category for every plugin.

I'm not too worried also because our existing category datasets are small, and we are eventually planning to move the search to the backend so that user's don't have to compute this on their devices

klai95 (approved):
Nicely done!

PR #319, opened 2021-11-12
attempt to read fork PR number when pull request does not present

potating-potato:
attempt to read fork PR number when pull request does not present

This fix is to address pull_requests being none in the payload when the PR is from a fork

### DIFF
--- a/backend/api/model.py
+++ b/backend/api/model.py
@@ -235,7 +235,14 @@ def move_artifact_to_s3(payload, client):
     repo = get_attribute(payload, ["repository", "name"])
     pull_request_number = get_attribute(payload, ['workflow_run', 'pull_requests', 0, 'number'])
     if not pull_request_number:
-        return
+        repo = client.repository(owner, repo)
+        head_owner = get_attribute(payload, ['workflow_run', 'head_repository', 'owner', 'login'])
+        head_branch = get_attribute(payload, ['workflow_run', 'head_branch'])
+        pull_requests = list(repo.pull_requests(head=f'{head_owner}:{head_branch}'))
+        if len(pull_requests) == 1:
+            pull_request_number = pull_requests[0].number
+        else:
+            return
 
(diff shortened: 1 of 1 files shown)


potating-potato:
@DragaDoncila it seems the test is failing because the test repo has changed 😅

DragaDoncila:
> @DragaDoncila it seems the test is failing because the test repo has changed 😅

yeah I worried about this happening, maybe we want to fork it over to `chanzuckerberg` and change the tests to point to the fork? Or we could fork your `napari-demo` instead and base the tests on that? I think it would be good to have a semi-independent plugin in `chanzuckerberg` org for testing purposes.

DragaDoncila:
@potating-potato I updated `main` to hopefully fix the backend tests if you want to rerun them.

I tried pointing to this branch and [updating the PR I had](https://github.com/DragaDoncila/example-plugin/pull/3), as well as [opening a new one](https://github.com/DragaDoncila/example-plugin/pull/6), but the bot still doesn't seem to be posting comments.

potating-potato:
> I tried pointing to this branch and [updating the PR I had](https://github.com/DragaDoncila/example-plugin/pull/3), as well as [opening a new one](https://github.com/DragaDoncila/example-plugin/pull/6), but the bot still doesn't seem to be posting comments.

this change won't be applied until deployed in production, because the github app is pointed to api.napari-hub.org

potating-potato:
> yeah I worried about this happening, maybe we want to fork it over to `chanzuckerberg` and change the tests to point to the fork? Or we could fork your `napari-demo` instead and base the tests on that? I think it would be good to have a semi-independent plugin in `chanzuckerberg` org for testing purposes.

we can use the chanzuckerberg repo (maybe even a dedicated branch), I can mark a story in our quality sprint

DragaDoncila (approved):
> this change won't be applied until deployed in production, because the github app is pointed to api.napari-hub.org

Ah great. In that case this looks great and I think the approach is good.

PR #323, opened 2021-11-16
Category chip

codemonkey800:
## Description

Implements the `CategoryChip` component to render categories / category hierarchies. This also installs the [sci-components](https://github.com/chanzuckerberg/sci-components) package so that we can leverage the science design system and also evaluate it at the same time.

## Demo

https://cat-chip-frontend.dev.imaging.cziscience.com/chip-test

image


codemonkey800:
This is linked to the wrong Shortcut ticket, I'll need to rename the branch and reopen the PR 😢

PR #324, opened 2021-11-16
Category chip

codemonkey800:
## Description

Implements the `CategoryChip` component to render categories / category hierarchies. This also installs the [sci-components](https://github.com/chanzuckerberg/sci-components) package so that we can leverage the science design system and also evaluate it at the same time.

## Demo

https://cat-chip-frontend.dev.imaging.cziscience.com/chip-test

image


liaprins-czi (changes requested):
This looks wonderful! I think the visual styling looks just like specified. There are just a couple of things on the interaction side that I think we can modify slightly:
- [ ] Can we expland the hoverable area to extend around the "i" icon so it reaches out to the end (see magenta circle in screenshot below). This way the user has more area to hover.
Screen Shot 2021-11-16 at 6 08 31 AM

- [ ] The way the tooltip disappears is kind of jarring and seems like there is a lag... can we make it disappear as soon as the user's mouse exits the hoverable zone.
- [ ] The tooltip should have a grey border around it

This is looking so good! Thank you!

codemonkey800:
@liaprins-czi thank you for the feedback! i'll start working on fixes. as for the tooltip, i forgot to mention that @kne42 is currently working on an implementation that we'll be able to use for both the preview page and categories, so i won't be working on styling at the moment (but i will look into the tooltip disappearance timing) 😄

codemonkey800:
@klai95 this implements the chip you'll be using for https://app.shortcut.com/imaging/story/172045/category-pills-for-search-plugin-pages

klai95 (approved):

codemonkey800:
hey @liaprins-czi i've pushed my fixes! when you get a chance, could you take a look at it pls?

https://cat-chip-frontend.dev.imaging.cziscience.com/chip-test

liaprins-czi (approved):
Looks great to me! Thanks!

PR #329, opened 2021-11-16
correct preview project url to use Bug Tracker from setup configuration

potating-potato:
according to our own [documentation](https://github.com/chanzuckerberg/napari-hub/blob/main/docs/customizing-plugin-listing.md), we first look at the .napari/config.yml, and use Report Issues if present, if not we would read the setup config Bug Tracker


codemonkey800 (approved):
awesome!

PR #330, opened 2021-11-17
update workflow name matching to remove constraint

potating-potato:
relief constraint in workflow name matching to use file name matches instead


DragaDoncila (approved):
Looks fine to me though I'm not particularly familiar with the different response payloads. We can only test this after release to prod right?

PR #320, opened 2021-11-15
remove development status from metadata

kne42:
Screen Shot 2021-11-15 at 11 32 23 AM

Screen Shot 2021-11-15 at 11 32 13 AM

PR #343, opened 2021-11-23
Category filters

codemonkey800:
## Description

- Refactors the filters UI to use the [multi-select complex filter component](https://chanzuckerberg.github.io/sci-components/?path=/story/complexfilter--multiple-select-with-search) from the [Science Design System](https://github.com/chanzuckerberg/sci-components).
- Adds new filters for the plugin category data.

## Demos

See #342.


frontend/src/components/PluginSearch/PluginComplexFilter.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+    searchFormStore.filters.operatingSystems.linux = true;
+    searchFormStore.filters.operatingSystems.mac = true;
+    searchFormStore.filters.operatingSystems.windows = true;
+
+    const component = renderFilter();
+    expect(component.asFragment()).toMatchSnapshot();
+  });
+
+  it('should render filter options on initial load', () => {
+    searchFormStore.filters.operatingSystems.linux = true;
```
klai95, line 31: Should `searchFormStore.filters.operatingSystems.mac = true;` also be included here?

frontend/src/components/PluginSearch/PluginComplexFilter.tsx
```diff
@@ -0,0 +1,138 @@
...
+ * Labels to use for a particular filter option.
+ */
+const FILTER_OPTION_LABELS: Record<string, string | undefined> = {
+  // Operating System
+  linux: 'Linux',
+  mac: 'macOS',
+  windows: 'Windows',
+
+  // Development Status
+  stable: 'Only show stable plugins',
```
klai95, line 31: What does `stable` mean in the context of plugins?

codemonkey800: Good find! It means only plugins that have a development status that indicates that the plugin is stable, but we actually removed this a while ago in #320. I'll just remove this then 😆

frontend/src/components/PluginSearch/PluginComplexFilter.test.tsx
codemonkey800: nah it doesn't, it's only test data. i could add it or not, and it won't affect the outcome of the test lol

frontend/package.json
```diff
@@ -50,7 +50,7 @@
     "axios": "^0.23.0",
     "cheerio": "^1.0.0-rc.10",
     "clsx": "^1.1.1",
-    "czifui": "^1.2.1",
+    "czifui": "^1.5.0",
```
klai95, line 53: Clarifying question: how do we know what version we are at and when to update to the next version?

codemonkey800:
E2E tests will be fixed in https://github.com/chanzuckerberg/napari-hub/pull/348

frontend/package.json
codemonkey800: You can check the version in the `package.json` file (which is what I'm updating actually lol), and there's no hard rule on when to update. I usually like to keep my dependencies up-to-date, so I say whenever possible. However, sometimes updates take refactoring time, so it may be more effort than for what it's worth at the time (unless it fixes a security vulnerability).

In this case, upgrading was simple and required not refactoring 😄

klai95 (approved):

PR #346, opened 2021-11-29
fix bug where variable override during pull request number look up

potating-potato:
fix bug where variable override during pull request number lookup 

the variable repo was accidentally overridden and resulted in lookup failure in code below

### DIFF
--- a/backend/api/model.py
+++ b/backend/api/model.py
@@ -232,10 +232,10 @@ def move_artifact_to_s3(payload, client):
     repo = get_attribute(payload, ["repository", "name"])
     pull_request_number = get_attribute(payload, ['workflow_run', 'pull_requests', 0, 'number'])
     if not pull_request_number:
-        repo = client.repository(owner, repo)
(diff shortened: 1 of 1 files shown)


backend/api/model.py
```diff
@@ -232,10 +232,10 @@ def move_artifact_to_s3(payload, client):
     repo = get_attribute(payload, ["repository", "name"])
     pull_request_number = get_attribute(payload, ['workflow_run', 'pull_requests', 0, 'number'])
     if not pull_request_number:
-        repo = client.repository(owner, repo)
+        github_repo = client.repository(owner, repo)
```
klai95, line 235: Is this code change. just for the sake of refactoring the variable name?

codemonkey800: there's another variable `repo` on line 232, and i think this assignment is causing the issue lol

codemonkey800 (approved):
LGTM 👍

PR #347, opened 2021-11-29
Clear all filters button

codemonkey800:
## Description

Implements the `Clear All` filters button. To get this working, I had to refactor the state to be created in a provider. This is so that:

1. State is available on initial render.
    - Currently, the state is initialized after the first render, but this causes issues with the new filter component because of how it handles controlled state. 
2. We don't pollute the global state for other connections.
    - Now that we're passing the state in a provider, the module state is being initialized on the server.
    - The current implementation creates state at the module level, which is shared between connections to the server.
    - If user A opens https://napari-hub.org/?query=segmentation, this will set the query state to `segmentation`. Then when user B opens `https://napari-hub.org`, the rendering will break for them and show a query when there shouldn't be

## Demo

https://clear-all-frontend.dev.imaging.cziscience.com

clear-all-button


liaprins-czi (approved):
Works and looks good to me! Thanks!

codemonkey800:
Tests will be fixed in https://github.com/chanzuckerberg/napari-hub/pull/348

PR #348, opened 2021-11-29
Fix unit and E2E tests for filters

codemonkey800:
Updates the test files to work with the new filter components.


klai95 (approved):
Looks good to me!

PR #353, opened 2021-12-02
Categories filter feature flag

codemonkey800:
## Description

Implements basic feature flagging architecture for hiding / showing features. This will allow us to merge code for new features into production without releasing the feature to the public.

The only feature flag I added is `categoryFilters`, which will hide the category filter components added in #343. The timeline for our current release schedule plans for us to release the category filters on January 13th, so the feature will be disabled for `staging` and `prod` until then.

## Enabling / Disabling a feature flag

Feature flags can be enabled by adding new environments to the `FEATURE_FLAGS` constant, or by using the query parameters `enable-feature=<feature-key` and `disable-feature=<feature-key>`.

## Future Work

The Aspen team is looking into integrating with a full fledged feature flag (FFFF) service that we may want to look into in the future. I've made this implementation as simple and lightweight as possible so it would be easy to refactor if we do end up switching.

### Category filters enabled

https://filter-ff-frontend.dev.imaging.cziscience.com/?enable-feature=categoryFilters

### Category filters disabled

https://filter-ff-frontend.dev.imaging.cziscience.com/?disable-feature=categoryFilters

### DIFF
--- a/frontend/src/utils/featureFlags.ts
+++ b/frontend/src/utils/featureFlags.ts
@@ -1,0 +1,78 @@
+import { PROD, STAGING } from '@/env';
+
+import { createUrl } from '.';
+
+export type FeatureFlagEnvironment = 'dev' | 'staging' | 'prod';
+
+export interface FeatureFlag {
+  environments: FeatureFlagEnvironment[];
+}
+
+/**
+ * Helper to preserve the keys in the type while forcing each value to be of
+ * type `FeatureFlag`.
+ */
+function createFeatureFlags<T>(flags: { [key in keyof T]: FeatureFlag }) {
+  return flags;
+}
+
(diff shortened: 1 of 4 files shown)


klai95 (approved):
Nicely done!

frontend/src/utils/featureFlags.ts
```diff
@@ -0,0 +1,78 @@
...
+    if (enabledFeatures.has(key)) {
+      return true;
+    }
+
+    if (disabledFetures.has(key)) {
+      return false;
+    }
+  }
+
+  if (PROD) {
```
klai95, line 69: Just a suggestion, but we could maybe use case statements here? There are also only 3 conditions though, so maybe it would not be necessary

PR #358, opened 2021-12-10
Fixes color for filter pills

codemonkey800:
## Description

This fixes the incorrect color showing up for the filter pills. The issue was caused by a server mismatch because of code running on the browser, but not the server. The issue is that the feature flag function was parsing the query parameters only on the client, so the server would render the HTML for the category filters and the client would not render it.

This caused React hydration step to behave weirdly by keeping some of the CSS classes on the server still attached in the client, but to the wrong element.

The fix is to make the feature flag function universal, so I refactored it into a hook that uses the Next.js router path. That way, both the server and client render the same HTML for a particular feature flag.

### DIFF
--- a/frontend/src/components/PluginSearch/PluginFilterByForm.tsx
+++ b/frontend/src/components/PluginSearch/PluginFilterByForm.tsx
@@ -6,7 +6,7 @@ import { Accordion } from '@/components/common/Accordion';
 import { Media } from '@/components/common/media';
 import { useSearchStore } from '@/store/search/context';
 import { FilterKey } from '@/store/search/search.store';
-import { isFeatureFlagEnabled } from '@/utils/featureFlags';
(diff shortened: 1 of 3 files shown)


klai95 (approved):
Looks good to me! I only have one question: why did the variable name change from `isFeatureFlagEnabled` to `useIsFeatureFlagEnabled`?

codemonkey800:
> Looks good to me! I only have one question: why did the variable name change from `isFeatureFlagEnabled` to `useIsFeatureFlagEnabled`?

@klai95 it's a naming convention for React hooks: https://stackoverflow.com/q/66151248

Technically, we don't have to rename it, but then our static code analyzer will not recognize it as a hook and so we won't be able to lint for errors against [rules of hooks](https://reactjs.org/docs/hooks-rules.html)

PR #363, opened 2021-12-14
Tooltip Copy

codemonkey800:
## Description

This adds the copy for each "workflow step" and "image modality" category.

## Demo

https://tooltip-copy-frontend.dev.imaging.cziscience.com


codemonkey800:
@liaprins-czi could you review the designs for this please? 

@neuromusic tagging you in case you want to check the copy 😄

neuromusic (dismissed):
Two content changes, otherwise LGTM

codemonkey800:
@liaprins-czi one thing I want to call out is the tooltip arrow is not exactly as designed. I'll save that for a future PR since it needs a custom implementation, but I've tried approximating it using what Material UI already provides:

image

liaprins-czi (changes requested):
This is looking really nice! Just a couple of things to update, from a design POV:
- The hoverable zone that triggers the tooltip should be limited to [just the "i" icon + surrounding radius equal to its padding](https://www.figma.com/file/1FVOiniqeJtTP4fjZZm4aO/Categories?node-id=475%3A471) (this prevents the tooltip from popping up when user unintentionally mouses over the tag in general, only when they intentionally mean to hover on the "i" area):
Screen Shot 2021-12-14 at 3 57 28 PM

- The hover color of the tag is too dark relativ to the non-hover color. The regular (non-hover color should be ##ECF8FF (which it is currently) and the hover state color should be ##D2EFFF (currenlty it is the darker blue color used on th hub).
- If possible, when hovering on a tag, can the light grey hover state of the plugin overall remove? Because if you click when hovering on a tag, it applies it as a filter (which it should), and does not go to the plugin page, which is what happens when you click anywhere else within the plugins card (not the blue tags though). So if we remove the grey hover state when hovering on a tag, it will help to indicate that clicking on the tag will _not_ take you to the plugin page.

Looking good!

codemonkey800:
@liaprins-czi I've pushed a commit addressing your feedback, please take a look when you get a chance 😄 

https://tooltip-copy-frontend.dev.imaging.cziscience.com/

---

> If possible, when hovering on a tag, can the light grey hover state of the plugin overall remove?

This was a fun one. It's actually _not_ possible to do this in pure CSS: https://stackoverflow.com/a/17924223
However, I was able to implement the effect with JavaScript, but very interesting to run into a limitation of CSS 😆 

This would be great for a short casual tech talk in the future lol

liaprins-czi:
> This would be great for a short casual tech talk in the future lol

That would be interesting! Thanks for looking into how to make this work!

liaprins-czi:
> one thing I want to call out is the tooltip arrow is not exactly as designed. I'll save that for a future PR since it needs a custom implementation, but I've tried approximating it using what Material UI already provides

Thanks for noting this; it looks totally fine to me for now!!

liaprins-czi (approved):
Looks great to me!

PR #368, opened 2021-12-15
Fix unnecessary autofocus when opening the filter

codemonkey800:
Darn auto focus


justinelarsen (approved):
LGTM! Thanks!

potating-potato:
should we be worried about the e2e test?

PR #371, opened 2021-12-21
Hide keyboard for plugin filter

codemonkey800:
## Description

Upgrades the `czifui` package to hide the virtual keyboard when the plugin filter is opened.

### Before

Before this fix, the virtual keyboard would pop up even if there was no input element.

https://user-images.githubusercontent.com/2176050/146989844-fa585382-2d49-4e27-88fc-f45f17e0cdf8.mp4

### After

With the fix, the virtual keyboard will now only show up if there is an input element.

https://user-images.githubusercontent.com/2176050/146989858-01c4f5b3-4fe5-4ef9-b41c-f6a45d5515a5.mp4

## Future work

There is still the issue where the viewport jumps when focusing on the input. I'll continue working with the SDS team to figure out a fix for this.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -50,7 +50,7 @@
     "axios": "^0.23.0",
     "cheerio": "^1.0.0-rc.10",
     "clsx": "^1.1.1",
-    "czifui": "^1.5.0",
+    "czifui": "^1.5.6",
     "dayjs": "^1.10.7",
     "framer-motion": "^4.1.17",
     "fs-extra": "^10.0.0",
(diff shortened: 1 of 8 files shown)


codemonkey800:
`czifui@1.5.4` also introduced a bug where the state is no longer updated when clicking on the button. I've created a PR to fix that issue: https://github.com/chanzuckerberg/sci-components/pull/99

liaprins-czi (approved):
Works great now! One tiny thing that should be changed (can be independent of this PR / issue) from design POV is that the "names" of the filter sections ("SORT", "FILTER BY REQUIREMENT", etc) should be in all caps, but currently they are not. Thanks!

codemonkey800:
> Works great now! One tiny thing that should be changed (can be independent of this PR / issue) from design POV is that the "names" of the filter sections ("SORT", "FILTER BY REQUIREMENT", etc) should be in all caps, but currently they are not. Thanks!

ahh it looks like it's for mobile only, it's a small change so I'll add it on

PR #372, opened 2021-12-22
read file locally when github workspace is configured

potating-potato:
Address issues when working across forks. 

Use locally checkout files when working on github action instead trying to fetch it remotely

Configure preview page to use category API


codemonkey800 (approved):
awesome, LGTM! 😄

PR #382, opened 2022-01-07
Frontend i18n

codemonkey800:
## Description

This PR along with several other PRs implement i18n for the frontend of the napari hub. This works by using both [Next.js' Internationalized Routing Feature](https://nextjs.org/docs/advanced-features/i18n-routing), and the excellent i18n library [next-i18next](https://github.com/isaachinman/next-i18next) for managing i18n state.

## Release Plan

The implementation is split into separate PRs to make it easier to review. When all the PRs are reviewed and ready to merge, I'll merge them one by one into each other until all dependent PRs are merged into this one. Finally, I'll merge this PR to push the i18n feature to staging.

### Pull Requests

1. #382: (this PR) Adds locale + fixture data, and it removes dead code for the now removed development status filter.
1. #383: Sets up i18n infrastructure.
1. #384: Replaces all raw strings with i18n translations.

### English

https://i18n-test-frontend.dev.imaging.cziscience.com

If everything is working correctly, there should be no noticeable difference since it's an underlying change.

### Spanish

https://i18n-test-frontend.dev.imaging.cziscience.com/es

This version uses text translated using Google Translate for demonstration.

PR #383, opened 2022-01-07
Frontend i18n - next-i18next setup and helper utilities + components

codemonkey800:
## Description

Part 2 of 3 PRs for #382.

This PR sets up the infrastructure for i18n in the napari hub using `https://github.com/isaachinman/next-i18next`, introducing the following changes to the codebase:

### 1. i18n Type Definitions

Type definitions are provided for all i18n resources. This covers all valid JSON data types including strings, numbers, arrays, and objects. This works by creating an ambient module declaration that specifies the type definitions for each i18n resource and using type inference to determine the full type information of the JSON.

As a result, `useTranslation()` will provide the following affordances:

- Autocomplete when specifying a namespace:

- Autocomplete when specifying an i18n key:

- Correct return type when getting value:

### 2. i18n Component

Wrapper over the [Trans](https://react.i18next.com/latest/trans-component) component with automatic typing for all namespaces and i18n keys:

This component allows us to render i18n strings that have nested React components in them. For example:

```tsx
<I18n
  i18nKey="preview:appBar.previewMessage"
  tOptions={{
    context: hasRepoMismatch ? 'mismatch' : '',
  }}
  values={{
    codeRepoLink,
    codeRepoName,
    pullRequestLink: PR_LINK,
    previewRepoName: PREVIEW_LINK_TEXT,
    pluginName:
      metadata.name.value ||
      t('pluginData:labels.pluginName.preview'),
  }}
/>
```

### 3. i18n for MDX

In order for i18n to work with MDX, we need to have access to the locale information. Since this is only available for normal routes, I needed to refactor the MDX renderer to use [next-mdx-remote](https://github.com/hashicorp/next-mdx-remote) instead of the official [@next/mdx](https://www.npmjs.com/package/@next/mdx).

MDX is now rendered using a catch all route. It works by checking if an MDX file is located in the i18n directory. For example, if the user is visited `/es/about`, the catch all route will check `i18n/es/about.mdx`. This also works recursively, so we can add nested MDX pages.


frontend/next.config.js
```diff
@@ -49,11 +32,12 @@ if (PROD && PREVIEW) {
...
 }
 
-module.exports = withMDX({
+module.exports = {
   ...previewOptions,
+  i18n,
 
   basePath: process.env.BASE_PATH || '',
-  pageExtensions: ['ts', 'tsx', 'mdx'],
+  pageExtensions: ['ts', 'tsx'],
```
klai95, line 40: Can you provide high-level context why `mdx` is removed?

codemonkey800: MDX is removed because we moved to a different strategy for rendering MDX pages. Let me know if you need any more details! 

> In order for i18n to work with MDX, we need to have access to the locale information. Since this is only available for normal routes, I needed to refactor the MDX renderer to use next-mdx-remote instead of the official @next/mdx.
> 
> MDX is now rendered using a catch all route. It works by checking if an MDX file is located in the i18n directory. For example, if the user is visited /es/about, the catch all route will check i18n/es/about.mdx. This also works recursively, so we can add nested MDX pages.

kne42 (approved):
LGTM! Quick question - if the locale you're trying to reach for the new MDX doesn't exist, will it default to `en`?

codemonkey800:
> LGTM! Quick question - if the locale you're trying to reach for the new MDX doesn't exist, will it default to `en`?

@kne42 that's correct! a lot of the code where we reference the locale includes a default value. for example:

```ts
export async function getStaticProps({ locale = 'en' }: GetStaticPropsContext) {
```

PR #384, opened 2022-01-07
Frontend i18n - Refactor hub strings

codemonkey800:
## Description

Part 3 of 3 PRs for #382.

This PR refactors all raw strings used in the hub to use translations either via the `useTranslation()` or `I18n` component.


klai95 (commented):
Why are we refactoring hub strings?

codemonkey800:
> Why are we refactoring hub strings?

This is so we can use the i18n translation functions introduced in #382. Now instead of using raw, hardcoded strings like `Hello, World`, we're using something like `t('common:helloWorld)` where `'common:helloWorld'` is a key that refer to a string, so it's essentially a giant dictionary of strings.

The secret sauce is that when a user visits the napari hub in their language, the i18n library will automatically use text in the user's language if it's available: https://i18n-test-frontend.dev.imaging.cziscience.com/es

kne42 (approved):

PR #399, opened 2022-01-23
dynamo integration (replace backend data storage s3 with dynamodb)

potating-potato:
DRAFT PR

Replace our data backend with dynamodb to remove s3 dependency mostly (after this PR only preview pages are stored in s3, and they probably should be in s3 going forward)

The changes are mostly:
1. include terraform resource to manage new data tables
2. entity class with pynamodb to define these tables in python
3. updated model.py logic to use dynamodb instead of s3
4. some minor updates to other part of code base


potating-potato:
cleaning up git history and merge conflict

PR #424, opened 2022-02-09
API access time improvements

potating-potato:
according to our metrics, the head call takes longer than data fetch...


.happy/terraform/modules/api-gateway-proxy-stage/main.tf
```diff
@@ -16,11 +16,9 @@ resource aws_api_gateway_stage api_stage {
 resource aws_api_gateway_method_settings api_gateway_stage_setting {
   rest_api_id = var.rest_api_id
   stage_name  = aws_api_gateway_stage.api_stage.stage_name
-  method_path = "*/*"
+  method_path = "{proxy+}/GET"
```
potating-potato, line 19: enables cache on GET request, previously we do not override any configuration

codemonkey800 (approved):
LGTM!

PR #426, opened 2022-02-10
only enable cache for prod api

liu-ziyang:
This will speed up dev stack creation, as well as lower cost for our dev and staging environment


codemonkey800 (approved):
anything for more speed

klai95 (approved):
LGTM!

PR #432, opened 2022-02-24
correct cache path to use ~1 to represent root

potating-potato:
correct the path used to cache the API doc. The current one is incorrectly formatted

details see:
https://stackoverflow.com/questions/65807902/how-do-i-get-resource-path-for-root-resource-in-apigateway

### DIFF
--- a/.happy/terraform/modules/api-gateway-proxy-stage/main.tf
+++ b/.happy/terraform/modules/api-gateway-proxy-stage/main.tf
@@ -28,7 +28,7 @@ resource aws_api_gateway_method_settings root_cache {
   count = var.tags.env == "prod" ? 1 : 0
   rest_api_id = var.rest_api_id
   stage_name  = aws_api_gateway_stage.api_stage.stage_name
-  method_path = "/GET"
+  method_path = "~1/GET"
 
   settings {
     caching_enabled = true


.happy/terraform/modules/api-gateway-proxy-stage/main.tf
```diff
@@ -28,7 +28,7 @@ resource aws_api_gateway_method_settings root_cache {
   count = var.tags.env == "prod" ? 1 : 0
   rest_api_id = var.rest_api_id
   stage_name  = aws_api_gateway_stage.api_stage.stage_name
-  method_path = "/GET"
+  method_path = "~1/GET"
```
codemonkey800, line 31: what does this mean lol

codemonkey800 (approved):
lol LGTM, but what does it mean

.happy/terraform/modules/api-gateway-proxy-stage/main.tf
potating-potato: this is the escaped "/" in api gateway. I can't find it being documented anywhere either..

potating-potato: ah here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-apigateway-stage-methodsetting.html

> ResourcePath
The resource path for this method. Forward slashes (/) are encoded as ~1 and the initial slash must include a forward slash. For example, the path value /resource/subresource must be encoded as /~1resource~1subresource. To specify the root path, use only a slash (/). To apply settings to multiple resources and methods, specify an asterisk (*) for the HttpMethod and /* for the ResourcePath. This parameter is required when you specify a MethodSetting.

Issue #451, opened 2022-03-30
Plugin details metadata has double IDs 


codemonkey800:
#### Description

While debugging #364, I noticed that the plugin metadata had duplicate IDs:

#### Steps/Code to Reproduce

1. Visit https://www.napari-hub.org/plugins/napari-segment-blobs-and-things-with-membranes.
2. Open dev tools
3. Inspect `License` metadata
4. Observe HTML and notice it uses duplicate IDs

#### Expected Results

It should only have 1 ID.

#### Actual Results

<img width="547" alt="image" src="https://user-images.githubusercontent.com/2176050/160858135-9e9b3ab8-516f-4f98-aabc-3f3d20edca93.png">

Issue #364, opened 2021-12-15
[Desktop] Clicking on License on right side menu is not pointing to correct info in screen


lauramarcos:
#### Description
When clicking on License from right side menu, page does not point to that particular info

#### Steps/Code to Reproduce
1. Go to napari-hub page
2. Look for a plugin including "License" info (https://staging.napari-hub.org/plugins/napari-segment-blobs-and-things-with-membranes)
3. Click on "License" menu on right side
4. Pay attention to info displayed on screen

#### Expected Results
License info should be visible in screen

#### Actual Results
Other info than License is displayed on screen

https://user-images.githubusercontent.com/63799148/146242409-c82ba9bd-90fb-4a41-acc8-27a6f695d327.mov

neuromusic:
ohhh this is interesting... it looks like maybe there's an anchor for the "license" metadata section on the left sidebar that the link is jumping to?

codemonkey800:
> ohhh this is interesting... it looks like maybe there's an anchor for the "license" metadata section on the left sidebar that the link is jumping to?

looks like this is the issue, the anchors are conflicting with the user's markdown headers. One thing we can look into is adding a prefix to either the plugin metadata IDs or the user's header IDs

PR #453, opened 2022-03-31
Fix plugin metadata IDs

codemonkey800:
Fixes:

- https://github.com/chanzuckerberg/napari-hub/issues/364
- https://github.com/chanzuckerberg/napari-hub/issues/451

This works by adding a the prefix `metadata-` to plugin metadata IDs. 

FYI: @justinelarsen @neuromusic 
For review: @klai95 @kne42

### DIFF
--- a/frontend/src/components/PluginDetails/PluginMetadata.tsx
+++ b/frontend/src/components/PluginDetails/PluginMetadata.tsx
@@ -13,6 +13,7 @@ import {
 } from '@/components/MetadataList';
 import {
   Metadata,
+  MetadataId,
   MetadataKeys,
   usePluginMetadata,
   usePluginState,
@@ -120,7 +121,7 @@ function PluginMetadataBase({
 
     return (
       <MetadataList
-        id={key}
+        id={`metadata-${key}`}
         inline={inline}
         label={label}
         empty={!value}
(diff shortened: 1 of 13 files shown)


frontend/src/components/PluginDetails/PluginMetadata.tsx
```diff
@@ -157,7 +158,7 @@ function PluginMetadataBase({
       className="h-56"
       render={() => (
         <>
-          {renderSingleItemList('version', { highlight: false })}
+          {renderSingleItemList('version')}
```
codemonkey800, line 161: I noticed that the version wasn't being highlighted for some reason, I must have mistakenly added this `highlight: false` to this field. Since the plugin version is a part of the required metadata for the preview page, I re-enabled highlighting for this field.

frontend/src/context/plugin.tsx
```diff
@@ -275,3 +275,5 @@ export type MetadataKeys = keyof Omit<
   Metadata,
   'actionRepository' | 'workflowSteps' | 'imageModality' | 'supportedData'
 >;
+
+export type MetadataId = `metadata-${MetadataKeys}`;
```
codemonkey800, line 279: This uses the new template literal types from TypeScript 4.1:

https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html

Issue #433, opened 2022-03-02
search by author w/ link from plugin page


neuromusic:
### Summary

Let users easily search by author by clicking the author name on plugin pages.

### Intended outcome

When viewing the plugin page, users will be presented with links for authors. When clicked, they will be taken back to the search page to search for plugins by the author.

### How will it work

We will convert author names into hyperlinks with the author name populated as the search query string. To handle multiple authors specified in the PyPI author field, we will identify individual authors by splitting the author field using commons separators (comms, ampersand, "and", etc) and provide links for each author.

PR #456, opened 2022-04-04
search by author

neuromusic:
This is an attempt at implementing #433

### DIFF
--- a/frontend/src/components/MetadataList/MetadataListSearchItem.tsx
+++ b/frontend/src/components/MetadataList/MetadataListSearchItem.tsx
@@ -2,10 +2,10 @@ import clsx from 'clsx';
 import { ReactNode } from 'react';
 
 import { Link } from '@/components/common/Link/Link';
+
 import { usePlausible } from '@/hooks';
 import styles from './MetadataList.module.scss';
-// import { isExternalUrl } from '@/utils';
-// import { SEARCH_PAGE } from '@/store/search/constants';
+import { isExternalUrl } from '@/utils';
 
 interface Props {
   children: ReactNode;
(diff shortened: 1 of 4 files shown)


frontend/src/context/plugin.tsx
```diff
@@ -176,7 +176,7 @@ export function usePluginMetadata() {
       value:
         plugin?.authors && isArray(plugin.authors)
           ? plugin.authors
-              ?.map((author) => author.name)
+              ?.flatMap((author) => author.name.split(/[,;&]+/))
```
neuromusic, line 179: @codemonkey800 even easier than I expected

frontend/src/components/MetadataList/MetadataListSearchItem.tsx
```diff
@@ -0,0 +1,58 @@
...
+}
+
+/**
+ * Component for rendering text items in metadata lists.
+ */
+export function MetadataListSearchItem({ children, className }: Props) {
+  const plausible = usePlausible();
+
+  // not sure how to get text from children?
+  const url = 'https://napari-hub.org/?search=' + encodeURIComponent(children);
```
neuromusic, line 22: @codemonkey800 I probably shouldn't hardcode the domain here... what's the best way to get this to populate with the hostname? (e.g. localhost:8080 etc)

frontend/src/context/plugin.tsx
codemonkey800: amazing, i didn't know you could pass a regex for split lol

frontend/src/components/MetadataList/MetadataListSearchItem.tsx
codemonkey800: you could probably use a forward slash `/?search=`

neuromusic: of course 🤦

neuromusic:
there's actually a new plugin that throws a wrench in the implementation here: https://www.napari-hub.org/plugins/recOrder-napari, for which the (valid) author is "Computational Microscopy Platform, CZ Biohub"

the current implementation will parse this and assign two authors to the plugin: "Computational Microscopy Platform" and "CZ Biohub", which is NOT what we want

further, since I'm currently parsing this as the data model level, this could have downstream effects (say, if/when we add filter-by-author)

Issue #465, opened 2022-04-18
Add support for `.napari-hub` folder


neuromusic:
### Summary

To improve clarity around the scope and separation of concerns between the napari hub configuration folder (currently `.napari/`) and the npe2 manifest file (canonically `src/napari.yml`), we will add support for `.napari-hub/` with the goal of deprecating our support for `.napari/`.

### Outcome
For any metadata that we currently direct plugin developers to set via files in `.napari/`, they will be able to set through the same files in `.napari-hub/`.

### How it will work

We will add `.napari-hub/` as a source for the relevant metadata, taking precedence over `.napari/` if both are found.

neuromusic:
implementation: https://github.com/chanzuckerberg/napari-hub/blob/9137a96a8d07904b60edccc604ea4ff951bbdb56/backend/utils/github.py#L107

PR #470, opened 2022-04-22
napari-hub takes precedence for fetching description and config.yml

snigdhaAgarwal:
[Bug](https://github.com/chanzuckerberg/napari-hub/issues/465#event-6450749958) resolution logic:
The get_github_metadata function tries to fetch DESCRIPTION.md and config.yml from .napari-hub. And if its not able to find them, it then tries to fetch from .napari.


potating-potato (approved):
LGTM!

PR #471, opened 2022-04-22
Manifest available through API call

snigdhaAgarwal:
Changes in model.py : Mimics the behaviour of get_plugin which makes version.json available
Changes in app.py: After receiving the manifest dictionary from get_manifest function, it checks for the value of process_count to throw the appropriate error. I have set 503 as the code for temporarily unavailable as opposed to some 3xx code based on [this](https://airbrake.io/blog/http-errors/503-service-unavailable#:~:text=A%20503%20Service%20Unavailable%20Error,server%20that's%20down%20for%20maintenance.).

### DIFF
--- a/backend/api/app.py
+++ b/backend/api/app.py
@@ -74,7 +74,7 @@ def plugin_manifest(plugin: str, version: str = None) -> Response:
         if manifest['process_count'] >= max_failure_tries:
             return app.make_response(("Plugin Manifest Not Found", 404))
         elif manifest['process_count'] < max_failure_tries:
-            response = app.make_response(("Temporarily Unavailable", 503))
+            response = app.make_response(("Temporarily Unavailable my plugin!", 503))
             response.headers["Retry-After"] = 300
             return response
     else:
--- a/backend/api/app.py
+++ b/backend/api/app.py
@@ -69,17 +69,18 @@ def versioned_plugin(plugin: str, version: str = None) -> Response:
 @app.route('/manifest/<plugin>/versions/<version>')
 def plugin_manifest(plugin: str, version: str = None) -> Response:
     max_failure_tries = 2
-    print(plugin)
     manifest = get_manifest(plugin, version)
 
     if 'process_count' in manifest:
-        print("Oops")
         if manifest['process_count'] >= max_failure_tries:
-            return app.make_response(("Plugin Manifest Not Found", 404))
+            return app.make_response(("Plugin Manifest Not Found. Installation failed or plugin does not implement npe2", 404))
         elif manifest['process_count'] < max_failure_tries:
-            response = app.make_response(("Temporarily Unavailable my plugin!", 503))
+            response = app.make_response(("Temporarily Unavailable. Attempting to build manifest. Please check back "
+                                          "in 5 minutes.", 503))
             response.headers["Retry-After"] = 300
             return response
+    elif manifest == {}:
+        return app.make_response(("Plugin does not exist", 404))
     else:
         return yaml.dump(manifest)
 
--- a/backend/api/app.py
+++ b/backend/api/app.py
@@ -5,7 +5,7 @@ from flask import Flask, Response, jsonify, render_template
 from flask_githubapp.core import GitHubApp
 
 from api.model import get_public_plugins, get_index, get_plugin, get_excluded_plugins, update_cache, \
-    move_artifact_to_s3, get_category_mapping, get_categories_mapping
+    move_artifact_to_s3, get_category_mapping, get_categories_mapping, get_manifest
 from api.shield import get_shield
 from utils.utils import send_alert, reformat_ssh_key_to_pem_bytes
 
@@ -62,6 +62,19 @@ def plugins() -> Response:
 @app.route('/plugins/<plugin>/versions/<version>')
 def versioned_plugin(plugin: str, version: str = None) -> Response:
     return jsonify(get_plugin(plugin, version))
+
+
+@app.route('/manifest/<plugin>', defaults={'version': None})
+@app.route('/manifest/<plugin>/versions/<version>')
+def plugin_manifest(plugin: str, version: str = None) -> Response:
+    max_failure_tries = 2
+    manifest = get_manifest(plugin, version)
+    if manifest and 'process_count' in manifest:
+        if manifest['process_count'] >= max_failure_tries:
+            return app.make_response(("Plugin Manifest Not Found", 404))
+        elif manifest['process_count'] < max_failure_tries:
+            return app.make_response(("Temporarily Unavailable", 503))
+    return jsonify(manifest)
 
 
 @app.route('/shields/<plugin>')
--- a/backend/api/app.py
+++ b/backend/api/app.py
@@ -69,8 +69,11 @@ def versioned_plugin(plugin: str, version: str = None) -> Response:
 @app.route('/manifest/<plugin>/versions/<version>')
 def plugin_manifest(plugin: str, version: str = None) -> Response:
     max_failure_tries = 2
+    print(plugin)
     manifest = get_manifest(plugin, version)
+
     if 'process_count' in manifest:
+        print("Oops")
         if manifest['process_count'] >= max_failure_tries:
             return app.make_response(("Plugin Manifest Not Found", 404))
         elif manifest['process_count'] < max_failure_tries:
--- a/backend/api/app.py
+++ b/backend/api/app.py
@@ -71,18 +71,21 @@ def plugin_manifest(plugin: str, version: str = None) -> Response:
     max_failure_tries = 2
     manifest = get_manifest(plugin, version)
 
-    if 'process_count' in manifest:
-        if manifest['process_count'] >= max_failure_tries:
-            return app.make_response(("Plugin Manifest Not Found. Installation failed or plugin does not implement npe2", 404))
-        elif manifest['process_count'] < max_failure_tries:
-            response = app.make_response(("Temporarily Unavailable. Attempting to build manifest. Please check back "
-                                          "in 5 minutes.", 503))
-            response.headers["Retry-After"] = 300
-            return response
-    elif manifest == {}:
+    if not manifest:
         return app.make_response(("Plugin does not exist", 404))
+
+    if 'process_count' not in manifest:
+        return manifest
+
+    current_tries = manifest['process_count']
+    if current_tries >= max_failure_tries:
+        return app.make_response(
+            ("Plugin Manifest Not Found. Installation failed or plugin does not implement npe2", 404))
     else:
-        return yaml.dump(manifest)
+        response = app.make_response(("Temporarily Unavailable. Attempting to build manifest. Please check back"
+                                      " in 5 minutes.", 503))
+        response.headers["Retry-After"] = 300
+        return response
 
 
 @app.route('/shields/<plugin>')
--- a/backend/api/app.py
+++ b/backend/api/app.py
@@ -3,6 +3,7 @@ from apig_wsgi import make_lambda_handler
 from werkzeug.middleware.dispatcher import DispatcherMiddleware
 from flask import Flask, Response, jsonify, render_template
 from flask_githubapp.core import GitHubApp
+import yaml
 
 from api.model import get_public_plugins, get_index, get_plugin, get_excluded_plugins, update_cache, \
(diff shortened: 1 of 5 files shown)


backend/api/app.py
```diff
@@ -64,6 +64,19 @@ def versioned_plugin(plugin: str, version: str = None) -> Response:
...
+@app.route('/manifest/<plugin>', defaults={'version': None})
+@app.route('/manifest/<plugin>/versions/<version>')
+def plugin_manifest(plugin: str, version: str = None) -> Response:
+    max_failure_tries = 2
+    manifest = get_manifest(plugin, version)
+    if manifest and 'process_count' in manifest:
+        if manifest['process_count'] >= max_failure_tries:
+            return app.make_response(("Plugin Manifest Not Found", 404))
+        elif manifest['process_count'] < max_failure_tries:
+            return app.make_response(("Temporarily Unavailable", 503))
```
potating-potato, line 76: good call for the 503! you may want to use "Retry-After" header in conjunction

backend/api/app.py
```diff
@@ -64,6 +64,19 @@ def versioned_plugin(plugin: str, version: str = None) -> Response:
     return jsonify(get_plugin(plugin, version))
 
 
+@app.route('/manifest/<plugin>', defaults={'version': None})
+@app.route('/manifest/<plugin>/versions/<version>')
+def plugin_manifest(plugin: str, version: str = None) -> Response:
+    max_failure_tries = 2
+    manifest = get_manifest(plugin, version)
+    if manifest and 'process_count' in manifest:
+        if manifest['process_count'] >= max_failure_tries:
```
potating-potato, line 73: I believe this manifest should be YAML, and not json?

backend/api/model.py
```diff
@@ -73,6 +74,26 @@ def get_plugin(plugin: str, version: str = None) -> dict:
...
+    :param plugin: name of the plugin to get
+    :param version: version of the plugin manifest
+    :return: plugin manifest dictionary
+    """
+    plugins = get_valid_plugins()
+    if plugin not in plugins:
+        return {}
+    elif version is None:
+        version = plugins[plugin]
+    plugin = get_cache(f'cache/{plugin}/{version}.yaml')
```
potating-potato, line 90: the manifest is YAML file, may need additional processing to read it to a map

backend/api/app.py
```diff
@@ -64,6 +64,19 @@ def versioned_plugin(plugin: str, version: str = None) -> Response:
     return jsonify(get_plugin(plugin, version))
 
 
+@app.route('/manifest/<plugin>', defaults={'version': None})
+@app.route('/manifest/<plugin>/versions/<version>')
+def plugin_manifest(plugin: str, version: str = None) -> Response:
+    max_failure_tries = 2
+    manifest = get_manifest(plugin, version)
+    if manifest and 'process_count' in manifest:
```
klai95, line 72: I think you need to have a case for `if manifest not in manifest`, and then create the manifest file and write `process_count = 0` in the logic for that. part of code

potating-potato (changes requested):
there may be more thoughts required on handling the file, it should be handling yaml file instead of json

backend/api/app.py
potating-potato: see https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503

backend/api/app.py
klai95: I think `if manifest` and `else` should be the two conditions that separate out the logic 

if manifest: 
  if `process_count` in manifest and manifest['process_count'] >= max_failure_tries: 404
  if `process_count` in manifest and manifest['process_count'] < max_failure_tries: 503
else:
  create manifest file
  manifest['process_count'] = 0

klai95: i can't get the indentation correct but hopefully this makes sense

backend/api/model.py
```diff
@@ -73,6 +74,26 @@ def get_plugin(plugin: str, version: str = None) -> dict:
...
+    Get plugin manifest file for a particular plugin, get latest if version is None.
+
+    :param plugin: name of the plugin to get
+    :param version: version of the plugin manifest
+    :return: plugin manifest dictionary
+    """
+    plugins = get_valid_plugins()
+    if plugin not in plugins:
+        return {}
+    elif version is None:
```
klai95, line 88: what about the case where plugin is in list of plugins and version is provided

backend/api/app.py
snigdhaAgarwal: What should be the value for Retry-After?

backend/api/app.py
snigdhaAgarwal: get_manifest will return a dictionary

backend/api/model.py
snigdhaAgarwal: I think the code that follows is de-facto following that condition. The if-else is just for setting the version number.

potating-potato (commented):
in general looks good to me, can you please deploy this remotely (push to a branch named dev-*) and check the behavior there as well, just in case? thanks!


ps. before merging, please update the API doc to include the new endpoint and the possible error codes

backend/api/app.py
```diff
@@ -64,6 +65,22 @@ def versioned_plugin(plugin: str, version: str = None) -> Response:
     return jsonify(get_plugin(plugin, version))
 
 
+@app.route('/manifest/<plugin>', defaults={'version': None})
+@app.route('/manifest/<plugin>/versions/<version>')
+def plugin_manifest(plugin: str, version: str = None) -> Response:
+    max_failure_tries = 2
+    manifest = get_manifest(plugin, version)
+    if 'process_count' in manifest:
```
klai95, line 73: Where are you checking for existence of manifest file?

backend/api/model.py
```diff
@@ -4,10 +4,11 @@
 from zipfile import ZipFile
 from io import BytesIO
 from collections import defaultdict
+from flask import abort
```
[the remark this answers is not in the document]
potating-potato: I searched the code change and didn't find a reference to this import, can you please point me to where this is used? or maybe you are planning to drop it?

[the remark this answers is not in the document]
snigdhaAgarwal: Oh its' not used anywhere. I need to remove it. I resolved the conversation before committing the changes. Pushing the change now

backend/api/app.py
snigdhaAgarwal: That logic is in [model.py](https://github.com/chanzuckerberg/napari-hub/blob/d2dd8d780a6e3a639d78896100f3ecb989db3a54/backend/api/model.py#L95) get_manifest function. If it's unable to get the yaml file it returns a dictionary with just process_count set to 0.

backend/api/s3.py
```diff
@@ -34,6 +35,40 @@ def get_cache(key: str) -> Union[Dict, List, None]:
         return None
 
 
+def get_cache_manifest(key: str) -> Union[Dict, List, None]:
```
potating-potato, line 38: one minor thing, it is possible to reduce the code duplication on the caching of different format, potentially you can extend the original method to support another parameter, say "format", and default to json handling

backend/api/model.py
```diff
@@ -73,6 +73,27 @@ def get_plugin(plugin: str, version: str = None) -> dict:
...
 
+def get_manifest(plugin: str, version: str = None) -> dict:
+    """
+    Get plugin manifest file for a particular plugin, get latest if version is None.
+
+    :param plugin: name of the plugin to get
+    :param version: version of the plugin manifest
+    :return: plugin manifest dictionary
+    """
+    plugins = get_valid_plugins()
```
klai95, line 84: We should avoid calling get_valid_plugins() for performance reasons

backend/api/app.py
```diff
@@ -74,7 +74,7 @@ def plugin_manifest(plugin: str, version: str = None) -> Response:
         if manifest['process_count'] >= max_failure_tries:
             return app.make_response(("Plugin Manifest Not Found", 404))
         elif manifest['process_count'] < max_failure_tries:
-            response = app.make_response(("Temporarily Unavailable", 503))
+            response = app.make_response(("Temporarily Unavailable my plugin!", 503))
```
potating-potato: typo?

backend/requirements.txt
```diff
@@ -13,4 +13,5 @@ setuptools
 wheel
 pkginfo
 datadog-lambda
-build
\ No newline at end of file
+build
+cryptography==36.0.2
```
potating-potato, line 17: looking at some related discussions I think we should add pyOpenSSL to our dependency list and see if this fixed the problem, instead of binning to older version of cryptography

backend/api/app.py
```diff
@@ -64,6 +65,26 @@ def versioned_plugin(plugin: str, version: str = None) -> Response:
     return jsonify(get_plugin(plugin, version))
 
 
+@app.route('/manifest/<plugin>', defaults={'version': None})
+@app.route('/manifest/<plugin>/versions/<version>')
+def plugin_manifest(plugin: str, version: str = None) -> Response:
+    max_failure_tries = 2
+    manifest = get_manifest(plugin, version)
+
```
klai95, line 73: nit: we can get rid of this newline

backend/api/app.py
```diff
@@ -64,6 +65,26 @@ def versioned_plugin(plugin: str, version: str = None) -> Response:
...
+    if 'process_count' in manifest:
+        if manifest['process_count'] >= max_failure_tries:
+            return app.make_response(("Plugin Manifest Not Found. Installation failed or plugin does not implement npe2", 404))
+        elif manifest['process_count'] < max_failure_tries:
+            response = app.make_response(("Temporarily Unavailable. Attempting to build manifest. Please check back "
```
klai95, line 78: nit: I think there is an additional space after `back`

backend/api/app.py
```diff
@@ -64,6 +65,26 @@ def versioned_plugin(plugin: str, version: str = None) -> Response:
     return jsonify(get_plugin(plugin, version))
 
 
+@app.route('/manifest/<plugin>', defaults={'version': None})
+@app.route('/manifest/<plugin>/versions/<version>')
+def plugin_manifest(plugin: str, version: str = None) -> Response:
+    max_failure_tries = 2
+    manifest = get_manifest(plugin, version)
+
+    if 'process_count' in manifest:
```
DragaDoncila, line 74: I would update this flow to be a bit easier to follow:

```python
if not manifest:
    return 404

if process_count not in manifest:
    return manifest
    
current_tries = manifest['process_count']
if current_tries >= max_failure_tries:
    return 404
else:
    return 503
```

DragaDoncila (changes requested):
Looks good just requested a slight rearranging of the api endpoint flow for clarity. I also notice we're writing to `cache/{plugin}/{version}.yml` not `cache/manifests/{plugin}/{version}.yml` and just want to make sure that's an explicit choice?

backend/api/app.py
DragaDoncila: Debugging lol - I think that's been fixed in latest changes

potating-potato:
This is looking good, thanks so much @snigdhaAgarwal!

potating-potato (approved):

klai95 (approved):
Looks good to me. Good job!

DragaDoncila (approved):
Looks good to me! Great work @snigdhaAgarwal! 🎉

backend/api/templates/swagger.yml
```diff
@@ -128,6 +130,26 @@ paths:
...
+        - manifest
+      parameters:
+      - name: name
+        in: path
+        description: name of plugin to query
+        required: true
+        schema:
+          type: string
+        example: napari-demo
+      responses:
```
potating-potato, line 146: need to include 404 and 503 here

potating-potato:
one last thing: we didn't include the HTTP 404 and 503 in the API doc, once addressed we can merge this PR

klai95:
> > Looks good just requested a slight rearranging of the api endpoint flow for clarity. I also notice we're writing to `cache/{plugin}/{version}.yml` not `cache/manifests/{plugin}/{version}.yml` and just want to make sure that's an explicit choice?
> 
> I wanted to stick to the same format as 'cache/{plugin}/{version}.json' for the yaml.

Yeah I agree. This would be consistent with the current implementation and structure on Amazon S3

Issue #467, opened 2022-04-18
Filter plugins based on Plugin Type


neuromusic:
#### Job stories

_napari hub_

- [x]  When I'm searching for plugins, I can filter by Plugin Type so that I can find plugins that offer certain functionality
- [x]  When I'm viewing a plugins card, I can read the Plugin Type so that I can better understand what the plugin does
- [x]  When I'm viewing a plugin listing, I can read the Plugin Type

_Preview Page_

- [x] When I'm viewing the Preview Page, I can see my Plugin Type in the metadata

codemonkey800:
> - When I'm searching for plugins, I can filter by Plugin Type so that I can find plugins that offer certain functionality

<img width="300" src="https://user-images.githubusercontent.com/2176050/165858502-b9e4b15c-3afa-4724-a72f-80169280ab0c.png">

> - When I'm viewing a plugin listing, I can read the Plugin Type
> - When I'm viewing the Preview Page, I can see my Plugin Type in the metadata

<img width="300" src="https://user-images.githubusercontent.com/2176050/165858541-e7e6c948-bd5b-4aa5-9c2b-66878dca1e12.png">

codemonkey800:
> When I'm viewing a plugins card, I can read the Plugin Type so that I can better understand what the plugin does

Just realized I missed this lol I'll make another PR to add it

codemonkey800:
> When I'm viewing a plugins card, I can read the Plugin Type so that I can better understand what the plugin does

![image](https://user-images.githubusercontent.com/2176050/165867415-9cbeff7d-c774-4f79-82d0-7de217ce4f81.png)

Issue #425, opened 2022-02-09
Filtering reader & writer plugins based on file extension support


neuromusic:
### Summary
Add support for filtering reader & writer plugins based on which file extension(s) they support.

### Intended outcome
When a user has a specific file type that they need to open (or save), they will be able to find plugins that can read (or write) their file.

### How it will work
The npe2 manifest lets plugin developers explicitly define which file extension(s) reader and writer plugins support. We will use this information to add the relevant filters to the napari hub and make them available through the napari hub API.

### Job stories

_napari hub_

- [x]  When I'm searching for plugins, I can filter by the file extensions that plugins can read so that I can find plugins that will load my data
- [x]  When I'm searching for plugins, I can filter by the file extensions that plugins can write to so that I can find plugins that will let me export my data
- [x]  When I'm viewing a plugin's listing, I can read the list of file extensions that can be loaded with this plugin
- [x]  When I'm viewing a plugin's listing, I can read the list of layers that this plugin can write

_Preview Page_

- [x] When I'm viewing my plugin's preview page, I can see the metadata for file extension and layer support

neuromusic:
closing in favor of #434 as the primary epic for this work

codemonkey800:
> - When I'm searching for plugins, I can filter by the file extensions that plugins can read so that I can find plugins that will load my data

<img width="244" alt="image" src="https://user-images.githubusercontent.com/2176050/165858926-70e25c89-d4b9-4d92-a9ee-55a8f97886c5.png">

> - When I'm searching for plugins, I can filter by the file extensions that plugins can write to so that I can find plugins that will let me export my data

<img width="239" alt="image" src="https://user-images.githubusercontent.com/2176050/165858906-c6fc98e4-6f52-448d-b45e-075f992fbf3e.png">

> - When I'm viewing a plugin's listing, I can read the list of file extensions that can be loaded with this plugin
> - When I'm viewing a plugin's listing, I can read the list of layers that this plugin can write
> - When I'm viewing my plugin's preview page, I can see the metadata for file extension and layer support

<img width="254" alt="image" src="https://user-images.githubusercontent.com/2176050/165858874-c9a330c7-09e2-44c9-ace8-8a5645c4ba1c.png">

PR #472, opened 2022-04-25
npe2 plugin type + reader/writer metadata UI

codemonkey800:
Implements:
- https://github.com/chanzuckerberg/napari-hub/issues/425
- https://github.com/chanzuckerberg/napari-hub/issues/467

## Demos

https://napari-hub-npe2.vercel.app

image

image

image

image


codemonkey800:
@liaprins-czi @neuromusic oh yeah also, here's a demo with npe2 stuff on the preview page:

https://napari-hub-npe2-preview-page.vercel.app/?enable-feature=npe2

neuromusic (changes requested):
Love the alphabetical extensions.

For "Plugin Type", can we be more explicit with the ordering... e.g., keep Reader/Writer next to each other and roughly prioritize the Plugin Types?

Proposed:

- Widget
- Reader
- Writer
- Sample Data
- Theme

codemonkey800:
> Love the alphabetical extensions.

Haha this was unintentional, I think I made it alphabetical in the data, but we can sort it explicitly to make it look nice

> For "Plugin Type", can we be more explicit with the ordering... e.g., keep Reader/Writer next to each other and roughly prioritize the Plugin Types?

sounds good to me!

neuromusic:
> we can sort it explicitly to make it look nice

yes! I think this would be great, especially as the list grows

codemonkey800:
@neuromusic alright I've sorted the plugin types and file extensions: https://napari-hub-npe2.vercel.app 🔥

DragaDoncila:
@neuromusic are we 100% on going with `widget` for the analysis style plugins? I know it's the canonical term but I remember there being some discussion about it being potentially confusing for non-technical/new to napari users.

liaprins-czi (approved):
Looks awesome to me! Thanks! 🫠

frontend/i18n/en/homePage.json
```diff
@@ -38,14 +38,41 @@
...
+        "label": "Limit to plugins with open source license"
+      },
+
+      "reader": {
+        "label": "Reader",
+        "tooltip": "Reader plugins let users load data from specific filetypes."
+      },
+      "sample_data": {
+        "label": "Sample data",
+        "tooltip": "Sample Data plugins let users access sample data in napari."
```
kne42, line 62: find it weird how "data" is capitalized in the tooltip but not the label

frontend/src/components/MetadataList/MetadataList.tsx
```diff
@@ -62,7 +64,8 @@ export function MetadataList({
               'font-bold whitespace-nowrap',
 
               // Render title inline with values.
-              inline ? 'inline mr-2' : 'mb-3',
+              inline && 'inline mr-2',
+              !inline && !inlineList && 'mb-3',
```
kne42, line 68: found this a little bit confusing, so basically it's equivalent to:
```
if inline: 'inline mr-2'
elif inlineList: ''
else: 'mb-3'
```

maybe include a comment about why inlineLists don't need any margins for better human readability

frontend/src/components/PluginDetails/PluginMetadata.tsx
```diff
@@ -147,87 +152,132 @@ function PluginMetadataBase({
...
+        <SkeletonLoader
+          className="h-56"
+          render={() => (
+            <>
+              <Media between={['screen-875', 'screen-1425']}>
+                <PluginGithubData />
+              </Media>
+
+              {renderSingleItemList('pythonVersion')}
+              {renderItemList('operatingSystems')}
```
kne42, line 275: not related to this PR but was wondering why you explicitly have a function for rendering a single item item list instead of checking in the function if there were multiple items present

kne42 (approved):
a few nitpicks otherwise lgtm

neuromusic:
@DragaDoncila you're right, earlier versions had "extension"... while I personally love this, for the launch, I think we should try to be consistent with the term used throughout napari.org and the codebase.

as we tackle usability testing, this might be a good question to ask end users, though... what their understanding/assumptions are about each of these plugin types

frontend/src/components/PluginDetails/PluginMetadata.tsx
codemonkey800: haha didn't think about it at the time, we might be able to do that. the code for both functions are similar

neuromusic (approved):
looks great!

frontend/src/components/MetadataList/MetadataList.tsx
codemonkey800: actually this styling is confusing. I fixed instead by adding a margin top to each list item for spacing

PR #474, opened 2022-04-26
Dev manifest endpoint

snigdhaAgarwal:
Testing deployment


DragaDoncila:
@potating-potato I think this PR can be closed and branch deleted, as Snigdha was using this just for testing purposes. We already have the manifest endpoint code merged into main from #471

PR #479, opened 2022-04-28
Add plugin type to search result

codemonkey800:
> #467
> - When I'm viewing a plugins card, I can read the Plugin Type so that I can better understand what the plugin does

Realized I forgot to add this in #472 😅 

image


neuromusic (approved):
LGTM!

PR #480, opened 2022-04-29
Plugin metadata links + authors filter

### LINKED ISSUE #476: User Story: Plugin page metadata links
### Primary persona(s)

- [x] Research Biologist
- [x] Imaging Scientist
- [x] Bioimage Analyst

### Job Stories
- **When I** view a plugin page, **I want to** click through the metadata to it can open the search page with the metadata already set in the search filters **so that** I can see other plugins that share this metadata.

### Acceptance Criteria

1. Metadata on the plugin page appears as a link if there is an associated search page filter.
2. Clicking on metadata will take me to the search page with the metadata value set in its associated search filter.

### Notes

Slack thread:

> - turn all metadata on plugin pages that have filters on the search page into clickable links w/ filter auto-populated
>   - _Could just have all clickable metadata on plugin page have underlines to indidcate it is a link + a HTML tooltip like “Show all plugins requiring Python 3.7” (not sure if that is the best phrasing to communicate the intent of that filter, though…)_

codemonkey800:
## Description

Closes #476 

Implements metadata links for plugin pages so that users can open the search page from a plugin page's metadata using that metadata for the filter value. This works by implementing a new `MetadataListMetadataItem` component that renders a link to the search page if there's a filter available for that metadata.

Also included a bit of cleanup and added a new eslint plugin to remove unused imports.

## Demo

Prod Data: https://napari-hub-metadata-links.vercel.app
Mock Data: https://napari-hub-metadata-links-mock-data.vercel.app

https://user-images.githubusercontent.com/2176050/166075352-4ca81f6b-28f8-4921-ac21-b38b6c1e83b9.mp4

### Authors Filter

image

### DIFF
--- a/frontend/eslint/typescript.js
+++ b/frontend/eslint/typescript.js
@@ -32,7 +32,7 @@ module.exports = {
     },
   },
 
-  plugins: ['simple-import-sort'],
+  plugins: ['simple-import-sort', 'unused-imports'],
 
   rules: {
     // It's helpful to split functionality into multiple functions within a class.
@@ -54,5 +54,16 @@ module.exports = {
     'no-restricted-syntax': config.rules['no-restricted-syntax'].filter(
       (rule) => rule.selector !== 'ForOfStatement',
     ),
+
+    'unused-imports/no-unused-imports': 'error',
+    'unused-imports/no-unused-vars': [
+      'error',
+      {
+        vars: 'all',
+        varsIgnorePattern: '^_',
+        args: 'after-used',
+        argsIgnorePattern: '^_',
+      },
+    ],
   },
 };
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -144,6 +144,7 @@
     "eslint-plugin-react": "^7.26.1",
     "eslint-plugin-react-hooks": "^4.3.0",
     "eslint-plugin-simple-import-sort": "^7.0.0",
+    "eslint-plugin-unused-imports": "^2.0.0",
     "expect-playwright": "^0.8.0",
     "express": "^4.17.1",
     "husky": "^7.0.4",
--- a/frontend/src/components/MetadataList/MetadataListMetadataItem.tsx
+++ b/frontend/src/components/MetadataList/MetadataListMetadataItem.tsx
@@ -1,7 +1,7 @@
 import clsx from 'clsx';
 import { isString } from 'lodash';
 import { useTranslation } from 'next-i18next';
-import { ReactNode } from 'react';
+import { ReactNode, useMemo } from 'react';
 import { useQuery } from 'react-query';
 
 import { spdxLicenseDataAPI } from '@/axios';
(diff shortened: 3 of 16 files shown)


liaprins-czi (approved):
Works great for me! 😶‍🌫️🥿🦐

neuromusic:
awesome @codemonkey800!!!

my only concerns are the metadata items where we have different filters from what we display on the plugin page (specifically, "python version", and "operating system")

image


For "Python version", we display whatever was listed in the setup file (e.g. `>=3.7` but then we have filters for the minor versions that can be parsed from that (3.7, 3.8, and 3.9).

This makes it unclear how to handle links, since it doesn't quite make sense to link only to 3.7.

We have the same issue for Operating system.

One proposal:

We already parse the Python version and Operating system to assign specific Python versions and OSs to plugins. This might be a nice opportunity to be more consistent between the filters we use and the metadata we display on the details page, by replacing the strings like `>=3.7` with the relevant "tags" like `3.7`, `3.8`, and `3.9` then linking those to their corresponding filter selection.

Same for Operating system.

Thoughts?

codemonkey800:
> We already parse the Python version and Operating system to assign specific Python versions and OSs to plugins. This might be a nice opportunity to be more consistent between the filters we use and the metadata we display on the details page, by replacing the strings like >=3.7 with the relevant "tags" like 3.7, 3.8, and 3.9 then linking those to their corresponding filter selection.

this makes sense to me!

> Same for Operating system.

for operating system, do we want the data to be strictly Windows, Linux, and Mac, or should it be based on the data? there are some plugins that use the other available operating system classifiers for their plugins:

image

neuromusic:
> for operating system, do we want the data to be strictly Windows, Linux, and Mac, or should it be based on the data? there are some plugins that use the other available operating system classifiers for their plugins

I don't think the distinction between POSIX and Unix is meaningful to 99.99% of napari users.

Let's stick with Windows/Mac/Linux as we currently have & have described in the documentation: https://github.com/chanzuckerberg/napari-hub/blob/main/docs/customizing-plugin-listing.md#operating-system

We can be more explicit in the docs, too, with how the mapping works.

liaprins-czi:
Good catch, @neuromusic ; this all sounds great to me!

codemonkey800:
@neuromusic @liaprins-czi how does this look? now we'll show all supported python versions / operating systems and show them as links:

image

neuromusic (approved):
LGTM!

liaprins-czi:
I think this should work well for now. In future there will inevitably newer versions of Python, and the Python versions supported list will get longer, so we may want to have this be a comma-separated list (like we decided to do for files extensions supported, too)... but we don't have to do that right away!

codemonkey800:
> I think this should work well for now. In future there will inevitably newer versions of Python, and the Python versions supported list will get longer, so we may want to have this be a comma-separated list (like we decided to do for files extensions supported, too)... but we don't have to do that right away!

@liaprins-czi  sounds good, I created this is a separate issue: https://github.com/chanzuckerberg/napari-hub/issues/484

PR #481, opened 2022-05-02
Snigdha/manifest endpoint

snigdhaAgarwal:
200 response schema wasn't appearing correctly on the backend API. This pull request contains the correct swagger.yml


backend/api/templates/swagger.yml
```diff
@@ -144,16 +144,16 @@ paths:
           type: string
         example: napari-demo
       responses:
+        404:
```
potating-potato, line 147: they should be ordered in ascending order, curious how it breaks it

potating-potato (commented):
the fix seems fine for the yamp dump, but want to under the ordering of the HTTP status code

potating-potato (approved):
LGTM! thanks for the fix

PR #495, opened 2022-05-18
Add support for conda channel, with default to conda forge checking

### LINKED ISSUE #464: Let users filter plugins based on availability in the napari desktop application
### Summary

The latest releases of the napari desktop application (aka "bundled app") only support plugins that are available on conda-forge. We want to make it easier for users to find plugins that are supported by the napari desktop application and give plugin developers flexibility in specifying their corresponding conda package by resolving this information on the napari hub.

### Outcome

The napari hub will support additional metadata capturing correspondance between PyPI packages and conda(-forge) packages, provide an API endpoint for the napari desktop application to identify installable plugins, and let users filter plugins that are available in the napari desktop application.

### How it will work

- **For plugin developers**: We will let plugin developers specify one or more conda packages (and channels) where their plugin can be found as additional metadata in their napari hub configuration file.

- **For napari developers**: We will add the conda packages where the plugin can be found to the metadata associated with napari plugins. If the plugin developer has specified this in their napari hub configuration, we will present this information. If the plugin developer has not specified this info, we will check for a plugin matching the PyPI name in the `conda-forge` channel and, if found, provide this.

- **For napari hub users**: We will add conda channel(s) to the plugin metadata and let users filter for plugins which are compatible with the napari desktop application (e.g. available on conda-forge)

potating-potato:
closes #464 

add support to fetching the conda channel from github metadata, with a default to check on conda forge if not present

### DIFF
--- a/backend/utils/conda.py
+++ b/backend/utils/conda.py
@@ -1,0 +1,35 @@
+import requests
+import re
+from typing import List
+
+CONDA_FORGE = 'https://api.anaconda.org/package/conda-forge'
+
+
+def get_conda_forge_package(package: str) -> List:
+    """Get package metadata from the napari hub.
+    Parameters
+    ----------
+    package : str
+        name of the package
+    Returns
+    -------
+    List of potential conda forge packaging info.
+    """
+
+    normalized_name = normalize_name(package)
+    response = requests.get(f'{CONDA_FORGE}/{normalized_name}')
+    if response.status_code != requests.codes.ok:
+        return []
+    else:
+        return [{
+            "channel": "conda-forge",
+            "package": normalized_name
+        }]
+
+
+def normalize_name(name: str) -> str:
+    """
+    Normalize a plugin name by replacing underscores and dots by dashes and
+    lower casing it.
+    """
+    return re.sub(r"[-_.]+", "-", name).lower()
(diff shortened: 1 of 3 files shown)


backend/utils/conda.py
```diff
@@ -0,0 +1,35 @@
...
+            "package": normalized_name
+        }]
+
+
+def normalize_name(name: str) -> str:
+    """
+    Normalize a plugin name by replacing underscores and dots by dashes and
+    lower casing it.
+    """
+    return re.sub(r"[-_.]+", "-", name).lower()
```
codemonkey800, line 35: is this the same standard that conda follows?

codemonkey800 (approved):

backend/utils/conda.py
potating-potato: I copied this from the napari code part, I assume so

neuromusic:
thanks for landing this @potating-potato !!

I just tried using the staging API to fetch this metadata, but its not working the way that I expected it to. in particular, most plugins are missing the `conda` key:

image

There are a few that have the key, however... Do we need to refresh the plugin metadata to populate the key?

potating-potato:
we do need to refresh the data, let me clean up staging and double-check this

backend/utils/conda.py
goanpeca: Yep it is :)

neuromusic:
🎉 
image

PR #505, opened 2022-05-31
Update docker build process to accelerate deployment

lgan-czi:
Background:
Due to improper caching of docker images, napari-hub rebuilds the entire project every time a change is pushed to branches for remote dev, taking around 8-10 minutes to build regardless of how much the project code got altered. This is inefficient and slows down development on the project.

Changes:
Small changes were made to the frontend dockerfile to improve docker layering. We also deprecated using our own happy scripts and moved to using reusable actions instead. 

Testing Approach:
We then tested deployments in our branch until the script was able to consistently create, update, and delete stacks. 

Results:
Build times for docker images with no changes went down from 8-10 minutes to around 2 minutes. Adding changes to frontend and backend code will still take longer to build, at maybe 6-7 minutes.


.github/workflows/push-rdev.yml
```diff
@@ -33,23 +33,21 @@ jobs:
...
-            echo "Creating stack $DEV_NAME"
-            ./scripts/happy --profile="" create $DEV_NAME
-          fi
+        uses: chanzuckerberg/github-actions/.github/actions/deploy-happy-stack@v1.3.1
+        with:
+          stack-name: ${{ env.STACK_NAME }}
+          create-tag: "true"
+          tag: ${{ env.HAPPY_BRANCH }}
+          tfe-token: ${{ secrets.TFE_TOKEN }}
+          happy_version: "0.20.0"
```
potating-potato, line 49: drop this happy version here so that we are using the default from the action

docker-compose.yml
```diff
@@ -21,10 +22,11 @@ services:
           - localstack.naparinet.local
   frontend:
     image: '${DOCKER_REPO}napari-hub-dev-frontend'
+    platform: linux/amd64
     build:
       context: frontend
       cache_from:
-        - '${DOCKER_REPO}napari-hub-dev-frontend:branch-main'
+        - '${DOCKER_REPO}napari-hub-dev-frontend:${HAPPY_BRANCH}'
```
potating-potato, line 29: if we are setting this ourselves, might as well change the variable naming so not to cause any future confusion

.github/workflows/push-rdev.yml
lgan-czi: Done!

docker-compose.yml
lgan-czi: Would something like LATEST_BRANCH work?

potating-potato: I am thinking we can just use the ${GITHUB_REF#refs/heads/} here to make it clear we are using the environment variable set by github

.github/workflows/push-rdev.yml
```diff
@@ -33,23 +33,21 @@ jobs:
...
-          pip install -r .happy/requirements.txt
-          DEV_NAME=${GITHUB_REF#refs/heads/dev-}
-          if $(./scripts/happy --profile="" list | grep -q $DEV_NAME); then
-            echo "Updating stack $DEV_NAME"
-            ./scripts/happy --profile="" update $DEV_NAME
-          else
-            echo "Creating stack $DEV_NAME"
-            ./scripts/happy --profile="" create $DEV_NAME
-          fi
+        uses: chanzuckerberg/github-actions/.github/actions/deploy-happy-stack@v1.3.1
```
potating-potato, line 43: we should also move to github action form happy script for prod and staging deployment, just that the tag for prod should be git tag instead of git branch

docker-compose.yml
lgan-czi: Oh I can directly replace? I'll try it.

.github/workflows/push-rdev.yml
lgan-czi: Okay, I'll update that too

.github/workflows/deploy.yml
```diff
@@ -25,17 +25,19 @@ jobs:
...
         with:
           registry: ${{ secrets.ECR_REPO }}
-      - name: Checkout Repo
-        uses: actions/checkout@v2
-      - name: Deploy Changes
-        shell: bash
-        env:
-          ENV: prod
-          TFE_TOKEN: ${{ secrets.TFE_TOKEN }}
+      - name: Set STACK_NAME and DOCKER_TAG environment variable names
```
potating-potato, line 28: I don't think this is readable still. It descript "what" it does, technically if I spend 1 more second reading the code below, I can see that myself. So this description or naming doesn't give me any additional context. Write more on the "Why" we need this. the name below "Deploy changes to prod" is more self-explanatory of what it does

docker-compose.yml
```diff
@@ -52,14 +54,15 @@ services:
...
     build:
       context: backend
       cache_from:
-        - '${DOCKER_REPO}napari-hub-dev-backend:branch-main'
+        - '${DOCKER_REPO}napari-hub-dev-backend:${DOCKER_TAG}'
       args:
         - BUILDKIT_INLINE_CACHE=1
         - HAPPY_COMMIT
-        - HAPPY_BRANCH
+        - DOCKER_TAG
```
potating-potato, line 65: we should clean up the docker compose now that we are dealing with it, none of the args are used I believe (maybe the buildkit one)

.github/workflows/push-rdev.yml
```diff
@@ -33,23 +33,19 @@ jobs:
         uses: docker/login-action@v1
         with:
           registry: ${{ secrets.ECR_REPO }}
-      - name: Checkout Repo
-        uses: actions/checkout@v2
+      - name: Set stack and tag name inputs for deploy-happy-stack, DOCKER_TAG is needed to allow cache_from in docker-compose.yml to work
+        run: |
+          echo STACK_NAME=${GITHUB_REF#refs/heads/dev-} >> $GITHUB_ENV
```
potating-potato, line 38: if you are using GITHUB_REF_NAME below, then use that here as well

.github/workflows/push-rdev.yml
```diff
@@ -33,23 +33,19 @@ jobs:
         uses: docker/login-action@v1
         with:
           registry: ${{ secrets.ECR_REPO }}
-      - name: Checkout Repo
-        uses: actions/checkout@v2
+      - name: Set stack and tag name inputs for deploy-happy-stack, DOCKER_TAG is needed to allow cache_from in docker-compose.yml to work
```
potating-potato, line 36: I appreciate you writing more context, but this is too long :) 

maybe "Set up docker build environment" will do

.github/workflows/push-rdev.yml
lgan-czi: Done!

.github/workflows/deploy.yml
```diff
@@ -25,24 +25,18 @@ jobs:
...
           registry: ${{ secrets.ECR_REPO }}
-      - name: Checkout Repo
-        uses: actions/checkout@v2
-      - name: Setup Python
-        uses: actions/setup-python@v3
+      - name: Set up docker build environment
+        run: | 
+          echo DOCKER_TAG=${GITHUB_REF_NAME} >> $GITHUB_ENV
+      - name: Deploy Changes to Prod
+        if: ${{ github.event_name == 'push' }}
```
potating-potato, line 32: can drop this if check, the check happens at job level anyway

.github/workflows/deploy.yml
```diff
@@ -60,21 +54,15 @@ jobs:
...
           registry: ${{ secrets.ECR_REPO }}
-      - name: Checkout Repo
-        uses: actions/checkout@v2
-      - name: Setup Python
-        uses: actions/setup-python@v3
+      - name: Set up docker build environment
+        run: | 
+          echo DOCKER_TAG=${GITHUB_REF_NAME} >> $GITHUB_ENV
+      - name: Deploy Changes to Staging
+        if: ${{ github.event_name == 'push' }}
```
potating-potato, line 61: similarly can drop the if check here

.github/workflows/push-rdev.yml
```diff
@@ -33,36 +33,24 @@ jobs:
...
-          BRANCH=${{ github.event.ref }}
-          DEV_NAME=${BRANCH#dev-}
-          echo "Deleting stack $DEV_NAME"
-          ./scripts/happy --profile="" delete $DEV_NAME
\ No newline at end of file
+        uses: lgan-czi/github-actions/.github/actions/deploy-happy-stack@7cb9d22ffbc9c772505aaed1fa6cdfef3d6678af
+        #uses: chanzuckerberg/github-actions/.github/actions/deploy-happy-stack@v1.3.1
+        with:
+          stack-name: ${{ env.STACK_NAME }}
+          delete-stack: true
```
potating-potato, line 56: add a new line at the end please

.github/workflows/deploy.yml
lgan-czi: done!

.github/workflows/deploy.yml
lgan-czi: done!

.github/workflows/push-rdev.yml
lgan-czi: done!

frontend/Dockerfile
```diff
@@ -18,9 +26,17 @@ FROM node:16-alpine
...
-COPY package.json yarn.lock postinstall.js /app/
+# Copy over dependency folders for production
+COPY postinstall.js /app/
+COPY package.json /app/
+COPY yarn.lock /app/
 COPY eslint/ /app/eslint/
+
+# cache yarn 
+ENV YARN_CACHE_FOLDER=/usr/local/yarn-cache
+VOLUME /usr/local/yarn-cache
```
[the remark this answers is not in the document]
lgan-czi: I'm going to try copying what I did in frontend tests

.github/workflows/push-rdev.yml
```diff
@@ -33,36 +33,29 @@ jobs:
...
-        with:
-          python-version: '3.9'
-          cache: 'pip'
-      - name: Install happy dependencies
+      - name: Set up docker build environment
         run: |
-          pip install --upgrade pip
-          pip install -r .happy/requirements.txt
+          if [[ ${#GITHUB_REF_NAME} -gt 32 ]]; then
+            echo "branch name is too long, please use a branch name 32 characters or shorter"
```
potating-potato, line 39: by any chance can you put a "your branch is x characters too long"?

potating-potato:
fyi the npe2 change introduced another image that you need to be aware of. We need to apply the same pattern there after it is merged: https://github.com/chanzuckerberg/napari-hub/pull/520/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3

potating-potato (approved):
LGTM

Issue #468, opened 2022-04-18
make metadata from the npe1 & npe2 manifests available through the API


neuromusic:
### Job stories

_Primary Persona(s): napari developer, data analytics_

- [x]  When I'm writing a query to get plugins from napari, I can get the full manifest for any plugin on the napari hub so that I can implement new features in napari that rely on any metadata in the manifest file
- [x]  When I'm writing a query to get plugins from napari, I can identify which plugins support npe1 or npe2 so that I can measure and monitor npe2 adoption

### Tasks

- [x] Set up new AWS Lambda infrastructure in backend codebase to enable the containerized implementation for npe2
- [x] Ensure new AWS Lambda is set up in a way that aligns with security & infrastructure guidelines of the organization

codemonkey800:
@klai95  @neuromusic For the frontend, we're expecting `npe2` data to look like this:

```ts
enum PluginType {
  Widget = 'widget',
  Reader = 'reader',
  Writer = 'writer',
  SampleData = 'sample_data',
  Theme = 'theme',
}

enum PluginWriterSaveLayer {
  Image = 'image',
  Points = 'points',
  Shapes = 'shapes',
}

interface NPE2Data {
  display_name: string;
  plugin_types?: PluginType[];
  reader_file_extensions?: string[];
  writer_file_extensions?: string[];
  writer_save_layers?: PluginWriterSaveLayer[];
}
```

Here's an example of data the frontend is expecting for each plugin data. This data should be available on the `/index` and `/plugin/<name>` APIs

```json5
{
  // Human readable display name
  display_name: 'Test Plugin',

  // Plugin types
  plugin_types: ['reader', 'writer'],

  // Reader open extensions
  reader_file_extensions: [
    'czi',
    'jpg',
    'pdf',
    'png',
    'psd',
    'tiff',
    'xcar',
    'zarr',
  ],

  // Writer save extensions
  writer_file_extensions: [
    'czi',
    'jpg',
    'pdf',
    'png',
    'psd',
    'tiff',
    'xcar',
    'zarr',
  ],

  // Writer save layers
  writer_save_layers: ['image', 'points', 'shapes'],
}
```

Ideally the backend should format the data the same to minimize additional work on the frontend. However, if we need to change up the interface for some reason then let me know and we can update the frontend interfaces in another PR

richaagarwal:
[...]

PR #520, opened 2022-06-01
Backend NPE2 Feature

klai95:
`Description`: The new napari plugin specification (`npe2`) provides a reliable way for plugin developers to explicitly declare information about their plugin, such as a human-readable display name or which file extensions the plugin can load. While the primary goal of this new interface is to improve napari, some of this metadata is also useful to napari users who are searching for plugins & we want to help users. This pull request surfaces metadata for plugins (`npe1` and `npe2`) for the napari plugin ecosystem.

`Product Requirements Document`: https://czi.quip.com/rloYAqqLEm59/PRD-Activity-Dashboard

`Project Link`: https://github.com/orgs/chanzuckerberg/projects/12

`Screenshot`: Screen Shot 2022-06-14 at 8 19 07 PM

`Technical Specification`: https://docs.google.com/document/d/1JWm-71Ak6T85pEbO5g9EOJgm-PMR9LLzLwgqG1vpLL4/edit#heading=h.i83ttmuejrpb


backend/api/model.py
```diff
@@ -67,9 +67,15 @@ def get_plugin(plugin: str, version: str = None) -> dict:
...
-        return plugin
+    plugin_metadata = get_cache(f'cache/{plugin}/{version}.json')
+    if plugin_metadata:
+        try:
+            _, manifest_attributes = build_manifest_metadata(plugin, version)
+            plugin_metadata.update(manifest_attributes)
+            return plugin_metadata
+        except Exception as e:
+            error = str(e)
+            return {'error': error}
```
potating-potato, line 78: this is not documented in the API spec, you should either:
* include this behavior in API documentation or
* roll back the behavior to return empty dict

alternatively, if we are modifying the API spec, I would recommend doing proper HTTP code with the error message, this would be a 404 I think

DragaDoncila: lol yep we were just going to ping you here. I think this should still just be an empty dict, I don't think there's a reason to throw an additional error. This is leftover from debugging I think

backend/api/model.py
```diff
@@ -67,9 +67,15 @@ def get_plugin(plugin: str, version: str = None) -> dict:
         return {}
     elif version is None:
         version = plugins[plugin]
-    plugin = get_cache(f'cache/{plugin}/{version}.json')
-    if plugin:
-        return plugin
+    plugin_metadata = get_cache(f'cache/{plugin}/{version}.json')
+    if plugin_metadata:
+        try:
+            _, manifest_attributes = build_manifest_metadata(plugin, version)
```
potating-potato, line 73: we should move the build manifest logic to during build process. The get_plugin should only fetch existing file

backend/api/model.py
```diff
@@ -179,8 +206,11 @@ def update_cache():
     - cache/{plugin}/{version}.json (skip if exists)
     """
     plugins = query_pypi()
-    plugins_metadata = get_plugin_metadata_async(plugins)
+    plugins_metadata = get_plugin_metadata_async(plugins, build_plugin_metadata)
     excluded_plugins = get_updated_plugin_exclusion(plugins_metadata)
+    manifest_metadata = get_plugin_metadata_async(plugins, build_manifest_metadata)
```
potating-potato, line 211: nit picking: group the getter for metadata together, instead of being separate by the exclude

backend/requirements.txt
```diff
@@ -14,4 +14,5 @@ wheel
 pkginfo
 datadog-lambda
 build
-pyOpenSSL
\ No newline at end of file
+pyOpenSSL
+npe2 # for npe2 manifest work and could potentially be moved to installation at runtime
```
DragaDoncila, line 18: I actually don't think we *could* get away without npe2, so we should remove this comment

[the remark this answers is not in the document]
potating-potato: I see we have one already under plugins, should remove this one here then

backend/utils/utils.py
```diff
@@ -124,3 +125,53 @@ def get_category_mapping(category: str, mappings: Dict[str, List]) -> List[Dict]
...
+    plugin_types = []
+    reader_file_extensions = []
+    writer_file_extensions = []
+    writer_save_layers = []
+    manifest_attributes = {
+        'display_name': display_name,
+        'plugin_types': plugin_types,
+        'reader_file_extensions': reader_file_extensions,
+        'writer_file_extensions': writer_file_extensions,
+        'writer_save_layers': writer_save_layers,
```
DragaDoncila, line 146:
```suggestion
    manifest_attributes = {
        'display_name': '',
        'plugin_types': [],
        'reader_file_extensions': [],
        'writer_file_extensions': [],
        'writer_save_layers': [],
```

plugins/get_plugin_manifest.py
```diff
@@ -0,0 +1,61 @@
...
+import boto3
+import yaml
+import subprocess
+import sys
+from .utils import discover_manifest
+
+s3 = boto3.client('s3')
+
+
+def lambda_handler(event, context):
```
DragaDoncila, line 11: I think we should rename this to something like "generate_manifest" or something more explicit

plugins/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM public.ecr.aws/lambda/python:3.8
+
+COPY requirements.txt .
+RUN ["pip", "install", "-r", "requirements.txt"]
+
+COPY . .
+CMD ["get_plugin_manifest.lambda_handler"]
```
DragaDoncila, line 7: if we change `lambda_handler` must change this here too

backend/api/model.py
DragaDoncila: We don't write a separate `manifest.json` of the parsed fields of the raw manifest, so that's why we're parsing the raw manifest here. We could parse at build time, and write a separate json, but I guess I don't see the benefit of that, since we're merging them at runtime anyway?

DragaDoncila: I guess this would help us keep `npe2` dep out of backend though so maybe worth it

plugins/Dockerfile
klai95: Resolved

plugins/get_plugin_manifest.py
klai95: Resolved

backend/utils/utils.py
klai95: Resolved

backend/requirements.txt
klai95: Resolved

backend/api/model.py
klai95: Resolved

backend/api/model.py
klai95: Resolved

backend/api/model.py
potating-potato: maybe I should clarify what I meant:
right now the logic is as follows:
UPDATE: build json and write json to cache
GETTER: read cache json

with manifest we should do the same, just that we are nesting both:
UPDATE: build json and yaml, write json and yaml to cache
GETTER: read cached json, read cached yaml, nest them

potating-potato: another confusion I do have is around the npe2 dependency: all of the processing logic should happen in the installer lambda, and supposedly the API model should take everything in the yaml and aggregate it to the get_plugin call, without understanding what is in there. Is there something I am missing here?

DragaDoncila: I think there's two points here where there's some confusion for us. 

- The first is that `get_plugin` calls `build_manifest_metadata`. If I understand correctly, you're wondering why this needs to be called if the yaml was already written in `lambda_handler`. This function's job is just to call `get_manifest` and then parse the raw manifest into the fields expected by the frontend. However, `get_manifest` also writes the `process_count=0` to a new file if it can't find the manifest. This is *the* trigger for the `lambda_handler` - if we don't call this method here, then neither individual plugins nor `index.json` will have the manifest metadata available until someone explicitly requests the raw manifest via `/manifest/plugin-name`. This is because we rely on the `process_count` writing to trigger the lambda. Once we move to SQS as you suggested, this may no longer be so weird.
- The second point of confusion is why we need `npe2` in the `backend` when the yaml has already been written by the `lambda_handler`. This is again because the raw manifest is different to the fields expected by the frontend, which is aggregate metadata. We read the raw manifest with `npe2`s own utilities - this is part of the tech spec and is imo the only sane way to do it, because this helps us ensure we are always up to date with the manifest schema, and allows us to trivially handle changes to the manifest spec. 

If our desire is not to have an `npe2` dependency in the backend, then my proposed solution is that we parse the raw manifest in `lambda_handler` into the aggregate fields required by the front end, and write this to an entirely separate json. Then we could just read and merge them in `get_plugin`. So, the cache folder for any individual plugin would have e.g.
- `0.0.1.json` (original plugin metadata, has nothing to do with npe2
- `0.0.1.yaml` (raw manifest)
- `0.0.1.manifest.json` (aggregate manifest fields for frontend)

Note that this solution would *not* solve the issue of writing `process_count` when we call `get_plugin`, because we *still need the trigger* to update `index.json` and the individual plugin caches.

potating-potato: ok I think I am aligned on the second point, I didn't know the raw manifest is a different thing and needs to be processed. In this case, my recommendation would be the following:
/manifest/{plugin} returns the raw manifest as before
/plugins/{plugin} reads the json and the raw manifest, returns the json with the entire manifest as one attribute, and we copy the entire manifest into it without modification. (This requires our frontend to look inside this manifest attribute though). It would look like: {"name": "plugin",..., "manifest": {converted manifest in json}}

We maintain one ground truth data source by having them as json and yaml file, while we aggregate them at run time with 2 cache lookup, which shouldn't be too bad. Thoughts?

richaagarwal:
@klai95 I know this is still in a draft stage, but when you get a chance, could you fill out the PR description to give a high-level summary of the changes this is introducing?  I think you mentioned the functionality here covers more than what's listed in https://github.com/chanzuckerberg/napari-hub/issues/468, so it could be helpful to note those additions here, and/or let me know if we should create another GitHub issue to capture those.

backend/api/s3.py
```diff
@@ -65,3 +65,22 @@ def cache(content: Union[dict, list, IO[bytes]], key: str, mime: str = None, for
...
+    :param version: version of plugin
+    :return: True if npe2, False otherwise
+    """
+    key = f'cache/{plugin}/{version}.yaml'
+    try:
+        manifest_body = s3_client.get_object(Bucket=bucket, Key=os.path.join(bucket_path, key))['Body']
+        manifest_body_str = manifest_body.read().decode('utf-8')
+        if manifest_body_str.startswith('#npe2'):
+            return True
+    except ClientError:
```
richaagarwal, line 84: When would we expect something to not be in the cache?

klai95:
> @klai95 I know this is still in a draft stage, but when you get a chance, could you fill out the PR description to give a high-level summary of the changes this is introducing? I think you mentioned the functionality here covers more than what's listed in #468, so it could be helpful to note those additions here, and/or let me know if we should create another GitHub issue to capture those.

Will do, thank you!

backend/api/s3.py
klai95: If a plugin has not started the installation process, there would be no manifest file in the s3 cache for the plugin. The lambda functions are triggered by s3 file creation / deletion at the moment, which means that if I manually delete a plugin's manifest file (.yaml) on s3, via the Cloudwatch event which gets executed every 5 minutes, the backend codebase will execute and as a result a new manifest file (.yaml) would be created on s3 for the plugin. The same logic applies in the case in which a new version of the plugin file (.json), e.g. version 0.0.2, has been created. This creation of the new version of the plugin will also trigger the backend codebase to generate a corresponding manifest file (.yaml). It is important to note that, for any version of the plugin, once the manifest file is written (successful or not), it would not be changed. This would mean that any metadata added to the manifest schema thereafter would reflect in future versions of the manifest file. I hope this helps, and please let me know if I can clarify further!

backend/api/app.py
```diff
@@ -125,6 +127,10 @@ def handle_exception(e) -> Response:
 @app.errorhandler(Exception)
 def handle_exception(e) -> Response:
     send_alert(f"An unexpected error has occurred in napari hub: {e}")
+    try:
```
richaagarwal, line 130: I'm curious about this - is there a reason we can't print the error outright?  Also, where do the print statements go? (It's in my notes that at some point we should set up more robust logging, but would love to understand current state first!)

klai95: The print statements appear in log files on Cloudwatch! As for the reason we can't print the error outright, to be honest I am not sure, as that was the way the code was before I joined. I do tend to agree with a more robust logging approach that can expedite development velocity, so I think that is a good future work item to capture, although I would say that particular work would not be related to this PR.

richaagarwal: Thanks for the context!  I'll capture this for future work; agreed it's out of scope for this PR.

klai95:
> @klai95 I know this is still in a draft stage, but when you get a chance, could you fill out the PR description to give a high-level summary of the changes this is introducing? I think you mentioned the functionality here covers more than what's listed in #468, so it could be helpful to note those additions here, and/or let me know if we should create another GitHub issue to capture those.

Just added the description for PR!

backend/api/app.py
potating-potato: fwiw, it seems we don't need to raise, supposedly the traceback call below would work without the raise

potating-potato (approved):
LGTM

please verify the frontend page and everything works as intended before merge

p.s. Happy to answer terraform related questions and comments

p.p.s. @codemonkey800 is there feature flag we can enable to see the npe2 related changes? btw here's the frontend page https://add-npe2-lambda-frontend.dev.imaging.cziscience.com/plugins/affinder

codemonkey800 (approved):

.happy/terraform/modules/lambda-container/variables.tf
```diff
@@ -26,6 +26,11 @@ variable log_retention_in_days {
   default = null
 }
 
+variable maximum_retry_attempts {
+  type    = number
+  default = 0
```
potating-potato, line 31: this should be default to 2

codemonkey800:
@potating-potato @klai95 you can check for npe2 using this URL:

https://add-npe2-lambda-frontend.dev.imaging.cziscience.com/plugins/affinder?enable-feature=npe2

however, I do not see any plugin data in the frontend or API 🤔 

image

image

backend/utils/utils.py
```diff
@@ -124,3 +123,47 @@ def get_category_mapping(category: str, mappings: Dict[str, List]) -> List[Dict]
...
+        'reader_file_extensions': [],
+        'writer_file_extensions': [],
+        'writer_save_layers': [],
+    }
+    if manifest is None:
+        return manifest_attributes
+    manifest_attributes['display_name'] = manifest.get('display_name', '')
+    if 'contributions' in manifest:
+        manifest_contributions = manifest['contributions']
+        if manifest_contributions.get('readers'):
```
potating-potato, line 145: should be checked with "if readers in ..." to be pythonic

potating-potato (changes requested):
need to update the exception message

potating-potato:
> > need to update the exception message
> 
> can you clarify what we need to update with the exception message itself? Right now we have process_count and error_message

https://github.com/chanzuckerberg/napari-hub/pull/520/files#r896239190

backend/utils/utils.py
```diff
@@ -124,3 +123,47 @@ def get_category_mapping(category: str, mappings: Dict[str, List]) -> List[Dict]
         return []
     else:
         return mappings[category]
+
+
+def parse_manifest(manifest: Optional[dict] = None):
```
potating-potato, line 128: if you are struggling with these "in" checks. see if you can convert the manifest to be default dicts, that at least get rid of key error with

potating-potato: https://realpython.com/python-defaultdict/

plugins/_tests/test_manifest.py
```diff
@@ -0,0 +1,75 @@
+from plugins.get_plugin_manifest import parse_manifest
```
DragaDoncila, line 1: this whole test file is useless atm because we're parsing using a dictionary rather than the plugin manifest. This needs to be refactored and moved to `backend/utils/_tests`

potating-potato:
> however, I do not see any plugin data in the frontend or API 🤔

@codemonkey800 
they maybe emptying data to test the update mechanism, right now the manifest API is outputting https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/add-npe2-lambda/manifest/affinder

same for the https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/add-npe2-lambda/plugins/affinder

@klai95 the manifest API is not outputting format to be application/yaml, it is showing text/html hence the rendering is weird

backend/api/app.py
```diff
@@ -1,4 +1,6 @@
 import os
+import traceback
```
potating-potato, line 2: this is not used. I meant you could do the same print as before, without having to reraise the exception

backend/api/s3.py
```diff
@@ -65,3 +65,22 @@ def cache(content: Union[dict, list, IO[bytes]], key: str, mime: str = None, for
                         yaml.dump(content).encode('utf8')) as stream:
             s3_client.upload_fileobj(Fileobj=stream, Bucket=bucket,
                                      Key=os.path.join(bucket_path, key), ExtraArgs=extra_args)
+
+
+def is_npe2_plugin(plugin, version):
```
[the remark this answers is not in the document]
potating-potato: second this opinion. There are 2 part of logic in here, one is fetch the file, the other is making a decision based on the file, should separate and put the logic in the correct place

klai95:
> @potating-potato @klai95 you can check for npe2 using this URL:
> 
> https://add-npe2-lambda-frontend.dev.imaging.cziscience.com/plugins/affinder?enable-feature=npe2
> 
> however, I do not see any plugin data in the frontend or API 🤔
> 
> image
> 
> image

Updated frontend link: https://dev-add-npe2-lambda-frontend.dev.imaging.cziscience.com/plugins/affinder?enable-feature=npe2

potating-potato (approved):

PR #521, opened 2022-06-02
Collections page UI

### LINKED ISSUE #510: P0 - /collections page

codemonkey800:
Closes #510

## Demo

https://napari-hub-collections-page.vercel.app/collections

image

### DIFF
--- a/frontend/src/components/CollectionsPage/CollectionCard.tsx
+++ b/frontend/src/components/CollectionsPage/CollectionCard.tsx
@@ -1,0 +1,66 @@
+import clsx from 'clsx';
+import Image from 'next/image';
+
+import { Link } from '@/components/Link';
+import { CollectionIndexData } from '@/types/collections';
+
+interface Props {
+  collection: CollectionIndexData;
+}
+
+/**
+ * Shared line height class names for text.
+ * TODO Extract this to design system
+ */
+const LINE_HEIGHT_CLASS_NAME = 'leading-[150%] screen-495:leading-[125%]';
+
+/**
+ * Renders an image, title, summary, and curator for a collection on the
+ * collection home page.
+ */
+export function CollectionCard({ collection }: Props) {
+  return (
+    <Link
+      className="hover:bg-napari-hover-gray focus:bg-napari-hover-gray"
+      href={`/collections/${collection.symbol}`}
+    >
+      <div className="relative w-full h-[50px] screen-495:h-[100px]">
+        <Image
+          className="object-cover"
+          src={collection.cover_image}
+          alt={`${collection.title} cover image`}
+          layout="fill"
+        />
+      </div>
+
+      <h2
+        className={clsx(
+          'my-[10px] screen-495:my-[20px]',
+          'font-semibold text-[11px] screen-495:text-[17px]',
+          LINE_HEIGHT_CLASS_NAME,
+        )}
+      >
+        {collection.title}
+      </h2>
+      <p
+        className={clsx(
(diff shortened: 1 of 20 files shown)


liaprins-czi (changes requested):
This is looking really great! There are just a couple of small things I noticed to change:
- the vertical spacing surrounding the "Collections" page title should be equal above it and below it; right now it looks like it is larger below (I think because of line-height of the texts maybe, it is creating a larger gap?)
- the vertical spacing between the bottom of the very last of the listed collections and the top of the footer should be 75px (for all pages except if there's pagination, it has it's own spacing)
- the line height of the text should be 1.5, so for the short description per collection with type size 14px, it should be 21px, but it's currently at 125% (~17.5px)

Thanks, and let me know if any questions!

neuromusic (approved):
breakpoints are off and resizing the window is laggy on narrower screens, but as long as we can address those issues before release, happy to see this merged :D

codemonkey800:
> This is looking really great! There are just a couple of small things I noticed to change:
> 
> * the vertical spacing surrounding the "Collections" page title should be equal above it and below it; right now it looks like it is larger below (I think because of line-height of the texts maybe, it is creating a larger gap?)
> * the vertical spacing between the bottom of the very last of the listed collections and the top of the footer should be 75px (for all pages except if there's pagination, it has it's own spacing)
> * the line height of the text should be 1.5, so for the short description per collection with type size 14px, it should be 21px, but it's currently at 125% (~17.5px)
> 
> Thanks, and let me know if any questions!

@liaprins-czi could you review again pls and check if above issues are fixed? thank you!!

liaprins-czi (approved):
Everything I noted originally (including breakpoint issue) is now fixed; looks good to me!

frontend/src/components/CollectionsPage/CollectionCard.tsx
```diff
@@ -0,0 +1,61 @@
...
+ */
+const LINE_HEIGHT_CLASS_NAME = 'leading-[150%] screen-495:leading-[125%]';
+
+/**
+ * Renders an image, title, summary, and curator for a collection on the
+ * collection home page.
+ */
+export function CollectionCard({ collection }: Props) {
+  return (
+    <Link
```
richaagarwal, line 23: How feasible would it be to move these into CSS rules in a separate file? e.g. having specific styling defined for a collection card,  collection button, etc.  Not sure if that is part of the "Extract this to design system" work or if there are other reasons not to do this, but might be worth considering to separate concerns & perhaps enable designers to more easily make updates.

codemonkey800: Ahh so for frontend we support both Tailwind and SCSS. Right now approach we're following is always use Tailwind and then SCSS if not possible. This helps us keep the resulting CSS bundle small and helps with building UIs fast

codemonkey800: However, we do have a growing need to share certain styles from a design system. We're working with the SDS (Science Design System) team right now to get that going but it's out of scope for this PR 🤣

richaagarwal: @codemonkey800 Sounds good! I'm not too familiar with Tailwind so would love an overview of how that works some time, but definitely not a blocker or pressing need.

codemonkey800: Sounds good, happy to give an overview of the frontend tech on the hub 😄

PR #552, opened 2022-06-14
Fixed issue with deletion upon merge to main

lgan-czi:
Fixed issue where deletion wouldn't happen on merge to main by adding an input for env and tfe-token


potating-potato (approved):
LGTM, my fault for not catching it in the previous PR

PR #554, opened 2022-06-14
Analytics for create a new collection

### LINKED ISSUE #513: P1 - Plausible Analytics for collections

codemonkey800:
Closes #513


neuromusic (changes requested):
looks good, however, the plausible "goals" are getting a bit unwieldy. we should consider a naming convention for better tracking. proposed a revision to this goal here (but requires change in plausible, too)

neuromusic (approved):

frontend/src/hooks/usePlausible.ts
```diff
@@ -54,6 +54,8 @@ export type Events = {
     url: string;
     categoryDimension: string;
   };
+
+  'CTA: New Collection': unknown;
```
richaagarwal, line 58: What does "unknown" signify here?

codemonkey800: this represents the type of the data payload for the event. for example, here's what it is for filters:

https://github.com/chanzuckerberg/napari-hub/blob/3e714bd1b264c69763a580e7d6a6c00a63f113ac/frontend/src/hooks/usePlausible.ts#L21-L25

since the only thing we need from this event is whether it was fired or not, we don't really need a payload type, so I passed in `unknown` for now. check this out for more info on the `unknown` type: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type

richaagarwal: Ah makes sense! Thanks for the overview!

Issue #182, opened 2021-07-08
Vertical spacing around "View project data" link and Install button is too high


neuromusic:
*from June 2020 pre-launch bug bash*

**Created by**: Lia

## Description 

Only at 495px - 600px

Here are design references:

https://www.figma.com/file/pDKU0CdSzpSWGaoXXQVJLz/v0-hub-final-designs?node-id=17%3A3667

https://www.figma.com/file/pDKU0CdSzpSWGaoXXQVJLz/v0-hub-final-designs?node-id=17%3A602

## Info

**Bug source**: Only at 495px - 600px

**Browser**: *unknown*

**PR**: https://github.com/chanzuckerberg/napari-hub/pull/125

## Attachments

https://dl.airtable.com/.attachments/f1de241daa1d7281713c1fbaa50802ab/5f4f8680/ScreenShot2021-06-16at3_48_40PM.png

PR #574, opened 2022-06-19
Write manifest to json, update npe2 field to use new manifest field

### LINKED ISSUE #573: manifest endpoint should be JSON
#### Description
Manifest endpoints are currently formatted in YAML, which is inconsistent with most web payload standards and the rest of the napari hub API endpoints.

#### Steps/Code to Reproduce

visit https://api.staging.napari-hub.org/manifest/hesperos

#### Expected Results

JSON payload

#### Actual Results

YAML payload

DragaDoncila:
This PR closes #573 by updating the manifest discovery and parsing logic to rely on `json` rather than `yaml`.

Additionally, [npe2 #182](https://github.com/napari/npe2/pull/182) will write a schema field for shimmed plugins, so I've updated our code for deciding whether a plugin is native `npe2` or not to use this field. Since that field won't be available until `npe2` release, I'm still doing some work to ensure we have correct metadata before release. We can further clean up some of the code once `npe2` is released.

I've had to update the path to the manifest file as now that we are writing `json`, we don't want to conflict with the original plugin json. This means raw manifests are being written to `{plugin}/{version}-manifest.json`.

### DIFF
--- a/.happy/terraform/modules/ecs-stack/main.tf
+++ b/.happy/terraform/modules/ecs-stack/main.tf
@@ -209,7 +209,7 @@ resource "aws_s3_bucket_notification" "plugins_notification" {
     lambda_function_arn = module.plugins_lambda.function_arn
     events              = ["s3:ObjectCreated:*"]
     filter_prefix       = var.env == "dev" ? local.custom_stack_name : ""
-    filter_suffix       = ".yaml"
+    filter_suffix       = "-manifest.json"
   }
 
   depends_on = [aws_lambda_permission.allow_bucket]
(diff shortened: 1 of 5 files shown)


potating-potato:
The code change looks fine to me, just want to confirm the change of API format is communicated to consumers on the napari side

DragaDoncila:
@potating-potato yes, we are not using it yet on the napari side, and I've spoken to @tlambert03 about the changes here. He is actually in the process of providing the manifest discovery as an `npe2` utility which we'll be able to use on our end, cleaning up *a lot* of code, taking some of the maintenance burden off us, and also ensuring consistency in the future.

But that's coming in a bit.

neuromusic (approved):
LGTM

potating-potato (approved):
Checked the dev endpoint, looks good to me

klai95 (approved):
Looks good to me!

PR #578, opened 2022-06-20
worked around terraform resource conflict issue in lambda module

potating-potato:
had to redo https://github.com/chanzuckerberg/napari-hub/pull/577 due to commit errors (577 has been forced undone)

please review context there


DragaDoncila:
@potating-potato just to clarify before I approve: this is the change we meant to make in #577 but due to commit issues, these particular changes didn't make it into #577 so you're adding them here?

DragaDoncila (approved):
Looks good! Thanks @potating-potato

PR #583, opened 2022-06-24
remove s3 trigger to plugins lambda to prevent abuse

potating-potato:
For safety, we are dropping s3 trigger to aws lambda to avoid potential abuse to trigger massive lambdas

### DIFF
--- a/.happy/terraform/modules/ecs-stack/main.tf
+++ b/.happy/terraform/modules/ecs-stack/main.tf
@@ -194,27 +194,6 @@ resource "aws_cloudwatch_event_rule" "update_rule" {
   tags                = var.tags
 }
 
-resource "aws_lambda_permission" "allow_bucket" {
-  statement_id  = "AllowExecutionFromS3Bucket"
-  action        = "lambda:InvokeFunction"
-  function_name = module.plugins_lambda.function_arn
-  principal     = "s3.amazonaws.com"
-  source_arn    = local.data_bucket_arn
-}
-
-resource "aws_s3_bucket_notification" "plugins_notification" {
-  bucket = local.data_bucket_name
-
-  lambda_function {
-    lambda_function_arn = module.plugins_lambda.function_arn
-    events              = ["s3:ObjectCreated:*"]
-    filter_prefix       = var.env == "dev" ? local.custom_stack_name : ""
-    filter_suffix       = ".yaml"
-  }
-
-  depends_on = [aws_lambda_permission.allow_bucket]
-}
-
 resource "aws_cloudwatch_event_target" "update_target" {
     rule = aws_cloudwatch_event_rule.update_rule.name
     arn = module.backend_lambda.function_arn
--- a/.happy/terraform/modules/ecs-stack/main.tf
+++ b/.happy/terraform/modules/ecs-stack/main.tf
@@ -194,27 +194,6 @@ resource "aws_cloudwatch_event_rule" "update_rule" {
   tags                = var.tags
 }
 
-resource "aws_lambda_permission" "allow_bucket" {
-  statement_id  = "AllowExecutionFromS3Bucket"
-  action        = "lambda:InvokeFunction"
-  function_name = module.plugins_lambda.function_arn
-  principal     = "s3.amazonaws.com"
-  source_arn    = local.data_bucket_arn
-}
-
-resource "aws_s3_bucket_notification" "plugins_notification" {
-  bucket = local.data_bucket_name
-
-  lambda_function {
-    lambda_function_arn = module.plugins_lambda.function_arn
-    events              = ["s3:ObjectCreated:*"]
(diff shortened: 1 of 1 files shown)


.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -194,27 +194,6 @@ resource "aws_cloudwatch_event_rule" "update_rule" {
...
-}
-
-resource "aws_s3_bucket_notification" "plugins_notification" {
-  bucket = local.data_bucket_name
-
-  lambda_function {
-    lambda_function_arn = module.plugins_lambda.function_arn
-    events              = ["s3:ObjectCreated:*"]
-    filter_prefix       = var.env == "dev" ? local.custom_stack_name : ""
-    filter_suffix       = ".yaml"
```
DragaDoncila, line 212: ummm how come this is `.yaml` still in `main`...? It's nowhere locally for me?

richaagarwal: Good catch @DragaDoncila ! I'm noticing the same locally as well as on the main branch here: https://github.com/chanzuckerberg/napari-hub/blob/main/.happy/terraform/modules/ecs-stack/main.tf#L212 - It looks like there are conflicts on this PR so I think @potating-potato is working off an older copy of main.

potating-potato: ha! there is merge conflict I didn't notice. addressed

.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -194,27 +194,6 @@ resource "aws_cloudwatch_event_rule" "update_rule" {
...
-  lambda_function {
-    lambda_function_arn = module.plugins_lambda.function_arn
-    events              = ["s3:ObjectCreated:*"]
-    filter_prefix       = var.env == "dev" ? local.custom_stack_name : ""
-    filter_suffix       = "manifest.json"
-  }
-
-  depends_on = [aws_lambda_permission.allow_bucket]
-}
-
```
richaagarwal, line 217: Do we also need to remove the cron job configured here - https://github.com/chanzuckerberg/napari-hub/blob/main/.happy/terraform/modules/ecs-stack/main.tf#L189 ?  Or does removing these two resources somehow prevent that cron job from being triggered?

potating-potato: this cron job is used to update other backend data, as long as the s3 connection is dropped, this lambda will not be invoked by anything

richaagarwal: @potating-potato Would that result in a lot of errors when `update_cache` is called and it's not able to access an s3 connection?

DragaDoncila: @richaagarwal it should not, no. The `update_cache` method should still be able to access what is currently in `S3`, and while files *may* be created with `process_count=0` (unlikely, as they are all already there), this creation would no longer trigger the `plugins` lambda.

richaagarwal: @DragaDoncila Could you expand on what it means that files might be created?  I think I might just be a little confused on the architecture here so happy to chat more separately - and don't need to be a blocker for this PR!

potating-potato:
renamed the branch into #585

.happy/terraform/modules/ecs-stack/main.tf
DragaDoncila: @richaagarwal yep I can expand. The current update process proceeds as follows:

- Every 5 minutes `update_cache` gets called to make sure our plugin metadata is up to date
- `update_cache` fetches PyPI metadata (doesn't concern us here)
- `update_cache` fetches `npe2` metadata through a call to `get_manifest` for each plugin
- the call to `get_manifest` checks if a `{plugin-name}/{version}-manifest.json` file exists
    - if it does, it returns the metadata
    - *if it doesn't, it creates a -manifest.json file with `process_count=0` in it*
    
(meanwhile, over in the `plugins` lambda function)
- a file was just created with suffix `manifest.json`
- this triggers the `plugins` lambda
- it tries to install the plugin and discover the npe2 metadata. 
       - If discovery works fine, it writes the metadata to the `{plugin-name}/{version}-manifest.json` file, and in the process removes any mention of `process_count` from that file
       - if discovery doesn't work (either because installation failed or for some other reason), it raises an error, which triggers the failure lambda. The failure lambda then updates *overwrites* the file with incremented process count. Since the file is created again, `plugins` lambda gets triggered and reattempts discovery (in theory... just twice more)

This PR removes the file creation trigger from the `plugins` lambda, so that function will now never run.

richaagarwal: Thank you @DragaDoncila - super helpful write-up!

PR #585, opened 2022-06-24
test removing s3 trigger in dev deployment

potating-potato:
renamed the branch to test it in dev

### DIFF
--- a/.happy/terraform/modules/ecs-stack/main.tf
+++ b/.happy/terraform/modules/ecs-stack/main.tf
@@ -194,27 +194,6 @@ resource "aws_cloudwatch_event_rule" "update_rule" {
   tags                = var.tags
 }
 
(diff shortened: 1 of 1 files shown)


DragaDoncila (approved):
Looks good. I checked the `aws` instance and the `plugins` lambda has no trigger, and we've seen no runs for any of the `-manifest.json`s being written during `update_cache`.

potating-potato:
@richaagarwal if this change also makes sense to you we can merge it

richaagarwal:
@potating-potato I am fine if you want to proceed based on Draga's review, especially given this is blocking releases at the moment. See this comment - https://github.com/chanzuckerberg/napari-hub/pull/583#discussion_r906290478 - I can set up additional time with either you or Draga to understand this a little better; right now I'm confused about which lambdas control which functions.

PR #592, opened 2022-06-29
Improvements to napari-hub author splitting

lgan-czi:
Background:

Currently, we populate the Authors filter by assigning the PyPI author as a single author, even though many plugin developers populate this string with a list of authors separated by commas, ampersands, "and", and Oxford commas. This limits the value of the Author filter since it isn't very reliable (e.g. the filter for "Nicholas Sofroniew" misses plugins where Nick is one of many authors). We want to fix this so that if plugin author information is sourced from PyPI, we would split the author string according to common separators (e.g. ,, ;, &, and and) and assign the results as individual authors to the plugin.

Changes:

On the backend, I modified pypi.py to attempt to use regex to split author strings by commonly used delimiters such as "&", ",", and the word "and" and putting those author in as separate entries. I also deprecated the use of the 'emails' entry in the "authors" field.

On the frontend, I added a fix for #597 by grouping the authors on the search cards in a single comma-separated string by changing some CSS in PluginSearchResult.tsx.

SIDE EFFECTS:

One side effect of these changes is that we'll no longer be able to search for authors as easily using the main search bar. 
Screen Shot 2022-07-08 at 3 41 56 PM

This could be solved by using a | symbol between the author names, but I'm not sure if community members know about this that much. Searching for authors using filters works great though.
Screen Shot 2022-07-08 at 3 42 45 PM

Testing Approach:

The additions were tested locally using pytest unit tests both locally and remotely in Github Actions. The changes were also pushed to a remote development branch and was checked to make sure the changes reflected in the filters dropdown. The plugins were also extracted from the prod and dev APIs and the responses were compared to make sure both had the same authors.

A video of the new dropdown is attached along with a screenshot of the jupyter notebook used to compare author lists.

https://user-images.githubusercontent.com/106401390/177395001-258cea0b-bf13-4ba5-9ae9-d7b16d90b9a0.mov

Screen Shot 2022-07-05 at 11 51 03 AM


backend/utils/pypi.py
```diff
@@ -72,13 +72,23 @@ def format_plugin(plugin: dict) -> dict:
     """
     version = get_attribute(plugin, ["info", "version"])
 
+    # parse raw author names string
+    raw_name = get_attribute(plugin, ["info", "author"])
+    regexp = r'&|,|\sand\s'
```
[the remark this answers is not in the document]
lgan-czi: Done!

backend/utils/pypi.py
```diff
@@ -72,13 +72,24 @@ def format_plugin(plugin: dict) -> dict:
...
+    # parse raw author names string 
+    raw_name = get_attribute(plugin, ["info", "author"])
+    # currently splitting by "&", ",", and the word "and"
+    regexp = r'&|,|\sand\s'
+    author_names = re.split(regexp, raw_name)
+    author_names = [name.strip() for name in author_names if name is not None]
+    authors = []
+    for name in author_names:
+        if name:
+            authors.append({'name': name, 'email': get_attribute(plugin, ["info", "author_email"])})
```
nclack, line 84: Consider using a list comprehension and reducing the number of calls to `get_attribute()`.  I find this more clear:
```suggestion
    # UNTESTED SNIPPET
    author_email=get_attribute(plugin, ["info", "author_email"])
    authors = [
        dict(name=name, email=author_email)
        for name in author_names
        if name
    ]

```

lgan-czi: Oh good idea! I think we're actually depreciating the 'email' entry, so I'm going to remove that, but I'll switch to using list comprehension.

neuromusic: I <3 list comprehensions

backend/utils/pypi.py
```diff
@@ -72,13 +72,20 @@ def format_plugin(plugin: dict) -> dict:
     """
     version = get_attribute(plugin, ["info", "version"])
 
+    # parse raw author names string 
+    raw_name = get_attribute(plugin, ["info", "author"])
+    # currently splitting by "&", ",", and the word "and"
+    regexp = r'&|,|\sand\s'
```
codemonkey800, line 78: I think there are a few edge cases where the name might be an organization that includes a comma in the full name. @neuromusic do you remember the plugin that had this issue? maybe in this case we can have an exclusion list that skips splitting on that substring somehow?

codemonkey800 (commented):
love the testing! 🤩

backend/utils/pypi.py
lgan-czi: Oh yeah, I counted three when I was doing data exploration, I think they're:
misic-napari by S. Panigrahi & L. Espinosa, IAM, LCB (IAM is the team, LCB is the lab)
napari-omero by The Open Microscopy and napari teams
recOrder-napari by Computational Microscopy Platform, CZ Biohub

lgan-czi, line 78: Was also wondering if I should also add in other patterns such as ":", ";", etc.
The current plugins have only used "&", ",", and the word "and", but I'm not sure if I should do something else too

neuromusic: Good call out, @codemonkey800 but we have a mitigation plan in place in #593

the TL;DR is that those plugins can use `.napari-hub/config.yml` or `CITATION.cff`, where we will not split strings

neuromusic: > One thing I'm worried about is the possibility of false positives from these splits.

I think it is warranted to be cautious about over-splitting and fair to restrict ourselves to the standards that have emerged in the napari community.

lgan-czi: Sounds good! I'll keep the current split pattern then.

codemonkey800: awesome, fix it at the source! 🔥

neuromusic:
@potating-potato if I was a dev landing page, where would I be? :P

potating-potato:
before we merge this, I want to hear about the #597, we may want to either merge in frontend code changes here as well, or delay merging of this depending on the result

potating-potato:
@lgan-czi see comments on #597, if you can land the frontend one liner change then we should be clear to merge

lgan-czi:
Oh yeah, I'll do that after I'm done with the zulip unittests

lgan-czi:
@potating-potato @neuromusic I fixed https://github.com/chanzuckerberg/napari-hub/issues/597, I think this pr is ready to merge after a review now

richaagarwal (dismissed):
Thanks for your work on this @lgan-czi ! Given the information you outlined in https://github.com/chanzuckerberg/napari-hub/issues/604, I think we should pause on this until we're able to get feedback from @neuromusic. Those seem like fairly significant side-effects so I want to make sure we're all aligned before moving forward.

neuromusic:
Some thoughts on $604...

1. this reflects an existing quirk of our search implementation (added a comment on #604 clarifying this)
2. our analytics doesn't capture search queries so we don't know the impact of this (e.g. how many users are searching for multiple authors)
3. this PR is meant to make author FILTERING more reliable for users who want to find plugins by author 

Given 1-3, I fear its scope creep to block this PR pending an improvement to multi-field search functionality and my opinion is that we should land this PR despite #604 and tackle multi-field search improvements separately.

richaagarwal:
Thanks for weighing in @neuromusic ! That makes sense - I think we can move forward in that case. (I'm not going to approve this PR since I haven't code reviewed, but I'll remove my request for changes!)

potating-potato (approved):
Looks like we have consensus now on the behavior, I think this can merge

PR #594, opened 2022-06-30
Dev use citation

lgan-czi:
Added ability to use citation author information to override github metadata, also added unittests for the get_github_metadata method

### DIFF
--- a/backend/utils/_tests/test_github.py
+++ b/backend/utils/_tests/test_github.py
@@ -1,10 +1,54 @@
 import unittest
 from unittest.mock import patch
-from backend.utils.github import get_citation_author
+import requests
+from backend.utils.github import get_citation_author, get_github_metadata
 
 from utils.github import get_github_repo_url, get_license, get_citations
-from utils.test_utils import FakeResponse, license_response, no_license_response, citation_string, complete_author_citations
+from utils.test_utils import FakeResponse, license_response, no_license_response, citation_string, complete_author_citations, config_yaml
 
+class MockResponse:
+    def __init__(self, text, json_data, status_code):
+        self.text = text
+        self.json_data = json_data
+        self.status_code = status_code
+
+    def json(self):
+        return self.json_data
+
+    def raise_for_status(self):
+        return None
+
+def mocked_requests_get_citation(*args, **kwargs):
+    if args[0] == 'https://github.com/HEAD/CITATION.cff':
+        return MockResponse(citation_string, {"text": citation_string}, 200)
+    elif args[0] == 'https://github.com/license?ref=HEAD':
+        return MockResponse(license_response, {"text": license_response}, 200)
+
+    return MockResponse(None, None, 200)
+
+def mocked_requests_no_citation_no_config(*args, **kwargs):
+    if args[0] == 'https://github.com/license?ref=HEAD':
+        return MockResponse(license_response, {"text": license_response}, 200)
+
+    return MockResponse(None, None, 400)
+
+def mocked_requests_get_citation_and_config(*args, **kwargs):
+    if args[0] == 'https://github.com/HEAD/CITATION.cff':
+        return MockResponse(citation_string, {"text": citation_string}, 200)
+    elif args[0] == 'https://github.com/license?ref=HEAD':
+        return MockResponse(license_response, {"text": license_response}, 200)
+    elif args[0] == 'https://github.com/HEAD/.napari-hub/config.yml':
+        return MockResponse(config_yaml, {"text": config_yaml}, 200)
+
+    return MockResponse(None, None, 200)
+
+def mocked_requests_get_config(*args, **kwargs):
+    if args[0] == 'https://github.com/license?ref=HEAD':
+        return MockResponse(license_response, {"text": license_response}, 200)
+    elif args[0] == 'https://github.com/HEAD/.napari-hub/config.yml':
+        return MockResponse(config_yaml, {"text": config_yaml}, 200)
+
+    return MockResponse(None, None, 200)
 
 class TestGithub(unittest.TestCase):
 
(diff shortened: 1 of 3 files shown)


backend/utils/_tests/test_github.py
```diff
@@ -1,9 +1,54 @@
...
+        self.status_code = status_code
+
+    def json(self):
+        return self.json_data
+
+    def raise_for_status(self):
+        return None
+
+def mocked_requests_get_citation(*args, **kwargs):
+    if args[0] == 'https://github.com/HEAD/CITATION.cff':
```
potating-potato, line 22: good job make it working! however the checks are too strict and maybe trouble later if we modified the codebase, at least we can make it to check "ends with" or "contains"

backend/utils/_tests/test_github.py
```diff
@@ -60,3 +105,39 @@ def test_invalid_citation(self):
...
+        metadata = get_github_metadata("https://github.com")
+        print(metadata)
+        assert metadata["authors"] == [{'name': 'Gi N. Fa'}, {'name': 'Given Family'}]
+
+    # test that authors field doesn't populate when no citation exists and no config exists
+    @patch('requests.get', side_effect=mocked_requests_no_citation_no_config)
+    @patch('os.getenv', return_value=False)
+    def test_get_github_metadata_with_no_citation(self, mock_requests_get, mock_os_get):
+        metadata = get_github_metadata("https://github.com")
+        print(metadata)
```
potating-potato, line 126: remove prints

backend/utils/_tests/test_github.py
lgan-czi: Should I swap this for the licenses too?

backend/utils/_tests/test_github.py
lgan-czi: Done!

backend/utils/_tests/test_github.py
```diff
@@ -60,3 +105,35 @@ def test_invalid_citation(self):
...
+
+    def test_citation_name_filtering(self):
+        author = get_citation_author(citation_string)
+        assert author == complete_author_citations
+
+    # test that authors field populates when citation exists
+    @patch('requests.get', side_effect=mocked_requests_get_citation)
+    @patch('os.getenv', return_value=False)
+    def test_get_github_metadata_with_citation(self, mock_requests_get, mock_os_get):
+        metadata = get_github_metadata("https://github.com")
```
lgan-czi, line 117: I just realized I'm hardcoding this string here, was wondering if I should replace this with an env variable at the start of the script

backend/utils/_tests/test_github.py
```diff
@@ -60,3 +105,35 @@ def test_invalid_citation(self):
...
         assert citations is None
+
+    def test_citation_name_filtering(self):
+        author = get_citation_author(citation_string)
+        assert author == complete_author_citations
+
+    # test that authors field populates when citation exists
+    @patch('requests.get', side_effect=mocked_requests_get_citation)
+    @patch('os.getenv', return_value=False)
+    def test_get_github_metadata_with_citation(self, mock_requests_get, mock_os_get):
```
lgan-czi, line 116: How much description should i give to my new tests?

backend/utils/_tests/test_github.py
```diff
@@ -1,9 +1,54 @@
...
+def mocked_requests_get_citation(*args, **kwargs):
+    if args[0] != None and "CITATION.cff" in args[0]:
+        return MockResponse(citation_string, {"text": citation_string}, 200)
+    elif args[0] == 'https://github.com/license?ref=HEAD':
+        return MockResponse(license_response, {"text": license_response}, 200)
+
+    return MockResponse(None, None, 200)
+
+def mocked_requests_no_citation_no_config(*args, **kwargs):
+    if args[0] == 'https://github.com/license?ref=HEAD':
```
potating-potato, line 30: can you update the other checks to also be more flexible? thanks

lgan-czi: Will do! Also how much of a description should I give to these tests?

backend/utils/_tests/test_github.py
lgan-czi: Done!

backend/utils/_tests/test_github.py
lgan-czi: I made it more flexible in the same way I made the other tests

backend/utils/_tests/test_github.py
potating-potato: you should mention shortly about the test input and expected output, and the rational of the processing

neuromusic:
@potating-potato the CITATION.cff format is more granular than other formats and actually has explicit fields for given name and surname. The logic [here](https://github.com/chanzuckerberg/napari-hub/pull/594/files#diff-a0e46834b4db883a33d09913e33d904718756cb6e60220accbf4b13c4c8390b4R199) is great, IMHO: it is consistent w/ the _de facto_ community standard for PyPI and results in author names formated consistently regardless of where they are sourced.

backend/utils/github.py
```diff
@@ -186,3 +190,11 @@ def get_artifact(url: str, token: str) -> Union[IO[bytes], None]:
                 return None
             return response.raw
     return None
+
+def get_citation_author(citation_str: str) -> Union[Dict[str, str], None]:
+    citation_yaml = yaml.safe_load(citation_str)
+    authors = []
+    for author_entry in citation_yaml['authors']:
+        if author_entry['given-names'] and author_entry['family-names']:
```
neuromusic, line 198: should probably make sure we can populate the author field even if we don't have both the given and family names?

lgan-czi: Oh okay, that's a good idea, should we only additionally populate if there's a first name? Or would one of first or last name be fine?

neuromusic: actually, I stand corrected: it looks like the CITATION.cff spec requires both family and given names, HOWEVER it also permits simply `name`, which we should support: https://github.com/citation-file-format/citation-file-format/blob/main/schema-guide.md#authors

backend/utils/github.py
```diff
@@ -186,3 +190,11 @@ def get_artifact(url: str, token: str) -> Union[IO[bytes], None]:
...
+            authors.append({'name':author_entry['given-names'] + " " + author_entry['family-names']})
```
neuromusic, line 199:
```suggestion
        if author_entry['given-names'] and author_entry['family-names']:
            authors.append({'name':author_entry['given-names'] + " " + author_entry['family-names']})
        elif author_entry['name']:
            authors.append({'name':author_entry['name']})
```

potating-potato (approved):
LGTM

codemonkey800 (approved):
nice!

Issue #598, opened 2022-07-07
Cache doesn't update due to issue with format_plugin


lgan-czi:
#### Description
The cache in S3 failed to populate all plugins when I pushed to build in the remote dev environment

#### Steps/Code to Reproduce
Push any branch to rdev, and then observe it's cloudwatch logs in backend, you should get an error about how 'str' object has no attribute 'items'.

#### Expected Results
The S3 cache should populate with information scraped from pypi and github on build.

#### Actual Results
The S3 cache only populates information for around 30 plugins before erroring out. I printed a stack trace and it appears that the error comes from when we gather metadata from pypi. The code tries to read a json, but gets a string instead...
![Screen Shot 2022-07-07 at 3 06 00 PM](https://user-images.githubusercontent.com/106401390/177879920-18d17702-3282-4e1f-83fc-78e4bf2442cb.png)

lgan-czi:
Update: It looks like this issue happened because the pypi api changes to not include previous version releases if we request for a specific version (url in this format: https://pypi.org/pypi/{plugin}/{version}/json) if that happens, our code for filling in the "first_released" field of our metadata will fail because it expects a "releases" field. 

This immediate issue could be fixed by only querying https://pypi.org/pypi/{plugin}/json, but I'm not sure what any potential downstream consequences of this move could be.

lgan-czi:
@potating-potato , @neuromusic , @richaagarwal tagging for visibility

richaagarwal:
Thank you for opening up this ticket @lgan-czi ! This is really helpful detail. 

I did some more digging to learn more from the pypi side, and it looks like the `releases` field we've been relying on has been deprecated according to this page: https://warehouse.pypa.io/api-reference/json.html. In the future, they may also remove it from the project level (https://pypi.org/pypi/{plugin}/json), and they're recommending using the [simple API](https://pypi.org/project/pypi-simple/) instead to get release-specific information.  This appears to have been done 5 days ago without warning (https://github.com/pypi/warehouse/commit/de96b45297c9ee1aa2ff86b60123ea69448cff19) due to stability concerns.

This impacts our `first_released` field as Larry noted, as well as our `release_date` field.

I wonder if we could tackle this in two parts:
- Hotfix: Ignore `first_released` and `release_date` fields in updates for now. (Technically `release_date` appears available as `upload_time_iso_8601` in the `urls` array, but it probably doesn't make sense to rely on this location)**
[...]

PR #605, opened 2022-07-12
Fixed pypi scraping issue causing issue #598

lgan-czi:
Background:

We were unable to update the napari-hub due to #598, which happened due to pypi deprecating the "releases" field when querying with a version, causing our code to fail. 

Changes:

I prevented use of versions by setting version=None whenever `get_plugin_pypi_metadata` gets called. This fixes issue #598  because not asking for a version will allow us to access past releases while still retaining the same information as if we ask for the latest version. This should work because we query for the latest version anyways when gathering metadata from pypi.

Testing:

Changes were tested using unit tests and then tested in the dev environment to see if the plugins will populate, which they did. 
Screen Shot 2022-07-12 at 11 52 46 AM

### DIFF
--- a/backend/api/model.py
+++ b/backend/api/model.py
@@ -148,7 +148,7 @@ def build_plugin_metadata(plugin: str, version: str) -> Tuple[str, dict]:
     cached_plugin = get_cache(f'cache/{plugin}/{version}.json')
     if cached_plugin:
         return plugin, cached_plugin
-    metadata = get_plugin_pypi_metadata(plugin, version)
+    metadata = get_plugin_pypi_metadata(plugin)
     github_repo_url = metadata.get('code_repository')
     if github_repo_url:
         metadata = {**metadata, **get_github_metadata(github_repo_url)}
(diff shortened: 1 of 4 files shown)


lgan-czi:
@potating-potato @richaagarwal pinging for visibility

backend/api/model.py
```diff
@@ -148,7 +148,7 @@ def build_plugin_metadata(plugin: str, version: str) -> Tuple[str, dict]:
     cached_plugin = get_cache(f'cache/{plugin}/{version}.json')
     if cached_plugin:
         return plugin, cached_plugin
-    metadata = get_plugin_pypi_metadata(plugin, version)
```
richaagarwal, line 151: This is minor, but is there a reason not to pass `version = None` here and not make any of the following changes?  (It could be nice to support versions depending on what we decide for follow-up work, but it'd be easy to put back so it's definitely a minor design decision for now).

lgan-czi:
Yeah it looks like the cache populates with version=None too, should be able to merge whenever

richaagarwal (approved):
Thanks for your quick work on this!

PR #606, opened 2022-07-12
chore: Update readme to reflect Golang Happy usage

alexlokshin-czi:


potating-potato (approved):
LGTM, though in the future our team may drop the local dev completely because we don't use it

PR #607, opened 2022-07-13
Fixed error I introduced into customizing-plugin-listing.md

lgan-czi:
Background:

When I updated the description of how we populate author data in customizing-plugin-listing.md, I accidentally moved the example of how to set the `author` value in package metadata. to the wrong place.

Changes:

I fixed my errors in customizing-plugin-listing.md by moving the example of how to set the `author` value in package metadata. to where it should be.

### DIFF
--- a/docs/customizing-plugin-listing.md
+++ b/docs/customizing-plugin-listing.md
@@ -195,9 +195,7 @@ authors:
 # ...
 ```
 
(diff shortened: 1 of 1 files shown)


potating-potato:
hmm not sure if I understand the difference here. the old content looks ok to me?

also btw you don't need a dev branch for these doc changes, a branch without the dev- prefix is fine

lgan-czi:
Oh oops, the issue is that my changes in dev-use-citation moved the code block demonstrating how to set the 'author' value to the wrong place. It should be right under the portion of the description that explains how to set the field. 
Here's a pic of the changes:

Old version I'm trying to fix:
Screen Shot 2022-07-13 at 1 22 08 PM



Here's my fix:
Screen Shot 2022-07-13 at 1 22 38 PM

potating-potato (approved):
LGTM

PR #613, opened 2022-07-22
fix links that point to pages with non-working CSS

dchudz:
In https://github.com/chanzuckerberg/napari-hub/pull/558 I fixed a link that was a 404.

That PR updated the link to go to https://napari.org/plugins/ instead. I think that URL worked fine at the time, but now the CSS there isn't working (sorry).

The working version of that content is https://napari.org/stable/plugins/index.html, so this PR updates the link to go there.

I noticed the same issue with a reference to `https://napari.org/plugins/manifest.html`, so this fixes that too.

(If the pages without "stable" in the URL are intended to work fine, then feel free to fix that instead and ignore this PR.)

### DIFF
--- a/backend/api/templates/swagger.yml
+++ b/backend/api/templates/swagger.yml
@@ -132,7 +132,7 @@ paths:
                 $ref: '#/components/schemas/Shield'
   /manifest/{name}:
     get:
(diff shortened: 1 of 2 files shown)


dchudz:
Closing to fix branch name. Reopened as https://github.com/chanzuckerberg/napari-hub/pull/614.

PR #614, opened 2022-07-22
fix links that point to pages with non-working CSS

dchudz:
In https://github.com/chanzuckerberg/napari-hub/pull/558 I fixed a link that was a 404.

That PR updated the link to go to https://napari.org/plugins/ instead. I think that URL worked fine at the time, but now the CSS there isn't working (sorry).

The working version of that content is https://napari.org/stable/plugins/index.html, so this PR updates the link to go there.

I noticed the same issue with a reference to `https://napari.org/plugins/manifest.html`, so this fixes that too.

(If the pages without "stable" in the URL are intended to work fine, then feel free to fix that instead and ignore this PR.)

### DIFF
--- a/backend/api/templates/swagger.yml
+++ b/backend/api/templates/swagger.yml
@@ -132,7 +132,7 @@ paths:
                 $ref: '#/components/schemas/Shield'
   /manifest/{name}:
     get:
-      summary: Get the manifest file for plugin. Schema here https://napari.org/plugins/manifest.html
+      summary: Get the manifest file for plugin. Schema here https://napari.org/stable/plugins/manifest.html
       tags:
         - manifest
       parameters:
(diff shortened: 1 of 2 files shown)


richaagarwal:
Thank you for opening this, @dchudz! Let me check with the rest of the team on the intended functionality here.

@codemonkey800 @neuromusic Could you take a look? Should https://napari.org/plugins/stable/ have been 404ing in the first place or do we need to update some of our redirects and docs?

neuromusic:
With the series of reorgs of napari.org over the past several months, links have indeed been moving.

https://napari.org/stable/* is current, although the intended behavior is for  https://napari.org/* to redirect.

The CSS issue on napari.org is a known issue and was fixed with a workaround in https://github.com/napari/napari.github.io/pull/352, with the proper fix to come. 

https://napari.org/plugins/ renders fine for me at this time.

dchudz:
> although the intended behavior is for [napari.org](https://napari.org/)* to redirect.

That makes sense, thanks! The redirect is working for me too now (it wasn't yesterday when I opened this MR).

I'll close this.

neuromusic:
Awesome, thanks for the keen eye @dchudz !!

Issue #608, opened 2022-07-13
Transition `releases` data off of non-versioned API 


richaagarwal:
This is a follow-up to https://github.com/chanzuckerberg/napari-hub/issues/598, in which we implemented a hotfix for a breaking change introduced in the pypi API we query, which removed the `releases` key from the versioned API endpoint we were using. More information here: https://github.com/pypi/warehouse/pull/11775.  For the hotfix, we changed the query to hit the non-versioned API endpoint instead, with the knowledge that the field is considered deprecated there as well, though it has yet to be removed.

As [this comment](https://github.com/pypi/warehouse/pull/11775) notes, there's no current timeline to remove `releases` from the non-versioned API endpoint, but it would be prudent to start thinking about a transition plan.  In #598 I outlined these two options:

Option 1: Re-introduce fields by using the recommended [simple API](https://pypi.org/project/pypi-simple/) instead. This in turn could be broken out into two parts reintroducing just those two fields, and then later possibly re-working all of format_plugin to rely on the simple API. (Ideally these would both be done at once, but depending on how important it is to get back to populating these fields, we could delay the latter work).
Option 2: If accessing `upload_time_iso_8601` from the urls array is a reliable source for the `release_date` (which it appears to be), we may not need to switch APIs at all, and instead could re-work how we handle `first_released`. Ideally, we would only populate `first_released` the first time we grab data for a plugin, in which case it would be the same as `release_date` with no need to ever get previous version releases in any given request.

It turns out that option 2 is not very straightforward given our current S3 architecture for storing data, so I'd recommend that we revisit this work when we are ready to prioritize moving to a database. 

@neuromusic let's connect on this when you're back!

richaagarwal:
This work may end up addressing the bug reported in #611 as well, as we have a hypothesis that the lag reported there is due to PyPI's non-versioned endpoint taking a while to catch up to the latest release's data.

neuromusic:
note: in order to support #712, #702, and #703, we'll need to ingest ALL release dates from PyPI

richaagarwal:
This is currently blocked as the simple API may not be an option, as it doesn't support upload time at the moment (see https://github.com/jwodder/pypi-simple/issues/5).

PR #624, opened 2022-08-03
Check plugin version before indexing

### LINKED ISSUE #611: New napari-plugin does not show up on the napari-hub
#### Description
Hi all,

a new plugin [napari-lattice](https://pypi.org/project/napari-lattice/) does not show up on the napari-hub, even though the [setup.cfg](https://github.com/BioimageAnalysisCoreWEHI/napari_lattice/blob/master/setup.cfg#L15) appears correct to us.

Can you give us guidance on how to publish this plugin on the hub?

Thanks!

CC @pr4deepr

See also
* https://github.com/BioimageAnalysisCoreWEHI/napari_lattice/issues/1https://github.com/BioimageAnalysisCoreWEHI/napari_lattice/issues/1

potating-potato:
this address some problems identified in #611 

perform a version check when indexing plugin metadata to prevent pypi api race condition, where the listing reports a newer version but history API is not up to date to use that version.

### DIFF
--- a/backend/api/model.py
+++ b/backend/api/model.py
@@ -148,7 +148,7 @@ def build_plugin_metadata(plugin: str, version: str) -> Tuple[str, dict]:
     cached_plugin = get_cache(f'cache/{plugin}/{version}.json')
     if cached_plugin:
         return plugin, cached_plugin
-    metadata = get_plugin_pypi_metadata(plugin, version=None)
+    metadata = get_plugin_pypi_metadata(plugin, version=version)
     github_repo_url = metadata.get('code_repository')
     if github_repo_url:
         metadata = {**metadata, **get_github_metadata(github_repo_url)}
--- a/backend/api/model.py
+++ b/backend/api/model.py
@@ -149,6 +149,8 @@ def build_plugin_metadata(plugin: str, version: str) -> Tuple[str, dict]:
     if cached_plugin:
         return plugin, cached_plugin
     metadata = get_plugin_pypi_metadata(plugin, version=version)
+    if not metadata:
+        return plugin, metadata
     github_repo_url = metadata.get('code_repository')
     if github_repo_url:
         metadata = {**metadata, **get_github_metadata(github_repo_url)}
--- a/backend/utils/pypi.py
+++ b/backend/utils/pypi.py
@@ -48,16 +48,18 @@ def get_plugin_pypi_metadata(plugin: str, version: str) -> dict:
     :param version: version of the plugin
     :return: metadata dict for the plugin, empty if not found
     """
-    if version:
-        url = f"https://pypi.org/pypi/{plugin}/{version}/json"
-    else:
-        url = f"https://pypi.org/pypi/{plugin}/json"
-        
+    # versioned url https://pypi.org/pypi/{plugin}/{version}/json does not track history anymore.
+    # see https://github.com/chanzuckerberg/napari-hub/issues/598
+    url = f"https://pypi.org/pypi/{plugin}/json"
+
     try:
         response = requests.get(url)
         if response.status_code != requests.codes.ok:
             response.raise_for_status()
         info = format_plugin(json.loads(response.text.strip()))
+        if version and info['version'] != version:
+            print(f"Index Error: Skipping {plugin}:{version}, mismatching PyPI version {info['version']}")
+            return {}
         return info
     except HTTPError:
         return {}


lgan-czi (dismissed):
LGTM

lgan-czi:
The other place it's used is in is preview.py, not sure if preview.py is affected by the versioning issue too
Screen Shot 2022-08-03 at 11 23 53 AM

klai95 (approved):
LGTM

richaagarwal:
I think this addresses one of the issues brought up in #611, but we still don't know why there's sometimes a delay before a plugin shows up at all, right?

backend/utils/pypi.py
```diff
@@ -48,16 +48,18 @@ def get_plugin_pypi_metadata(plugin: str, version: str) -> dict:
     :param version: version of the plugin
     :return: metadata dict for the plugin, empty if not found
     """
-    if version:
-        url = f"https://pypi.org/pypi/{plugin}/{version}/json"
-    else:
-        url = f"https://pypi.org/pypi/{plugin}/json"
-        
+    # versioned url https://pypi.org/pypi/{plugin}/{version}/json does not track history anymore.
```
klai95, line 51: Could you provide some context as to when https://pypi.org/pypi/{plugin}/{version}/json no longer tracked history anymore?

lgan-czi: Apparently including the previous history in the versioned API was causing stability issues with pypi

lgan-czi: Here's a link to the discussion pypi had https://github.com/pypi/warehouse/pull/11775

richaagarwal: Ziyang linked to it in line 52 right below. I'd recommend reading through issue 598 in full.

klai95: Thanks both, will do

richaagarwal (changes requested):
.

backend/utils/pypi.py
```diff
@@ -48,16 +48,18 @@ def get_plugin_pypi_metadata(plugin: str, version: str) -> dict:
...
+    # see https://github.com/chanzuckerberg/napari-hub/issues/598
+    url = f"https://pypi.org/pypi/{plugin}/json"
```
richaagarwal, line 53: I'm actually a little confused now, isn't this going to re-introduce the issue that was fixed in https://github.com/chanzuckerberg/napari-hub/pull/605? The versioned endpoint doesn't have the `releases` data.

backend/api/model.py
```diff
@@ -148,7 +148,9 @@ def build_plugin_metadata(plugin: str, version: str) -> Tuple[str, dict]:
     cached_plugin = get_cache(f'cache/{plugin}/{version}.json')
     if cached_plugin:
         return plugin, cached_plugin
-    metadata = get_plugin_pypi_metadata(plugin, version=None)
+    metadata = get_plugin_pypi_metadata(plugin, version=version)
```
richaagarwal, line 151: What is the purpose of passing in the version here if we never use it?

richaagarwal (approved):
@potating-potato Thanks for looking into this! Some questions/thoughts:

Is the idea that we would wait for PyPI's non-versioned endpoint to "catch up" to the latest release? https://github.com/chanzuckerberg/napari-hub/issues/611 was originally opened due to an observed lag (>4 hours) before certain plugins were showing up at all on the hub (see https://github.com/chanzuckerberg/napari-hub/issues/611#issuecomment-1192816215). Do you think this lag is due to this issue with PyPI's non-versioned endpoint being out of date? If so, we would still expect this lag to continue, right? 

I'm okay with merging this to fix the particular versioning conflict issue you summarized in https://github.com/chanzuckerberg/napari-hub/issues/611#issuecomment-1204277858, but let's leave https://github.com/chanzuckerberg/napari-hub/issues/611 open so we can also address the lag -- it seems like PyPI's non-versioned endpoint isn't as reliable as we might've thought, and we might need to consider implementing option 2 in https://github.com/chanzuckerberg/napari-hub/issues/608 sooner than later.

potating-potato:
> @potating-potato Thanks for looking into this! Some questions/thoughts:
> 
> Is the idea that we would wait for PyPI's non-versioned endpoint to "catch up" to the latest release? 

Yes

we would still expect this lag to continue, right?

I am assuming so at the moment, I will keep 611 open

PR #628, opened 2022-08-11
Upgrade to MUI v5 and SDS v7

codemonkey800:
## Description

- Updates the hub to use Material UI v5 and SDS, and refactors the codebase to fix any broken changes introduced by the upgrade.
- Update ESLint plugin to use new `@mui` imports
- Refactor theme file into multiple files so we can use MUI theme in TypeScript
- Adds Tailwind config for hub primary and hub gray colors.
- Adds npm scripts for running dev mode using `prod` and `staging` APIs

## Demo

https://napari-hub-mui-upgrade.vercel.app

### DIFF
--- a/frontend/src/components/CategoryChip/CategoryChip.tsx
+++ b/frontend/src/components/CategoryChip/CategoryChip.tsx
@@ -85,7 +85,7 @@ function BaseCategoryChip(
         ],
       )}
       classes={{
-        label: clsx('pl-sds-s', hasTooltip && 'pr-0'),
+        label: clsx('pl-[12px]', hasTooltip && 'pr-0'),
       }}
       onClick={(event) => {
         event.preventDefault();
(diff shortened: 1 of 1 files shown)


frontend/src/components/ApplicationProvider.tsx
```diff
@@ -54,7 +53,7 @@ function MaterialUIProvider({ children }: ProviderProps) {
   }, []);
 
   return (
-    <EmotionThemeProvider theme={defaultTheme}>
+    <EmotionThemeProvider theme={theme}>
       <ThemeProvider theme={theme}>
         <StylesProvider
           // By default, Material UI will inject styles at the bottom of the
```
nclack, line 59: Just double-checking: This comment mentions material-ui. Is it still relevant? Is there a better reference link?

nclack (approved):
This looks good to me.  Lots of changes, but I double-checked you got all the material-ui stuff, and I like the theme refactoring.

liaprins-czi (changes requested):
This looks great to me. The only thing is that the category pills on the plugin cards are not quite having equal padding all the way around; I don't know ifI just noticed that now and it was always the case though? If so, it can be a separate bug!
Screen Shot 2022-08-12 at 11 48 08 AM

The padding that's on the bottom of the pill looks right to me, but on top looks too much and on left looks not enough

liaprins-czi (commented):
The borders on the pills look really good to me now! Thanks!!

PR #634, opened 2022-08-18
Frontend data validation

codemonkey800:
## Description

Adds client-side validation for backend data on the frontend using the [zod](https://github.com/colinhacks/zod) package. This will allow us to consume backend data safely and provide fallbacks for when data is invalid, enabling us to fail more gracefully.

Plugin data has way more leeway and provides more fallbacks and sanitization than collections. This is because plugin data is directly from PyPi, so we need to graceful with throwing errors when data is invalid.

However, collection data will throw an error if data is invalid. Since we have more control over the data source for collections, and since we also have validation built into the repo, it makes more sense to be a bit more strict on invalid data. We can implement a more graceful approach in the future if that behavior is also desired.

## Demos

Error that could happen from collection page if data is invalid:

image

### DIFF
--- a/frontend/src/utils/validate/helpers.ts
+++ b/frontend/src/utils/validate/helpers.ts
@@ -1,10 +1,9 @@
 /* eslint-disable no-param-reassign */
 
-import { defaultsDeep, get, hasIn, set } from 'lodash';
-import { DeepPartial, PickByValue } from 'utility-types';
+import { get, set } from 'lodash';
+import { PickByValue } from 'utility-types';
 import { z } from 'zod';
 
-import { DEFAULT_PLUGIN_DATA } from '@/constants/plugin';
 import { PluginData, PluginIndexData } from '@/types';
 
 /**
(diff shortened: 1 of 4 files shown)


frontend/src/types/collections.ts
```diff
@@ -33,10 +33,15 @@ export interface CollectionIndexData {
   symbol: string;
 }
 
+export enum CollectionVisibility {
```
klai95, line 36: Just curious, since the excluded plugin file on s3 has the `invalid` value for visibility, would it make sense to add it in `CollectionVisibility` too?

neuromusic: idk what `invalid` means here

neuromusic: to clarify: I'm not sure why a plugin would have `invalid` as a value for visibility: https://github.com/chanzuckerberg/napari-hub/wiki/Customizing-your-plugin's-listing#visibility

frontend/src/constants/plugin.ts
```diff
@@ -17,7 +17,7 @@ export const DEFAULT_PLUGIN_DATA: PluginData = {
   python_version: '',
   release_date: '',
   report_issues: '',
-  requirements: '',
+  requirements: [],
```
richaagarwal, line 20: @codemonkey800 I left some comments regarding this file in https://github.com/chanzuckerberg/napari-hub/pull/636 - curious what your thoughts are on duplication of that and potential for the typing to get out of sync. Is there a "source of truth" we could have defined somewhere in a more shared location? Even if this is duplicated in two areas, let's make sure they're working off the same expectations.

richaagarwal: I think some a (shared) JSON file works fine as well!  Not familiar with protobuf. I'm good with tackling this in another ticket though; do you mind opening up an issue for that?

frontend/src/utils/validate/helpers.ts
```diff
@@ -0,0 +1,155 @@
...
+
+      const result = validatePlugin(plugin as unknown as T);
+      if (expectedInvalidDataResult === DELETED) {
+        expect(hasIn(result, key)).toBe(false);
+      } else {
+        expect(result[key]).toEqual(expectedInvalidDataResult);
+      }
+    });
+  });
+}
```
kne42, line 155: why have a test generator in the util file instead of wherever the other tests are maintained?

kne42 (approved):
not going to be the best review since i havent touched this code in a while but generally LGTM, one nitpick on where a test generation function belongs

PR #635, opened 2022-08-18
SDS Phase 1 Rewrite

codemonkey800:
## Description

Updates the hub to use SDS bases:

- Colors
- Borders
- Typography
- Corners
- Shadows

## Demo

https://sds-rewrite-phase-1.vercel.app/

### DIFF
--- a/frontend/src/components/Layout/__snapshots__/Layout.test.tsx.snap
+++ b/frontend/src/components/Layout/__snapshots__/Layout.test.tsx.snap
@@ -184,7 +184,7 @@ exports[`<Layout /> should match snapshot 1`] = `
             </div>
           </div>
           <button
-            class="MuiButtonBase-root MuiButton-root MuiButton-contained MuiButton-containedPrimary MuiButton-sizeMedium MuiButton-containedSizeMedium MuiButton-disableElevation bg-hub-primary-400 text-sm font-semibold text-black h-[35px] col-span-2 screen-495:col-span-1 css-zddlty-MuiButtonBase-root-MuiButton-root"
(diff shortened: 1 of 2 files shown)


frontend/tailwind.config.js
```diff
@@ -33,6 +33,7 @@ module.exports = {
 
       fontFamily: {
         barlow: fontFamily,
+        jetbrains: 'JetBrains Mono, monospace',
```
klai95, line 36: Just curious and for context, what is this line achieving?

liaprins-czi (changes requested):
Thanks for sharing these updates! There are a few things I noticed don't seem to match the design:

- [ ] I was not seeing the hover or pressed state colors appear on the Install and Subscribe buttons
- [ ] The hover state on the copy-past link/button within the Install modal seemed to be too dark, like ti is using the color that should be for the pressed state, but on hover
- [ ] The code block line numbers are having a strange shadow treatment
Screen Shot 2022-09-07 at 9 30 53 AM
- [ ] The text in the filter dropdowns has too much letter-spacing (especially the sub-headers like in the Workflow step category filters); it should be regular (0) letter-spacing like it is for the rest of the text elsewhere
- [x] The sub-headers in the Workflow step category filter dropdown are all-caps, but should be just regular sentence-case capitalization

It'a also possible some of these are caused by how the hub-specific SDS library was put together by Connor, so we can check if anything needs to be changed there if needed

liaprins-czi (changes requested):
All the fixes look good, except that the copy-paste "button" within the Install modal is now a darker color than before (it should still be the lighter blue color, but it changed to be the color of other actual buttons, the medium blue color). Is it possible to make it back to its lighter blue color, and still have it change to the next darker blue shade on hover (like original request)? If not, no worries, it is not too big of a deal!

liaprins-czi (approved):
Looks great!

Issue #601, opened 2022-07-08
Improve data validation


codemonkey800:
We had an issue on the hub recently that caused the frontend to break when interacting with a feature that had broken data:

https://github.com/chanzuckerberg/napari-hub/pull/600

We should do some more data validation maybe in both the frontend and backend to ensure things like this don't happen. Would also be nice to have some monitoring to alert us when users encounter exceptions like this so we get alerted earlier if a user encounters it first.

klai95:
I will set up some time with @codemonkey800 to talk about how frontend and backend should handle data validation.

klai95:
Data Validation on Backend:
- Filter out data where empty author names are empty (s3 and frontend won't see this data)
- Make sure utils.py:get_attribute(...) has correct default data structure for each attribute

richaagarwal:
@klai95 @codemonkey800 Do you think it would make sense to consider other fields when filtering out data - for instance what happens if there are no python versions, or the plugin name is missing?

klai95:
The action item `Filter out data where empty author names are empty (s3 and frontend won't see this data)` has already been addressed, and therefore would not be captured in https://github.com/chanzuckerberg/napari-hub/pull/636. @richaagarwal

klai95:
Update: https://github.com/chanzuckerberg/napari-hub/pull/636 has been merged

PR #636, opened 2022-08-18
Improve Backend Data Validation

klai95:
# Description
Pull request that addresses https://github.com/chanzuckerberg/napari-hub/issues/601 in order to improve backend data validation as it pertains to the expected default values associated with each of the attribute frontend receives

### DIFF
--- a/backend/utils/pypi.py
+++ b/backend/utils/pypi.py
@@ -72,10 +72,10 @@ def format_plugin(plugin: dict) -> dict:
     :param plugin: plugin dictionary from pypi
     :return: formatted plugin dictionary
     """
-    version = get_attribute(plugin, ["info", "version"])
+    version = get_attribute(plugin, ["info", "version"], "")
 
     # parse raw author names string 
-    raw_name = get_attribute(plugin, ["info", "author"])
+    raw_name = get_attribute(plugin, ["info", "author"], "")
     # currently splitting by "&", ",", and the word "and"
     regexp = r'&|,|\sand\s'
     author_names = re.split(regexp, raw_name)
--- a/backend/utils/pypi.py
+++ b/backend/utils/pypi.py
@@ -84,32 +84,32 @@ def format_plugin(plugin: dict) -> dict:
 
     return {
         "name": get_attribute(plugin, ["info", "name"], ""),
-        "summary": get_attribute(plugin, ["info", "summary"]),
-        "description": get_attribute(plugin, ["info", "description"]),
-        "description_content_type": f'{get_attribute(plugin, ["info", "description_content_type"])}',
+        "summary": get_attribute(plugin, ["info", "summary"], ""),
+        "description": get_attribute(plugin, ["info", "description"], ""),
+        "description_content_type": f'{get_attribute(plugin, ["info", "description_content_type"], "")}',
         "authors": authors,
-        "license": get_attribute(plugin, ["info", "license"]),
-        "python_version": get_attribute(plugin, ["info", "requires_python"]),
+        "license": get_attribute(plugin, ["info", "license"], ""),
+        "python_version": get_attribute(plugin, ["info", "requires_python"], ""),
         "operating_system": filter_prefix(
-            get_attribute(plugin, ["info", "classifiers"]),
+            get_attribute(plugin, ["info", "classifiers"], []),
             "Operating System"),
         "release_date": get_attribute(plugin, ["releases", version, 0,
-                                               "upload_time_iso_8601"]),
+                                               "upload_time_iso_8601"], ""),
         "version": version,
         "first_released": min(
-            get_attribute(release, [0, "upload_time_iso_8601"])
-            for _, release in get_attribute(plugin, ["releases"]).items()
-            if get_attribute(release, [0, "upload_time_iso_8601"])),
+            get_attribute(release, [0, "upload_time_iso_8601"], "")
+            for _, release in get_attribute(plugin, ["releases"], "").items()
+            if get_attribute(release, [0, "upload_time_iso_8601"], "")),
         "development_status": filter_prefix(
-            get_attribute(plugin, ["info", "classifiers"]),
+            get_attribute(plugin, ["info", "classifiers"], ""),
             "Development Status"),
 
         # below are plugin details
         "requirements": get_attribute(plugin, ["info", "requires_dist"], []),
-        "project_site": get_attribute(plugin, ["info", "home_page"]),
-        "documentation": get_attribute(plugin, ["info", "project_urls", "Documentation"]),
-        "support": get_attribute(plugin, ["info", "project_urls", "User Support"]),
-        "report_issues": get_attribute(plugin, ["info", "project_urls", "Bug Tracker"]),
-        "twitter": get_attribute(plugin, ["info", "project_urls", "Twitter"]),
-        "code_repository": get_github_repo_url(get_attribute(plugin, ["info", "project_urls"]))
+        "project_site": get_attribute(plugin, ["info", "home_page"], ""),
+        "documentation": get_attribute(plugin, ["info", "project_urls", "Documentation"], ""),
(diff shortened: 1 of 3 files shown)


backend/utils/utils.py
```diff
@@ -9,12 +9,13 @@
 slack_url = os.environ.get('SLACK_URL')
 
 
-def get_attribute(obj: dict, path: list):
+def get_attribute(obj: dict, path: list, default_val: Any):
```
richaagarwal, line 12: Instead of specifying this as a parameter (and thus having to pass it in each time this function is called), what do you think about having some kind of dictionary or json file to explicitly map a field to its expected data type? Then, we could return the default value in line 27 by doing a look up, e.g. `data_types[field_name]` or something like that. 

We could also make this optional and default to a string if a value isn't passed, or the field isn't in the dictionary (if you go that route). In that case, we'd just be keeping track of fields that are lists / or other non-string values.

richaagarwal:
@klai95 I left one comment that I think could improve the readability of this and make it a little easier to keep up to date in the future.  

One question: if something is supposed to be a string or a list or whichever type we specify, how are we validating that here?  For instance, if we get back a value for `classifiers` or `requirements` (and thus don't need to default to an empty list), but that value is a string and not a list, what will happen?

richaagarwal:
Another question - should the types somehow be shared between the front-end and back-end? I noticed in @codemonkey800 's [front-end data validation PR,](https://github.com/chanzuckerberg/napari-hub/pull/634) he updated this file: https://github.com/chanzuckerberg/napari-hub/blob/main/frontend/src/constants/plugin.ts

Would be great if we could minimize duplication of this somehow.

backend/utils/pypi.py
```diff
@@ -72,44 +72,44 @@ def format_plugin(plugin: dict) -> dict:
...
+                                               "upload_time_iso_8601"], ""),
         "version": version,
         "first_released": min(
-            get_attribute(release, [0, "upload_time_iso_8601"])
-            for _, release in get_attribute(plugin, ["releases"]).items()
-            if get_attribute(release, [0, "upload_time_iso_8601"])),
+            get_attribute(release, [0, "upload_time_iso_8601"], "")
+            for _, release in get_attribute(plugin, ["releases"], "").items()
+            if get_attribute(release, [0, "upload_time_iso_8601"], "")),
         "development_status": filter_prefix(
```
richaagarwal, line 103: I left a separate comment about the typing defined in https://github.com/chanzuckerberg/napari-hub/blob/main/frontend/src/constants/plugin.ts, but it looks like in there this field is expected to be an array. 

(Ideally we could somehow use a shared schema so it's not defined in both the frontend and backend code)

klai95: If we are purposing the shared schema work in a separate PR, what changes should we capture in this PR, just to get some clarity?

richaagarwal: @klai95 @codemonkey800 I'll leave it up to you two on how to coordinate follow-up work around a shared schema, but I'd recommend that at least for this first effort, even if we define this in both frontend and backend code, that you should both be working off the same criteria.  I left this comment originally because I noticed in https://github.com/chanzuckerberg/napari-hub/pull/634 and https://github.com/chanzuckerberg/napari-hub/blob/main/frontend/src/constants/plugin.ts that Jeremy is working off slightly different definitions (e.g. `development_status` is a list there, but a string in this PR). I'm curious about this discrepancy.

@klai95 I think the other feedback I gave would still hold, in terms of cleaning up this code slightly and keeping track of the types in a dictionary or something more readable.

klai95:
After talking to @DragaDoncila, we think that it would be helpful, in a separate PR, to incorporate the shared schema potentially using the snippet of code as an initial prototype. In this case, the below code would live in a newly created python file, and hopes to address a few of the current problems we have:

(1) make the return statement in `pypi.py` more readable, as it currently contains 20 lines of code
(2) create more tests by purposing the fetching of the attribute values in a separate file, so that, to  
      @richaagarwal's point here: https://github.com/chanzuckerberg/napari-hub/pull/636#issuecomment-1220299267, we 
      can validate / test each attribute more rigorously 

The idea is we would create a dictionary in which the key is the attribute name and the value is a tuple of (path, default_val, helper_method). For example,
`field_dict = {
        'name': (["info", "name"], "", None)
}`
Note that `helper_method` takes the value `None` in this case because it is only calling `get_attribute` and does not require additional helper methods like some other attributes do

Then, instead of calling `get_attribute on every line where we have an attribute, we can do the following:
`plugin_data = {}
for key, (path, default_val, helper_method) in field_dict.items():
    plugin_data[key] = helper_method(plugin, path) or default_val
return plugin_data`
Note that `helper_method` would return false in the case where we would want to assign the attribute to the default value

klai95:
This PR ensures the correct default data structure is assigned to each attribute that aligns with frontend's expectation. However, it is worth noting that this PR does not currently address the case in which PyPI deviates from our expected schema.

klai95:
Another point worth noting is to be mindful of additional items we should add around filtering out data for all list fields when the values within that list are empty. To clarify in further detail, this work would entail returning an empty list when all the nested values are empty. It does not seem like currently, based on the remaining list fields, there is anything that would apply to this potential issue, but it is worth calling out just for context.

backend/utils/utils.py
```diff
@@ -7,6 +7,13 @@
 
 # Environment variable set through ecs stack terraform module
 slack_url = os.environ.get('SLACK_URL')
+# Dictionary in which the key is the last element in path of type list
+# and the value is the default value of the attribute. Keys that do not
+# exist in this dictionary will have the default value of the empty string
+path_to_default_values_dict = {
```
richaagarwal, line 13: Maybe rename this to `pypi_field_default_values`?  The key is the field name and not the full path, so I think this will be more descriptive.  I don't think we need to specify the data structure in the name, so `dict` seems okay to remove.

backend/utils/utils.py
```diff
@@ -18,13 +25,17 @@ def get_attribute(obj: dict, path: list):
     :return: the value if the path is accessible, empty string if not found
     """
     current_location = obj
+    path_to_default_values_dict_key = path[len(path) - 1]
```
richaagarwal, line 28: To get the last item in a list, you can actually do list[-1], so in this case, `path[-1]`.  Just a suggestion, but `pypi_field_name = path[-1]` might be more readable here.

backend/utils/utils.py
```diff
@@ -18,13 +25,17 @@ def get_attribute(obj: dict, path: list):
...
         if isinstance(current_location, dict) and token in current_location:
             current_location = current_location[token]
         elif isinstance(current_location, list) and token < len(current_location):
             current_location = current_location[token]
         else:
-            return ""
+            if path_to_default_values_dict_key in path_to_default_values_dict:
+                return path_to_default_values_dict[path_to_default_values_dict_key]
+            else:
+                return ""
```
richaagarwal, line 38: Another minor suggestion, but this can be a one-liner by calling the dictionary with a `.get`, so e.g. `name_of_dict.get(key, "")`

klai95:
Is there anything else that this PR needs to capture before it is merged?

backend/utils/utils.py
```diff
@@ -7,6 +7,13 @@
 
 # Environment variable set through ecs stack terraform module
 slack_url = os.environ.get('SLACK_URL')
+# Dictionary in which the key is the last element in path of type list
```
richaagarwal, line 10: One more minor thing - I think this comment needs to be updated now that you've updated the variable name below.

klai95: Just pushed a commit to update the comment!

richaagarwal (approved):

klai95:
@DragaDoncila @codemonkey800 Hi all! Is there any additional feedback that you would like to capture in this PR? Wanted to get your eyes on this PR before I merge it, even with Richa's approval just in case there was something missing

DragaDoncila (approved):
@klai95 looks great!

PR #641, opened 2022-08-23
Remove configuration for failure lambda

### LINKED ISSUE #637: Remove `failure` lambda and `plugins` lambda destination config
### Summary
Remove configuration for `failure` lambda so that `plugins` lambda has no destination and the `failure` lambda no longer gets deployed.

Partially addresses #587.

DragaDoncila:
This PR closes #637 by removing all configuration for the `failure` lambda. I've also configured the `plugins` lambda timeout and fixed the maximum retry configuration for the `plugins` lambda in the process.

The deployed functions are [here](https://us-west-2.console.aws.amazon.com/lambda/home?region=us-west-2#/functions) and we can see:
- no `dev-remove-failure-lambda-failure` function
- [no destination configured](https://us-west-2.console.aws.amazon.com/lambda/home?region=us-west-2#/functions/dev-remove-failure-lambda-plugins?tab=configure) on `dev-remove-failure-lambda-plugins` function
- [timeout configured](https://us-west-2.console.aws.amazon.com/lambda/home?region=us-west-2#/functions/dev-remove-failure-lambda-plugins?tab=configure) to 300 seconds (5 minutes) on the `dev-remove-failure-lambda-plugins` function
- [asynchronous invocation maximum retries](https://us-west-2.console.aws.amazon.com/lambda/home?region=us-west-2#/functions/dev-remove-failure-lambda-plugins?tab=configure) configured to 0 on the `dev-remove-failure-lambda-plugins` function

This does *not* bring back the triggers for the `plugins` lambda.

### DIFF
--- a/.happy/terraform/modules/ecs-stack/main.tf
+++ b/.happy/terraform/modules/ecs-stack/main.tf
@@ -142,11 +142,16 @@ module plugins_lambda {
   }
 
   log_retention_in_days = 14
-  timeout               = 900
+  timeout               = 300
   memory_size           = 10240
   ephemeral_storage_size = 10240
-  maximum_retry_attempts = 0
+
   create_async_event_config = true
+  attach_async_event_policy = true
+
+  maximum_event_age_in_seconds = 300
+  maximum_retry_attempts       = 0
+
 }
 
 module api_gateway_proxy_stage {
@@ -225,7 +230,7 @@ data aws_iam_policy_document plugins_policy {
 
     resources = ["${local.data_bucket_arn}/*"]
   }
-  
+
 }
 
 resource aws_iam_role_policy policy {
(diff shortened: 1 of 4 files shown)


.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -144,35 +142,10 @@ module plugins_lambda {
   }
 
   log_retention_in_days = 14
-  timeout               = 900
+  timeout               = 300
```
richaagarwal, line 145: Decreasing the timeout length makes sense to me; I'm wondering if it should be even less than 5 minutes given that the cron job runs every 5 minutes (or maybe the cron job is updated to run every 10 minutes instead?) It'll be nice to be confident that the function returns or times out before it's invoked again.

potating-potato (approved):
LGTM!

.happy/terraform/modules/ecs-stack/main.tf
DragaDoncila: @richaagarwal I'll make it 2.5 minutes - I'd like to avoid a whole host of timeout errors when we first start invoking the function. Once I have a good sense of how long it's taking to complete we can further decrease the timeout if needed (or adjust the cron job).

richaagarwal (approved):

PR #642, opened 2022-08-26
Backend Error Fix

klai95:
Per the incident https://app.datadoghq.com/apm/error-tracking?query=&issueId=931045fc-2551-11ed-acc9-da7ad0900002&start=1661530894372&end=1661534494372&paused=false, this pull request addresses the case in which `config` is `None`, which impacts the staging and prod environments as a result. The solution is therefore to set the default value in the case in which `config is None` to `config = {}`

### DIFF
--- a/backend/utils/github.py
+++ b/backend/utils/github.py
@@ -148,7 +148,6 @@ def get_github_metadata(repo_url: str, branch: str = 'HEAD') -> dict:
         # if the yaml.safe_load method returns None, then assign {} to config
         if config is None:
             config = {}
-            print(yaml_file)
         hub_config = {key: config[key] for key in hub_config_keys if key in config}
         github_metadata.update(hub_config)
 
--- a/backend/utils/github.py
+++ b/backend/utils/github.py
@@ -145,6 +145,9 @@ def get_github_metadata(repo_url: str, branch: str = 'HEAD') -> dict:
         yaml_file = get_file(repo_url, ".napari/config.yml", branch=branch)
     if yaml_file:
         config = yaml.safe_load(yaml_file)
+        # if the yaml.safe_load method returns None, then assign {} to config
+        if config is None:
+            config = {}
         hub_config = {key: config[key] for key in hub_config_keys if key in config}
         github_metadata.update(hub_config)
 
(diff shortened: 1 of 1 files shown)


backend/utils/github.py
```diff
@@ -145,6 +145,10 @@ def get_github_metadata(repo_url: str, branch: str = 'HEAD') -> dict:
         yaml_file = get_file(repo_url, ".napari/config.yml", branch=branch)
     if yaml_file:
         config = yaml.safe_load(yaml_file)
+        # if the yaml.safe_load method returns None, then assign {} to config
+        if config is None:
+            config = {}
+            print(yaml_file)
```
richaagarwal, line 151: This print statement won't end up in the Datadog errors since this will no longer be erroring, but I think logging this at the warning level would bubble up to Datadog (see https://docs.python.org/3/howto/logging.html).  In the meantime, I'd try running this code locally.

backend/utils/github.py
```diff
@@ -145,6 +145,10 @@ def get_github_metadata(repo_url: str, branch: str = 'HEAD') -> dict:
         yaml_file = get_file(repo_url, ".napari/config.yml", branch=branch)
     if yaml_file:
         config = yaml.safe_load(yaml_file)
+        # if the yaml.safe_load method returns None, then assign {} to config
+        if config is None:
+            config = {}
```
richaagarwal, line 150: Looking at the remaining code, it seems like it'd be better to ensure that it doesn't execute at all, however I think this is a good quick fix for now.

richaagarwal (approved):

richaagarwal:
Given that this is impacting prod and preventing the `update_cache` cron job from completing successfully, I think this is worth deploying as a hotfix.

backend/utils/github.py
klai95: I agree, and that makes sense!

backend/utils/github.py
klai95: Yeah I was just trying to test some things out I was going to delete it

klai95:
Note: Look into logging for the case when `if config is None` to examine plugin data

backend/utils/github.py
klai95: Action item: look into how to view print statements that are written in code in Datadog

Issue #621, opened 2022-07-28
Use `npe2 fetch` for discovering plugin manifests


DragaDoncila:
### Summary
This issue captures the work described in TODO number 3 of #587. Refactor `plugins/get_plugin_manifest.generate_manifest` to use `npe2 fetch` command for discovering/generating manifests.

**Rationale**: Following npe2 [#185](https://github.com/napari/npe2/pull/185) the library now provides a fetch command which takes a plugin name and version, and retrieves its manifest (building it if the plugin is npe1). Using this tool is a good separation of concerns between the napari hub and the npe2 library, and ensures we use the same mechanisms a plugin developer would to ensure their plugin's compatibility with npe2 and the napari hub. Additionally, the mechanism is more full featured than our initial implementation: avoids plugin installs for packages where the wheel can be retrieved, more options for customization and error message reporting.

### Intended Outcome
Our manifest discovery & generation is identical to that used in `napari` and reproducible by plugin developers locally.

### How will it work?
We will replace the majority of the code in `plugins/get_plugin_manifest` to instead just run the `npe2 fetch` command. We still need to decide on certain customizable options for this command:

- install napari on initial failure in case plugin was missing napari dependency? (may need small npe2 change)
- validate that npe1 plugins can actually be imported before generating manifest? (may need small npe2 change)
- catch missing dependency errors from plugin and attempt to fix?
- turn off env building for plugin install to speed things up, since we're running in a dedicated container anyway? (needs small npe2 change)

DragaDoncila:
This issue can be closed following the work in #645.

PR #645, opened 2022-08-30
Update plugins lambda handler to use `npe2.fetch_manifest` for discovery

DragaDoncila:
As per #621 , this PR removes the manual manifest discovery code in the `plugins` lambda and replaces it with `npe2.fetch_manifest`.

To ensure idempotency and limit the likelihood of repeated processing of the same plugin, we also add a check for whether the manifest file already exists.

I've also added tests that can be run locally and on PRs touching the `plugins/` directory. These are not end to end tests i.e. they mock the s3 client and do not interact with AWS. They're pretty comprehensive otherwise though.

### DIFF
--- a/plugins/get_plugin_manifest.py
+++ b/plugins/get_plugin_manifest.py
@@ -1,6 +1,7 @@
 import json
 import os
 import boto3
+from botocore.exceptions import ClientError
 from npe2 import fetch_manifest
 
 s3 = boto3.client('s3')
@@ -16,12 +17,14 @@ def generate_manifest(event, context):
     plugin = event['plugin']
     version = event['version']
     key = os.path.join(bucket_path, f'cache/{plugin}/{plugin}.{version}-manifest.json')
-    # print(key)
     try:
         existing_manifest = s3.get_object(Bucket=bucket, Key=key)
         return
-    # will fail on nonexistent, need to find proper exception class
-    except Exception as e:
+    except ClientError as e:
+        # we don't want to hide any "real" s3 errors
+        if not e.response['Error']['Code'] == 'NoSuchKey':
+            raise e        
+            
         try:
             manifest = fetch_manifest(plugin, version)
             s3_body = manifest.json()
--- a/plugins/get_plugin_manifest.py
+++ b/plugins/get_plugin_manifest.py
@@ -4,9 +4,9 @@ import boto3
 from botocore.exceptions import ClientError
 from npe2 import fetch_manifest
 
-s3 = boto3.client('s3')
+s3 = boto3.resource('s3')
 # Environment variable set through ecs stack terraform module
-bucket = os.environ.get('BUCKET')
+bucket_name = os.environ.get('BUCKET')
 bucket_path = os.environ.get('BUCKET_PATH', '')
 
 def generate_manifest(event, context):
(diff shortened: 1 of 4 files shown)


plugins/get_plugin_manifest.py
```diff
@@ -1,75 +1,33 @@
...
-        )
-        s3_client.put_object(Body=s3_body, Bucket=bucket, Key=key)
+        existing_manifest = s3.get_object(Bucket=bucket, Key=key)
+        return
+    except ClientError as e:
+        # we don't want to hide any "real" s3 errors
+        if not e.response['Error']['Code'] == 'NoSuchKey':
+            raise e        
+            
+        try:
```
klai95, line 28: Just want to clarify, is the except clause saying we would want to try `fetch_manifest` if e.response['Error']['Code'] == 'NoSuchKey' is `True`? Would like to get some context on the logic!

plugins/get_plugin_manifest.py
```diff
@@ -1,75 +1,33 @@
...
-        raise e
-    finally:
-        response = s3_client.delete_object(
-            Bucket=bucket,
-            Key=key
-        )
-        s3_client.put_object(Body=s3_body, Bucket=bucket, Key=key)
+        existing_manifest = s3.get_object(Bucket=bucket, Key=key)
+        return
+    except ClientError as e:
```
richaagarwal, line 23: I know boto3 doesn't make this very readable to do, but see this discussion, I tend to agree with the answer here: https://stackoverflow.com/a/34562141

plugins/get_plugin_manifest.py
DragaDoncila: @klai95 that's correct. The logic is that if we get `NoSuchKey` back from s3, the manifest file doesn't yet exist so we need to try to discover it. However, we don't want to hide other s3 errors e.g. `AccessDenied` - that's why we reraise the exception for all other error codes.

This logic will change based on Richa's comment below though.

plugins/get_plugin_manifest.py
DragaDoncila: @richaagarwal good point. I've updated this & the tests!

plugins/get_plugin_manifest.py
```diff
@@ -1,75 +1,31 @@
...
-        manifest = pm.get_manifest(plugin_name)
-        if not manifest.contributions:
-            raise RuntimeError(f"Manifest {plugin_name} exists but has no contributions.")
-    return manifest, is_npe2
+from botocore.exceptions import ClientError
+from npe2 import fetch_manifest
 
+s3 = boto3.resource('s3')
+# Environment variable set through ecs stack terraform module
+bucket_name = os.environ.get('BUCKET')
```
richaagarwal, line 9: If no `BUCKET` environment variable is set (though it's currently set, it's not guaranteed to remain that way), would we want to exit and raise an error at this stage, or is it preferred to have it fail at line 21?

DragaDoncila: @richaagarwal good point - I've updated so we raise a `RuntimeError` immediately if `bucket_name` is not set.

richaagarwal (approved):
This is great!  Really nice work on simplifying the logic and adding tests.

Issue #638, opened 2022-08-22
Prototype backend ETL & API for PyPI downloads


neuromusic:
- ETL pipeline that will dump aggregate data from snowflake to S3
- expose aggregate data in S3 to frontend via API endpoint

scope:
- does NOT need to be live/daily data for prototype (on demand or stale is fine)

Related front-end ticket(s): https://github.com/chanzuckerberg/napari-hub/issues/586

richaagarwal:
@codemonkey800 To provide expected data schema

@klai95 To write a snowflake query similar to what's provided in https://github.com/chanzuckerberg/napari-hub/issues/586#issuecomment-1195938647, but aggregated by month instead, and write that output to S3.  Maybe using snowflake python connector? https://docs.snowflake.com/en/user-guide/python-connector.html

richaagarwal:
@bnelson-czi Would you be able to provide snowflake credentials, via LastPass or some other place?  Looking at https://docs.snowflake.com/en/user-guide/python-connector-api.html#label-snowflake-connector-methods, it looks like we need a username, password, and account ID.  If possible, we should set up a generic `napari-hub` user with read-only permissions. Lmk if you'd like to discuss more!

bnelson-czi:
@richaagarwal I'd probably avoid using individual user credentials, but I can ask IT to create a service account. How does that sound?

bnelson-czi:
Looks like we can follow the steps here: https://docs.snowflake.com/en/user-guide/data-load-s3-config-storage-integration.html

richaagarwal:
@bnelson-czi Oh, interesting! Is the idea that we would write directly from Snowflake to S3 instead of using a python connector, and then read from those S3 files / write new ones to S3 as necessary? Who manages the snowflake writes to S3, in that case? 

Some concerns with this that come to mind:
1) It makes our stack a little more dependent on snowflake in terms of outside config/integrations - for this stage of work, it'd be nice to set up something more visible and easy for our existing team to manage (e.g. in python) rather than adding additional infra overhead. We'll definitely need to revisit our infra but we don't have the resources to do that right now.
2) If we need to do any ETL, we're reading in the S3 files and then writing additional ones, so it seems like this would add a step to the process. 

If we're not able to easily get some kind of user credentials to explore the python connector option, I'd say this work might be blocked and we need to re-assess our approach.  Curious if @neuromusic and @klai95 have any thoughts here.

bnelson-czi:
[...]

PR #651, opened 2022-09-06
Area chart fixture data

codemonkey800:
## Description

This PR adds fixture data for the activity dashboard of various plugins from @bnelson-czi's data dump:
https://github.com/chanzuckerberg/napari-hub/issues/586#issuecomment-1195938647

The main purpose of this data is to test rendering the area chart, so it's likely not correct. We can make an update in a future PR with correct data once #638 is resolved


frontend/src/fixtures/activity.json
```diff
@@ -0,0 +1,11650 @@
+{
+  "arcos-gui": {
+    "points": [
+      {
+        "x": 1627801200000,
```
klai95, line 5: Not sure if this was addressed elsewhere, but do you think it would be helpful to clarify what the values of x and y represent?

richaagarwal (approved):

PR #654, opened 2022-09-09
Fix area chart client-side API requests

### LINKED ISSUE #652: Staging is down because of webpack config
#### Description
A new webpack configuration option was added in #650 that is causing the staging build to be down. The offending line is here:

https://github.com/chanzuckerberg/napari-hub/blob/5871592cca401cb4926ce685dc4c2899314e62f4/frontend/next.config.js#L70

This was originally added so that the frontend could make client-side requests to the API server. However, that means we need to include the URL at build time. However, the `API_URL` variable is not provided in the environment until runtime. In the past, this was fine because the only place we used `API_URL` was on the server. 

Since the `API_URL` isn't provided at build time, we'll need to fix this by proxying the API request through Next.js.

#### Steps/Code to Reproduce
Visit https://staging.napari-hub.org/

#### Expected Results
Staging should work

#### Actual Results
image

codemonkey800:
Closes #652 

This works by proxying client-side API requests through the Next.js server since the `API_URL` is available at runtime instead of build time.

### DIFF
--- a/frontend/mock-server.js
+++ b/frontend/mock-server.js
@@ -80,17 +80,17 @@ app.get('/activity/:plugin/stats', (req, res) => {
   if (stats) {
     res.json(stats);
   } else {
-    res.status(404);
+    res.status(404).send('not found');
   }
 });
 
 app.get('/activity/:plugin', async (req, res) => {
-  const stats = get(activity, [req.params.plugin, 'points'], null);
+  const points = get(activity, [req.params.plugin, 'points'], null);
 
-  if (stats) {
-    res.json(stats);
+  if (points) {
+    res.json(points);
   } else {
-    res.status(404);
+    res.status(404).send('not found');
   }
 });
 
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -67,8 +67,6 @@ module.exports = {
 
     config.plugins.push(
       new EnvironmentPlugin({
-        API_URL: '',
-
         BASE_PATH: '',
 
         // Path to JSON file that has the same structure as the backend's plugin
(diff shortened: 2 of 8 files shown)


richaagarwal (dismissed):
@codemonkey800 It looks like `API_URL` is currently defined in several places. `BACKEND_URL` is also the same value as `API_URL`, and is hardcoded in a few places. Maybe this is a good opportunity to try to consolidate them.

I also noticed that you defined the values for staging and prod in this PR: https://github.com/chanzuckerberg/napari-hub/pull/628/files#diff-da6498268e99511d9ba0df3c13e439d10556a812881c9d03955b2ef7c6c1c655R13-R14. We could define these depending on the prod or staging environment, but it looks like this gets slightly trickier with our dev environments. For those, the naming would be `https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/` + `{branch_name}`, which I don't believe we currently define anywhere in our CI.  But I don't see why we couldn't.

Let's try to ensure `API_URL` is available at build time so that we don't have to add all this additional code.

richaagarwal:
One last thing, is there a reason this line introduced in an earlier PR works and we wouldn't be able to do something similar now? https://github.com/chanzuckerberg/napari-hub/blame/main/frontend/src/utils/HubAPIClient.ts#L19

richaagarwal:
@codemonkey800 Consolidation of variables was just one suggestion and I think pull requests are a good place to give feedback; in this case I agree it's not critical, but I found it a little confusing. In terms of creating follow-up issues, if you find any feedback useful or valuable for future work, I'd ask that you create those issues. If my feedback isn't making sense or you don't agree with it, also happy to chat more (perhaps over a call), though again it doesn't need to be blocking feedback.  

That said, consolidation of variables was only one part of my feedback.  I think we can and should try to define this as build time.  Ideally we would revert the line in your PR that is causing this issue, since we're not actually directly calling the API yet since the back-end work is in progress.  However, I will go ahead and approve this for now to give more wiggle room with some of these conversations and figuring out next steps.

richaagarwal (approved):

PR #659, opened 2022-09-13
Fix category chip tooltip for tap devices

### LINKED ISSUE #413: [Mobile] When tapping on (i) in category pill in homepage, category pill is set as a filter instead of only displaying the tooltip
#### Description
As tooltip is a button, and the chip is a button, so we're storing a button within a button and it's clicking it twice

#### Steps/Code to Reproduce
1. Go to homepage
2. Search for a plugin including category pills
3. Tap on any (i) in a category pill

#### Expected Results
Tolltip is displayed

#### Actual Results
Category is set as filter and tooltip is displayed afterwards

https://user-images.githubusercontent.com/63799148/151427153-fb9f2360-22d9-4fac-8b10-1994e25a88da.mov

codemonkey800:
## Description

Closes #413 

Fixes the issue with the tooltip not being openable on mobile / tap devices. This works by changing the layout of the tooltip to be for the entire chip on tap devices so that users can long press for the tooltip. Unfortunately, this is not ideal like the info icon, but is the typical pattern used for tooltips on tap devices

## Demo

https://user-images.githubusercontent.com/2176050/190026422-ab5dc601-58da-4f51-a8a9-ff1fcf93f275.mp4

### DIFF
--- a/frontend/src/components/CategoryChip/CategoryChip.tsx
+++ b/frontend/src/components/CategoryChip/CategoryChip.tsx
@@ -3,9 +3,17 @@ import Chip, { ChipProps } from '@mui/material/Chip';
 import Tooltip from '@mui/material/Tooltip';
 import clsx from 'clsx';
 import { useTranslation } from 'next-i18next';
-import { ForwardedRef, forwardRef, Fragment, useRef, useState } from 'react';
+import {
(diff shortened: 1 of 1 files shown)


klai95 (commented):
Looks good to me!

PR #661, opened 2022-09-14
Use HubSpot for signup form

### LINKED ISSUE #644: Live dev site with mailing list signup form sending info to HubSpot instead of MailChimp

codemonkey800:
## Description

Closes #644

Refactors the signup form to use the hub spot form API instead of MailChip. This works by using the HubSpot API to create an invisible form in the `SignupForm` component. We then use DOM APIs to update the form and submit it.

This feels a bit hacky, but is the suggested approach in the [docs](https://legacydocs.hubspot.com/docs/methods/forms/advanced_form_options):

image

## Demo

https://user-images.githubusercontent.com/2176050/190032829-bfbc9683-fc12-4647-9a67-00d8553ce04d.mp4

### DIFF
--- a/frontend/src/global.d.ts
+++ b/frontend/src/global.d.ts
@@ -14,6 +14,10 @@ declare namespace NodeJS {
 }
 
 namespace HubSpotFormAPI {
+  /**
+   * HubSpot form creation options.
+   * https://legacydocs.hubspot.com/docs/methods/forms/advanced_form_options
+   */
   export interface CreateFormOptions {
     region: string;
     portalId: string;


frontend/src/components/SignupForm/SignupForm.tsx
```diff
@@ -48,85 +45,144 @@ export function SignupForm({ onSubmit, variant = 'default' }: Props) {
...
+
   return (
     <ColumnLayout
       className="bg-napari-light p-sds-xl screen-495:p-12"
       // Use 3-column layout instead of 4-column layout.
       classes={isHome ? { fourColumn: 'screen-1150:grid-cols-napari-3' } : {}}
     >
+      <Script
+        onLoad={() => setIsHubSpotReady(true)}
+        src="//js.hsforms.net/forms/v2.js?pre=1"
```
neuromusic, line 115: this doesn't set any cookies does it?

frontend/src/global.d.ts
```diff
@@ -12,3 +12,21 @@ declare namespace NodeJS {
     readonly PREVIEW_PULL_REQUEST: string;
   }
 }
+
+namespace HubSpotFormAPI {
+  export interface CreateFormOptions {
+    region: string;
+    portalId: string;
+    formId: string;
+    target?: string;
```
klai95, line 21: Could you clarify what `field` is and why it is optional?

klai95 (approved):
I left one clarifying question, but otherwise looks good to me!

neuromusic:
@codemonkey800 has verified that the new form is functional and sending data to Mailchimp on this dev branch

klai95:
@codemonkey800 @neuromusic After a user subscribes via their email, if they were to do the same thing again, would it tell them the email address has already been subscribed, or would it show "Thank you for joining our newsletter" again?

PR #667, opened 2022-09-22
Pin Python dependencies 

richaagarwal:
We've been having intermittent issues with our build process lately, and pinning our python dependencies will be one step towards more consistent, repeatable builds.  I'll cover the plugin requirements in a separate PR.


klai95 (approved):
Looks good to me! My only question is: are the specified versions here ever going to change?

richaagarwal:
> Looks good to me! My only question is: are the specified versions here ever going to change?

That's a good question, @klai95!  We have [Dependabot](url) enabled on this repository, which is configured so that we'll receive alerts and an automatic pull request when security vulnerabilities are found with any of our packages.

For non-critical updates, though, we'd update versions as needed: I'd imagine we'll do this periodically if required for a new dependency that requires other dependencies to be upgraded, or if a new release has a particular feature we want, but otherwise it's not necessary to have the latest version.

PR #669, opened 2022-09-26
Fix staging and prod builds

richaagarwal:
This is a follow-up to https://github.com/chanzuckerberg/napari-hub/pull/668, which only addressed the issue for dev builds.  This updates the stack to `1.7.4` and and specifies the "create-or-update" operation for staging and production workflows as well.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -29,7 +29,7 @@ jobs:
         run: | 
           echo DOCKER_TAG=${GITHUB_REF_NAME} >> $GITHUB_ENV
       - name: Deploy Changes to Prod
-        uses: chanzuckerberg/github-actions/.github/actions/deploy-happy-stack@v1.4.0
+        uses: chanzuckerberg/github-actions/.github/actions/deploy-happy-stack@v1.7.4
         env:
           ENV: prod
         with:
@@ -38,6 +38,7 @@ jobs:
           tag: ${{ env.DOCKER_TAG }}
           tfe-token: ${{ secrets.TFE_TOKEN }}
           env: "prod"
+          operation: "create-or-update"
   staging:
     name: deploy to staging
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -29,7 +29,7 @@ jobs:
         run: | 
           echo DOCKER_TAG=${GITHUB_REF_NAME} >> $GITHUB_ENV
       - name: Deploy Changes to Prod
-        uses: chanzuckerberg/github-actions/.github/actions/deploy-happy-stack@v1.4.0
+        uses: chanzuckerberg/github-actions/.github/actions/deploy-happy-stack@v1.7.4
```
klai95, line 32: Just a clarifying question: how/what does version 1.7.4 address the problem we have been experiencing?

.github/workflows/deploy.yml
```diff
@@ -60,7 +61,7 @@ jobs:
           echo STACK_NAME=${GITHUB_REF_NAME} >> $GITHUB_ENV
           echo DOCKER_TAG=${GITHUB_REF_NAME} >> $GITHUB_ENV
       - name: Deploy Changes to Staging
-        uses: chanzuckerberg/github-actions/.github/actions/deploy-happy-stack@v1.4.0
+        uses: chanzuckerberg/github-actions/.github/actions/deploy-happy-stack@v1.7.4
```
klai95, line 64: same question here

.github/workflows/deploy.yml
richaagarwal: @klai95 That's a good question!  It looks like when we looped in infra about this issue, they introduced a fix here: https://github.com/chanzuckerberg/github-actions/pull/147 and released that to a new version, `1.7.4` here: https://github.com/chanzuckerberg/github-actions/pull/144

.github/workflows/deploy.yml
klai95: will mark as resolved per comment addressed above

.github/workflows/deploy.yml
klai95: @richaagarwal Thanks, that makes sense! Just wanted to ask the question so we have some context/artifact to look back on that explains how the issue was fixed!

klai95 (approved):
Looks good to me!

Issue #620, opened 2022-07-28
Update `plugins` lambda and its triggers to remove chance of infinite loop when fetching `npe2` data


DragaDoncila:
### Summary
This issue captures the work required to complete TODOs 1 and 2 in #587. This work item is composed of two subtasks:

1. Disable any retry mechanism for failed manifest discovery
2. Update `plugins` lambda trigger to remove chance of file changes retriggering the same lambda

**Rationale**: we think it unlikely that discovery failures would be fixed on a retry, and thus a retry mechanism would be useful only in the case where we failed due to an external reason e.g. PyPI was temporarily down and we couldn't download the package. We can consider implementing a retry mechanism if monitoring reveals a need for it.

### Intended Outcome
Plugins lambda can update successfully built manifest files for new plugins without the potential for an infinite loop.

### How will it work?

Disabling the retry mechanism will involve removing the `failure` lambda entirely and ensuring the `plugins` lambda doesn't error and is not configured for asynchronous retries. 

We still need to decide how to update the triggering mechanism e.g.:

1. Keep the file creation trigger for the lambda, but rather than relying on `process_count` in some plugin's `{plugin}.{version}-manifest.json`, we write a separate (could be blank) file e.g. `trigger.json` to serve as a trigger for the `plugins` lambda. Lambda proceeds as normal discovering manifest. If discovery is successful, lambda writes **correct** manifest to `{plugin}.{version}-manifest.json` and deletes `trigger.json`. If discovery is unsuccessful, `{plugin}.{version}-manifest.json` is **not written** at all (but trigger file is updated with error message?).
2. Completely remove s3 file creation as a trigger for the `plugins` lambda and instead invoke it directly in the backend cache updating methods. In **very rough** pseudocode terms, in `backend/api/model/build_manifest_metadata`:

```python
if "{plugin}/{version}-manifest.json" not in cache and "{plugin}-{version}" not in failed_manifests:
    # we'd want to do some check that the lambda isn't already in action?
    boto3_client.invoke_async(FunctionName='plugins', InvokeArgs='plugin, version')
```

Then, in `plugins/get_plugin_manifest/generate_manifest`:

```python
try:
    manifest = build_manifest(plugin, version)
    if manifest:
        s3.put(manifest, '{plugin}.{version}-manifest.json')
     else:
        update_failed_manifests(plugin, version)
except Exception as e:
        update_failed_manifests(plugin, version)
```

richaagarwal:
[...]

PR #673, opened 2022-10-03
Invoke `plugins` lambda from backend for `npe2` manifest discovery

DragaDoncila:
As described in #620 this PR updates the manifest API endpoint handler to directly invoke the `plugins` lambda when the manifest file for a plugin & version is not already cached.

Check out the metadata for `napari-blossom` on the frontend:
https://dev-npe2-triggers-frontend.dev.imaging.cziscience.com/plugins/napari-blossom


DragaDoncila:
I've now let this branch run on all plugins in `dev-npe2-triggers` for the first time. Just wanted to leave a few notes here:

- The initial cloudwatch update triggered 235 lambda invocations - this is consistent with the number of currently listed plugins (232 - I have no idea where the extra 3 come from...).
- Of 235 invocations, 13 errored. All errors were timeout/max memory usage errors from the lambda. This was semi-expected as I had just lowered the resource allocations to their minimums for testing purposes. All failing plugins do not provide a `wheel` distribution, so the fetch process has to build it (hence the increased resource requirements).
- Even with the 13 invocations timing out at 60 seconds, the average duration of the initial run of 235 invocations was 4.3s.
- After ~25 minutes, all but 19 plugins had complete metadata information in `index.json` - roughly consistent with the 13 errored lambdas, though I will look into the other 6.

The frustrating thing is that despite the lambda being configured not to retry, 12 of the 13 initially failing invocations just kept being reinvoked - more than 6 times?! I can't explain that, honestly. They just keep going?

I will upgrade the memory allocation and see how that affects the remaining failing invocations, and investigate the 6 plugin difference missing in `index.json`.

DragaDoncila:
Increased memory size to 256MB and plugins are now still timing out (but memory usage reported at 200MB). Will update timeout to 2 min.

DragaDoncila:
Ok after those updates all lambdas completed successfully. We still have 12 plugins with no manifest metadata in `index.json` and I've summarized them below:

## Error message written to manifest

These might need to be retried now that I've updated memory size.

napari-yolov5 - no space left on device (lambda memory error)
platelet-unet-watershed - [Errno 2] No usable temporary directory found in ['/tmp', '/var/tmp', '/usr/tmp', '/var/task']
zarpaint - [Errno 2] No usable temporary directory found in ['/tmp', '/var/tmp', '/usr/tmp', '/var/task']
nfinder - [Errno 2] No usable temporary directory found in ['/tmp', '/var/tmp', '/usr/tmp', '/var/task']

## Missing in npe2api - legitimate errors

skeleton-finder
napari-mri

## Actually 'empty' manifests

napari-error-reporter - napari package, not a plugin
napari-console - napari package, not a plugin
beetlesafari - accurate, doesn't return any hook implementations
devbio-napari - accurate, napari tools menu stuff

## Potential fetch error
napari-mat-images - does have a hook_implementation, fetch finds no contributions
napari-features - does have a hook_implementation, fetch finds no contributions

richaagarwal:
@DragaDoncila Thanks so much for these updates!  Regarding the 12 plugins with no manifest data/ that failed, did the retries (even though they're not supposed to happen at all) eventually settle to around 6 times or are they still being retried?  

Also - let me know when this PR is ready for review!

DragaDoncila:
> Regarding the 12 plugins with no manifest data/ that failed, did the retries (even though they're not supposed to happen at all) eventually settle to around 6 times or are they still being retried?

@richaagarwal just summarizing here from our discussion today. All lambdas ceased to be retried after I updated the resource allocations, as there were no external lambda errors e.g. timeouts or memory issues. The reason they were retrying is *not* default aws behavior, it was because our manifest files were never being written, so our cloudwatch cron job kept leading to the lambda being invoked.

As discussed, this is not great as we're relying on file existence as a measure of idempotency. I've now updated the lambda handler to write an empty file at the beginning of discovery, and overwrite it with either an error message or the real manifest once discovery is complete. This means even an unsuccessful lambda termination doesn't lead to retry, as the file will exist. I've also updated the error messages returned from the API to reflect this.

As a future piece of work if we don't like the fact that we're overwriting the file, we can update the python handler to time its own execution and write an error message just before the lambda would time out.

Finally we discussed cleaning the `prod` and `staging` environments of existing manifest files placed there by the original system, and we decided it would be fine to manually run a command to clear these out from `staging` just before we merge this PR, and from prod soon after. This will allow the new method to populate all manifests correctly. 

For people's reference, the command to print all files that would be deleted is:

```sh
aws s3 rm s3://napari-hub-dev/dev-npe2-triggers/cache/ --recursive --exclude "*" --include "*-manifest.json" --dryrun
```

I've run this command in the dev branch bucket to watch another run of invocations on the full plugin list. This time a single plugin timed out (`napari-yolov5`) while attempting to build the wheel, and I've confirmed that the `plugins` lambda doesn't get invoked again. I've updated the timeout to 2.5min (as a subsequent run for this plugin *almost* finished), and hopefully that will take care of this last plugin.

Regardless, I think this PR is ready for review!

DragaDoncila:
small update: `napari-yolov5` finished successfully in 124s - so close 😆

backend/api/model.py
```diff
@@ -95,11 +98,27 @@ def get_manifest(plugin: str, version: str = None) -> dict:
...
+    # plugin_metadata being None indicates manifest is not cached and needs processing 
+    if plugin_metadata is None:
+        client = boto3.client('lambda')
+        lambda_event = {'plugin': plugin, 'version': version}
+        response = client.invoke(
+            FunctionName=os.environ.get('PLUGINS_LAMBDA_NAME'),
+            InvocationType='Event',
+            Payload=json.dumps(lambda_event),
+        )
+        return {'processed': False}
```
richaagarwal, line 111: I find the naming of `processed` here confusing - based on the name, I'd assume a `False` value to mean that processing was unsuccessful. Should it be `True` here and `False` below?

richaagarwal: Ah, nevermind, I see it means something else. One idea is to remove the `processed` field entirely and return the specific error message here that you end up defining above, e.g. setting `error` to "Plugin Manifest Not Found. Manifest discovery failed."

backend/api/model.py
```diff
@@ -95,11 +98,27 @@ def get_manifest(plugin: str, version: str = None) -> dict:
...
-        return plugin_metadata
-    else:
-        cache({"process_count": 0}, f'cache/{plugin}/{version}-manifest.json')
-        return {"process_count": 0}
+    
+    # plugin_metadata being None indicates manifest is not cached and needs processing 
+    if plugin_metadata is None:
+        client = boto3.client('lambda')
+        lambda_event = {'plugin': plugin, 'version': version}
+        response = client.invoke(
```
richaagarwal, line 106: Leaving a comment here that this invocation ultimately calls `generate_manifest` could make this section easier to follow.

richaagarwal: Also, if we're not doing anything with `response` since this is presumably an asynchronous call?, then it could be helpful to remove it to make it more explicit that we're not waiting for the response.

richaagarwal:
@DragaDoncila I left some comments, but after reviewing further, I realized there are a few things with the previous structure we had in place that might be problematic with this new code. 

As the PR currently is, the plugins lambda could theoretically be invoked outside of the cron job, which isn't ideal. We'd ideally avoid that possibility by separating out calls to the S3 cache vs. calls to `get_manifest`, since `get_manifest` is now directly invoking the plugins lambda.

backend/api/model.py
DragaDoncila: @richaagarwal good point, I've updated this code now to remove the `processed` field entirely and just rely on the `error` reported.

DragaDoncila:
@richaagarwal thanks for the thorough review! I've updated the logic now so that `get_manifest` *just* retrieves the manifest from `cache`. I've added a new method `discover_manifest` which is *only* called from `build_manifest_metadata` when `get_manifest` returns a `"Manifest not yet processed"` message. 

I've verified that the `plugins` lambda is no longer invoked when, for example, we make a call to `manifest/{name}` in the api, but that it is successfully invoked during the cron job.

backend/api/app.py
```diff
@@ -69,24 +69,23 @@ def versioned_plugin(plugin: str, version: str = None) -> Response:
...
-    else:
+    error = manifest['error']
+    if error == 'Manifest not yet processed.':
         response = app.make_response(("Temporarily Unavailable. Attempting to build manifest. Please check back"
                                       " in 5 minutes.", 503))
         response.headers["Retry-After"] = 300
         return response
+    else:
+        return app.make_response(
+            ("Plugin Manifest Not Found. Manifest discovery failed.", 404))
```
DragaDoncila, line 88: @richaagarwal we could include the `manifest['error']` message in this response, but I wasn't sure if we wanted to publicly expose it. I don't see any issue with doing so, but wanted to check with you first.

richaagarwal: I don't think we need to for now! But I'll defer to your judgment

richaagarwal (approved):
This looks great to me! Thanks so much for your thorough work & testing on this. The code comments really help with readability too 💛

I wonder if we should also add `staging` to the feature flags here: https://github.com/chanzuckerberg/napari-hub/blob/main/frontend/src/utils/featureFlags.ts#L27. I'll leave it up to you if you want to add that before merging or not, or if you prefer to open a separate PR for that.

DragaDoncila:
@richaagarwal I'll update the feature flag here to enable on `staging`. Before I merge I'll also delete the `-manifest.json` files on `staging`.

DragaDoncila:
Synced with @richaagarwal about failing test and as it seems unrelated, I will go ahead and merge this now!

PR #675, opened 2022-10-05
Data validation for citation

klai95:
Per the [stack trace](https://app.datadoghq.com/apm/error-tracking?query=&issueId=ecb2d880-442e-11ed-89d9-da7ad0900002&start=1664985275287&end=1664988875287&paused=false) the line that caused the issue is: https://github.com/chanzuckerberg/napari-hub/blob/main/backend/utils/github.py#L170. This PR will handle the data validation for the issue and ensure Datadog no longer contains errors.

### DIFF
--- a/backend/utils/github.py
+++ b/backend/utils/github.py
@@ -2,7 +2,6 @@ import json
 import logging
 import os.path
 import re
-import traceback
 from typing import Dict, Union, IO
 
 import requests
--- a/backend/utils/github.py
+++ b/backend/utils/github.py
@@ -207,6 +207,7 @@ def get_citation_author(citation_str: str) -> Union[Dict[str, str], None]:
         citation_yaml = yaml.safe_load(citation_str)
     except yaml.YAMLError as e:
         logging.error(e)
+        return []
     authors = []
     for author_entry in citation_yaml['authors']:
         if 'given-names' in author_entry and 'family-names' in author_entry and author_entry['given-names'] and author_entry['family-names']:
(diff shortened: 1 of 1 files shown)


backend/utils/github.py
```diff
@@ -201,6 +204,8 @@ def get_citation_author(citation_str: str) -> Union[Dict[str, str], None]:
     :return: list of mappings between the string 'name' and the author name
     """
     citation_yaml = yaml.safe_load(citation_str)
+    if citation_yaml is None:
+        return {}
```
richaagarwal, line 208: @klai95 I think this should be `return []` since this function is returning a list of authors or not.

klai95: Yup made the change thanks for the catch

backend/utils/github.py
```diff
@@ -200,7 +203,10 @@ def get_citation_author(citation_str: str) -> Union[Dict[str, str], None]:
     :param citation_str: citation string to parse
     :return: list of mappings between the string 'name' and the author name
     """
-    citation_yaml = yaml.safe_load(citation_str)
+    try:
+        citation_yaml = yaml.safe_load(citation_str)
```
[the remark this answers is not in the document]
klai95: I think we should wrap all the code in try and below it include the except clause right?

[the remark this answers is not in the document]
richaagarwal: Yeah, I think you need to add a `return []` to that. I think that's preferable to moving more code inside the try/except block.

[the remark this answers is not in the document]
klai95: I don't see any errors as of now. I will monitor it to ensure everything works

backend/utils/github.py
```diff
@@ -1,6 +1,8 @@
 import json
+import logging
 import os.path
 import re
+import traceback
```
richaagarwal, line 5: Let's remove this now unused import

klai95: Thanks for the catch, just made the change!

richaagarwal (approved):

Issue #664, opened 2022-09-19
[Activity dashboard] Pull PyPI downloads data nightly 


klai95:
In https://github.com/chanzuckerberg/napari-hub/issues/638, we introduced API endpoints that connect to PyPI downloads data.  As part of that work, we did a one-time pull of data from Snowflake.  Now, let's make that a recurring cron job that runs once every 24 hours.

### Additional details
- Even though activity dashboard functionality is currently behind a feature flag, let's develop this data integration such that it runs successfully in all our environments (dev, staging and prod)
- The underlying data should be getting updated once a day, and @bnelson-czi 's job runs at 3am PT for ~10 minutes, so maybe 5am PT or after is a good time to target.
- We will be introducing other activity dashboard metrics in the future, so consider a folder/naming structure such as /activity-dashboard-metrics/pypi-downloads.csv

Earlier notes from Kevin
- The follow-up work to tackle improvements to the snowflake code as well as connecting it to a cron job so that we don't need to keep re-running it; see more at https://github.com/chanzuckerberg/napari-hub/pull/662 for the activity dashboard prototype.
- As such, we can integrate the read and write of data between `Snowflake` and `S3` in our existing infrastructure that enables for automatic updates. Further more, additional data sources can be integrated seamlessly as well should we ever need to pull from new external sources.

klai95:
Below is the code to generate Snowflake data; as part of this issue, I will need to inject this SQL code directly in the backend codebase.
```
select file_project, date_trunc('month', timestamp) as month, count(*) as num_downloads
from
(
  select country_code, project, project_type, file_project, file_version, file_type, details_installer_name, details_installer_version,
      CONCAT(
        DETAILS,
        DETAILS_INSTALLER, DETAILS_INSTALLER_NAME, DETAILS_INSTALLER_VERSION,
        DETAILS_PYTHON,
        DETAILS_IMPLEMENTATION_NAME, DETAILS_IMPLEMENTATION_VERSION,
        DETAILS_DISTRO, DETAILS_DISTRO_NAME, DETAILS_DISTRO_VERSION, DETAILS_DISTRO_ID, DETAILS_DISTRO_LIBC, DETAILS_DISTRO_LIBC_LIB, DETAILS_DISTRO_LIBC_VERSION,
        DETAILS_SYSTEM, DETAILS_SYSTEM_NAME, DETAILS_SYSTEM_RELEASE,
        DETAILS_CPU,
[...]

PR #710, opened 2022-10-25
[Activity dashboard] Pull PyPI downloads data nightly

klai95:
Closes https://github.com/chanzuckerberg/napari-hub/issues/664

Screen Shot 2022-10-26 at 4 14 07 PM
Screen Shot 2022-10-26 at 4 14 42 PM

### DIFF
--- a/.happy/terraform/modules/ecs-stack/main.tf
+++ b/.happy/terraform/modules/ecs-stack/main.tf
@@ -199,11 +199,11 @@ locals {
       service    = "apigateway"
       source_arn = "${local.execution_arn}*"
     },
-    AllowExecutionFromCloudWatch = {
+    AllowExecutionFromCloudWatchForPlugin = {
       service    = "events"
       source_arn = aws_cloudwatch_event_rule.update_rule.arn
     },
-    AllowExecutionFromCloudWatch = {
+    AllowExecutionFromCloudWatchForActivity = {
       service    = "events"
       source_arn = aws_cloudwatch_event_rule.activity_rule.arn
     }
--- a/.happy/terraform/modules/ecs-stack/main.tf
+++ b/.happy/terraform/modules/ecs-stack/main.tf
@@ -181,7 +181,7 @@ resource "aws_cloudwatch_event_target" "update_target" {
 resource "aws_cloudwatch_event_rule" "activity_rule" {
   name                = "${var.env}-${local.custom_stack_name}-activity"
   description         = "Schedule update for activity data"
-  schedule_expression = "rate(5 minutes)"
+  schedule_expression = "cron(0 13 * * ? *)"
   tags                = var.tags
 }
 
--- a/.happy/terraform/modules/ecs-stack/main.tf
+++ b/.happy/terraform/modules/ecs-stack/main.tf
@@ -202,6 +202,10 @@ locals {
     AllowExecutionFromCloudWatch = {
       service    = "events"
       source_arn = aws_cloudwatch_event_rule.update_rule.arn
+    },
+    AllowExecutionFromCloudWatch = {
+      service    = "events"
+      source_arn = aws_cloudwatch_event_rule.activity_rule.arn
     }
   }
 }
--- a/.happy/terraform/modules/ecs-stack/main.tf
+++ b/.happy/terraform/modules/ecs-stack/main.tf
@@ -163,7 +163,7 @@ module api_gateway_proxy_stage {
 
 # Cron job running the update endpoint
 resource "aws_cloudwatch_event_rule" "update_rule" {
-  name                = "${var.env}-${local.custom_stack_name}-update"
+  name                = "${local.custom_stack_name}-update"
   description         = "Schedule update for backend"
   schedule_expression = "rate(5 minutes)"
   tags                = var.tags
@@ -179,7 +179,7 @@ resource "aws_cloudwatch_event_target" "update_target" {
 
 # Cron job updating the activity data
 resource "aws_cloudwatch_event_rule" "activity_rule" {
-  name                = "${var.env}-${local.custom_stack_name}-activity"
+  name                = "${local.custom_stack_name}-activity"
   description         = "Schedule update for activity data"
   schedule_expression = "cron(0 13 * * ? *)"
   tags                = var.tags
@@ -267,13 +267,13 @@ data aws_iam_policy_document plugins_policy {
 }
 
 resource aws_iam_role_policy policy {
-  name     = "${local.custom_stack_name}-${var.env}-policy"
+  name     = "${local.custom_stack_name}-policy"
   role     = module.backend_lambda.role_name
   policy   = data.aws_iam_policy_document.backend_policy.json
 }
 
 resource aws_iam_role_policy plugins_lambda_policy {
-  name     = "${local.custom_stack_name}-${var.env}-plugins-lambda-policy"
+  name     = "${local.custom_stack_name}-plugins-lambda-policy"
   role     = module.plugins_lambda.role_name
   policy   = data.aws_iam_policy_document.plugins_policy.json
 }
--- a/.happy/terraform/modules/ecs-stack/main.tf
+++ b/.happy/terraform/modules/ecs-stack/main.tf
@@ -177,6 +177,22 @@ resource "aws_cloudwatch_event_target" "update_target" {
     }
 }
 
+# Cron job updating the activity data
+#resource "aws_cloudwatch_event_rule" "activity_rule" {
+#  name                = "${var.env}-${local.custom_stack_name}-activity"
+#  description         = "Schedule update for activity data"
+#  schedule_expression = "rate(1 day)"
+#  tags                = var.tags
+#}
+
+#resource "aws_cloudwatch_event_target" "activity_target" {
+#    rule = aws_cloudwatch_event_rule.activity_rule.name
+#    arn = module.backend_lambda.function_arn
+#    input_transformer {
+#        input_template = jsonencode({path = "/activity/update", httpMethod = "POST"})
+#    }
+#}
+
 locals {
   allowed_triggers = {
     LambdaPermission = {
--- a/.happy/terraform/modules/ecs-stack/main.tf
+++ b/.happy/terraform/modules/ecs-stack/main.tf
@@ -178,20 +178,20 @@ resource "aws_cloudwatch_event_target" "update_target" {
 }
 
 # Cron job updating the activity data
-#resource "aws_cloudwatch_event_rule" "activity_rule" {
-#  name                = "${var.env}-${local.custom_stack_name}-activity"
-#  description         = "Schedule update for activity data"
-#  schedule_expression = "rate(1 day)"
-#  tags                = var.tags
-#}
+resource "aws_cloudwatch_event_rule" "activity_rule" {
+  name                = "${var.env}-${local.custom_stack_name}-activity"
+  description         = "Schedule update for activity data"
+  schedule_expression = "rate(5 minutes)"
+  tags                = var.tags
+}
 
-#resource "aws_cloudwatch_event_target" "activity_target" {
-#    rule = aws_cloudwatch_event_rule.activity_rule.name
-#    arn = module.backend_lambda.function_arn
-#    input_transformer {
-#        input_template = jsonencode({path = "/activity/update", httpMethod = "POST"})
-#    }
-#}
+resource "aws_cloudwatch_event_target" "activity_target" {
+    rule = aws_cloudwatch_event_rule.activity_rule.name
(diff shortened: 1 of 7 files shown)


.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -177,6 +177,22 @@ resource "aws_cloudwatch_event_target" "update_target" {
     }
 }
 
+# Cron job updating the activity data
```
klai95, line 180: I have commented out this piece of code for now, in case unexpected errors keep getting triggered. I will uncomment this piece of code once we have the Snowflake credentials pipeline completed from @manasaV3.

backend/api/model.py
```diff
@@ -374,6 +375,59 @@ def get_categories_mapping(version: str) -> Dict[str, List]:
...
+    # connect to Snowflake DB
+    ctx = sc.connect(
+        user=SNOWFLAKE_USER,
+        password=SNOWFLAKE_PASSWORD,
+        account="CZI-IMAGING",
+        warehouse="IMAGING",
+        database="IMAGING",
+        schema="PYPI"
+    )
+    # TODO (klai): Use Snowflake stored procedures to encapsulate the SQL code so that it is not publicly exposed
```
klai95, line 395: As noted in my comment, using stored procedures would probably be the best practice here so that we are not exposing the sql code publicly. See more here: https://docs.snowflake.com/en/sql-reference/stored-procedures-overview.html

klai95:
I have already tested this PR locally via running `python -m api.app` on the backend, and sending the `http://127.0.0.1:12345/activity/update` request on Postman. As a result, I am able to, on demand, verify that the overwrite / update operation works as expected. I had hard-coded the file to be created in `napari-hub-dev`, but the code at the moment works generally. Here is `activity_dashboard.csv`: https://s3.console.aws.amazon.com/s3/buckets/napari-hub-dev?region=us-west-2&tab=objects

klai95:
Currently, `preview/_tests/test_preview_meta.py::test_parse_preview_matches_hub` is the only test that fails in `Backend Tests / pytest (pull_request)`, which is a separate issue from this PR. For context, the reason for this failure is that the npe2 work has been released, but the current implementation does not have updated preview page code to handle npe2 metadata. This work will be captured in an upcoming PR by @DragaDoncila.

backend/api/s3.py
```diff
@@ -63,6 +63,10 @@ def cache(content: Union[dict, list, IO[bytes]], key: str, mime: str = None):
                                      Key=os.path.join(bucket_path, key), ExtraArgs=extra_args)
 
 
+def write_activity_data(csv_string: str):
+    s3_client.put_object(Body=csv_string, Bucket=bucket, Key=os.path.join(bucket_path, "activity_dashboard.csv"))
```
klai95, line 67: If it is preferred to restructure the file name and structure, I can also make that change. For example (not to be taken literally), `pypi_downloads.csv` would be located in `napari-hub[env]/activity_dashboard_data`, so that we can have other files in the same subfolder.

.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -177,6 +177,22 @@ resource "aws_cloudwatch_event_target" "update_target" {
...
+#resource "aws_cloudwatch_event_rule" "activity_rule" {
+#  name                = "${var.env}-${local.custom_stack_name}-activity"
+#  description         = "Schedule update for activity data"
+#  schedule_expression = "rate(1 day)"
```
manasaV3, line 184: Do we want to leverage a cron job like `cron(0 6 * * ? *)` for triggering the call everyday right after the data update is run?

backend/api/s3.py
klai95: @manasaV3 Sounds good, we can discuss further how exactly we want to structure the file name and file path in `s3`. Do you have any suggestions? @richaagarwal

.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -177,6 +177,22 @@ resource "aws_cloudwatch_event_target" "update_target" {
...
+#  tags                = var.tags
+#}
+
+#resource "aws_cloudwatch_event_target" "activity_target" {
+#    rule = aws_cloudwatch_event_rule.activity_rule.name
+#    arn = module.backend_lambda.function_arn
+#    input_transformer {
+#        input_template = jsonencode({path = "/activity/update", httpMethod = "POST"})
+#    }
```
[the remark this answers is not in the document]
klai95: The format is the same as the `/update` endpoint, which is also a `POST` one. The only difference is that the `/update` endpoint updates every 5 minutes, whereas the `/activity/update` endpoint updates every 1 day.

.happy/terraform/modules/ecs-stack/main.tf
klai95: We do want something like that, but based on https://docs.aws.amazon.com/systems-manager/latest/userguide/reference-cron-and-rate-expressions.html and after talking to @bnelson-czi, he said he is running data science pipelines at 3am PT. The `Hours` parameter is measured in UTC time, so I think we want `cron(0 13 * * ? *)`, which would translate to running backend services at 6am PT daily.

backend/api/model.py
klai95: @neuromusic Curious if this is necessary. If not, I can just delete this todo and update the PR.

neuromusic: I don't feel like publicizing our schema via a SQL query is a major security risk

klai95:
After code pairing with @manasaV3, we have verified that the Terraform changes in this PR also work as expected, as `activity_dashboard.csv` has been generated in `napari-hub-dev/dev-pypi-data-cron/` after manually triggering the update via Eventbridge: https://us-west-2.console.aws.amazon.com/events/home?region=us-west-2#/eventbus/default/rules/dev-dev-pypi-data-cron-activity and providing the Snowflake credentials as environment variables in the `dev-pypi-data-cron-backend` lambda. Moreover, I have also verified that the file gets overwritten at the right schedule (cadence of time).

backend/api/model.py
```diff
@@ -374,6 +375,58 @@ def get_categories_mapping(version: str) -> Dict[str, List]:
...
+              ELSE 'other'
+          END AS download_type,
+          to_timestamp(timestamp) as timestamp
+      from imaging.pypi.downloads
+      where download_type = 'pip'
+      and project_type = 'plugin'
+      order by timestamp desc
+    )
+    group by 1,2
+    order by NUM_DOWNLOADS desc, MONTH, FILE_PROJECT""")
```
manasaV3, line 423: I did a little bit of poking around and found that of all the fields that we concatenate to form the details_all only `DETAILS_SYSTEM_RELEASE, DETAILS_IMPLEMENTATION_VERSION, DETAILS_DISTRO_VERSION, DETAILS_SYSTEM_NAME, DETAILS_OPENSSL_VERSION` have data that matches our check for ci at the moment. So based on the historic data we currently have, this query could be simplified to be:
```
SELECT 
    file_project, 
    date_trunc('month', timestamp) AS month, 
    COUNT(*) as num_downloads
FROM
    (
        SELECT 
            file_project,
            CONCAT(DETAILS_SYSTEM_RELEASE, DETAILS_IMPLEMENTATION_VERSION, DETAILS_DISTRO_VERSION, DETAILS_SYSTEM_NAME, DETAILS_OPENSSL_VERSION) AS details_all,
            to_date(timestamp) AS timestamp
        FROM 
            imaging.pypi.downloads
        WHERE 
            (details_all NOT LIKE '%aws%' AND details_all NOT LIKE '%-azure%' AND details_all NOT LIKE '%gcp%' AND details_all NOT LIKE '%amzn%')  
            AND details_installer_name = 'pip'
            AND project_type = 'plugin'
    )   
GROUP BY file_project, month
ORDER BY num_downloads DESC, month, file_project
```

@bnelson brought up a point, that it could be possible in the future that other fields could also contain ci related data, and also we could learn other patterns in the data that point to ci. So, it is more extensible to have a view supported by the data science team, that we could query on instead of directly querying the underlying data. 

The view would support an additional field that classifies the record as `ci | pip | ...`

klai95: Do you have any thoughts on this, @richaagarwal? I see it as a nice to have for the future iterations but not necessary something that needs to be captured right at this moment. Happy to discuss more though. We may also need to have more conversations and get more cllarity from @bnelson-czi to align on what this new change may look like.

bnelson-czi: I'm all for cleaning up and optimizing this query. At the same time, I'm leaning toward pushing this out to a future iteration so we can do a more thorough investigation.

.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -177,15 +177,35 @@ resource "aws_cloudwatch_event_target" "update_target" {
     }
 }
 
+# Cron job updating the activity data
+resource "aws_cloudwatch_event_rule" "activity_rule" {
+  name                = "${local.custom_stack_name}-activity"
```
klai95, line 182: Previously on AWS Eventbridge, it displayed `dev-dev-[branch_name`] due to this line of code. Since this is a light lift, the change has been implemented.

.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -163,7 +163,7 @@ module api_gateway_proxy_stage {
 
 # Cron job running the update endpoint
 resource "aws_cloudwatch_event_rule" "update_rule" {
-  name                = "${var.env}-${local.custom_stack_name}-update"
+  name                = "${local.custom_stack_name}-update"
```
klai95, line 166: Refer to https://github.com/chanzuckerberg/napari-hub/pull/710/files#r1006172752

.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -247,13 +267,13 @@ data aws_iam_policy_document plugins_policy {
 }
 
 resource aws_iam_role_policy policy {
-  name     = "${local.custom_stack_name}-${var.env}-policy"
+  name     = "${local.custom_stack_name}-policy"
```
klai95, line 270: Refer to https://github.com/chanzuckerberg/napari-hub/pull/710/files#r1006172752

.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -247,13 +267,13 @@ data aws_iam_policy_document plugins_policy {
...
   role     = module.backend_lambda.role_name
   policy   = data.aws_iam_policy_document.backend_policy.json
 }
 
 resource aws_iam_role_policy plugins_lambda_policy {
-  name     = "${local.custom_stack_name}-${var.env}-plugins-lambda-policy"
+  name     = "${local.custom_stack_name}-plugins-lambda-policy"
```
klai95, line 276: Refer to https://github.com/chanzuckerberg/napari-hub/pull/710/files#r1006172752

backend/api/model.py
manasaV3: @klai95 What are your thoughts on cleaning up the inner query to have fewer fields projected and optimize the `WHERE` clause?

@bnelson-czi  Created https://github.com/chanzuckerberg/imaging-tech-analytics/issues/18 to capture this. :)

manasaV3 (approved):

backend/api/model.py
richaagarwal: I'm all for simplifying the query if possible, but I'll defer to @bnelson-czi 's judgment on whether reducing the fields queried for determining the `ci usage` feels accurate to do so.  Agree that separating it into a view could be tackled by future work and not a high priority at this moment.

richaagarwal: @klai95 After discussing more with @manasaV3, let's hold on making updates to this query for the purposes of this PR, and we can tackle follow-up improvements in future work.

klai95: @richaagarwal Sounds good and makes sense to me!

richaagarwal (approved):
Looks great!

backend/api/app.py
```diff
@@ -109,6 +110,12 @@ def get_category(category: str, version: str) -> Response:
     return jsonify(get_category_mapping(category, get_categories_mapping(version)))
 
 
+@app.route('/activity/update', methods=['POST'])
```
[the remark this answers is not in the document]
richaagarwal: @codemonkey800 Good point! @manasaV3 actually opened up this issue here: https://github.com/chanzuckerberg/napari-hub/issues/719.  Our `/update` endpoint is currently unauthenticated as well, so though it's an existing issue, I agree it'll be good to get moving on it.  I think we can use the other issue to track work and doesn't need to be a blocker for this PR.

Issue #685, opened 2022-10-06
Filter plugin installs data


klai95:
For the `/activity/<plugin>` endpoint, we should filter installs data to the past _n_ months, where _n_ can be specified by the frontend, but otherwise defaults to 12. Additionally, the current month's data is not returned, so the data would always end at the previous month (e.g. in October 2022, we would return available data for October 2021 - September 2022).

### Additional notes
Currently on staging, the data is bunched together as the frontend is not set up to handle more than 12 data points:

<img width="1138" alt="Screen Shot 2022-10-06 at 4 17 03 PM" src="https://user-images.githubusercontent.com/70177777/194420263-30e8dabf-333c-4239-aed5-6e18cdca3724.png">

klai95:
@manasaV3 @codemonkey800 and me have just discussed the following points and we hope to get feedback from @richaagarwal about them:

Ensure frontend has more granular control and building an api that is more extensible; moreover, we are going to set a limit as a param to avoid malicious users from hitting our endpoints too many times, and also set up from and to parameters so users can choose to see which time frame of the chart they would be interested in looking at, and maybe by default if from and to are not specified, we just default to the past 12 months.

I can see how this work will help drive frontend-to-backend shared data schema efforts as well.

richaagarwal:
@klai95 Thanks for noting this, Kevin!  I've updated the issue's description above. Let me know if that captures this. In the future, we can certainly look into accommodating specific start and end dates as well. 

I also removed the bug label as this work is not yet released to production :)

klai95:
> @klai95 Thanks for noting this, Kevin! I've updated the issue's description above. Let me know if that captures this. In the future, we can certainly look into accommodating specific start and end dates as well.
> 
> I also removed the bug label as this work is not yet released to production :)

Just two clarifying questions: 

(1) Does this mean we are set on not displaying the current month's data? I know during the meeting some folks were talking about prediction models, so just want to get some clarity!

(2) When you say we should filter installs data to the past n months, is there an upper bound for what n could be? For example, if someone requests for a value of 100 for n, do we have a set limit (say arbitrarily 36 just for example) that cannot be exceeded to limit the amount of data that needs to be returned? 

@richaagarwal

richaagarwal:
@klai95 For our initial launch (what we've been referring to internally as P0 work), we won't be displaying the current month's data.  In follow-up work, there will likely be changes to what & how we display the data. 

[...]

PR #713, opened 2022-10-25
Filter plugin installs data

klai95:
Closes https://github.com/chanzuckerberg/napari-hub/issues/685

Development Branch: https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/dev-filter-install-data#/

Case 1: `Limit=0`
Screen Shot 2022-11-02 at 6 49 23 PM

Case 2: No Limit Parameter Specified: We Expect 12 Objects To Be Returned
Screen Shot 2022-11-02 at 6 50 38 PM

Case 3: `Limit=15`
Screen Shot 2022-11-02 at 6 51 20 PM

Case 4: `Limit=5`
Screen Shot 2022-11-02 at 6 51 52 PM

Case 5: Non-integer or negative integer values for `Limit`
Screen Shot 2022-11-02 at 7 02 46 PM
Screen Shot 2022-11-02 at 7 03 44 PM
Screen Shot 2022-11-02 at 7 04 05 PM


klai95:
Currently, preview/_tests/test_preview_meta.py::test_parse_preview_matches_hub is the only test that fails in Backend Tests / pytest (pull_request), which is a separate issue from this PR. For context, the reason for this failure is that the npe2 work has been released, but the current implementation does not have updated preview page code to handle npe2 metadata. This work will be captured in an upcoming PR by @DragaDoncila.

klai95:
Requesting for the `GET` endpoint 
​/activity​/natari on https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/dev-filter-install-data#/activity/get_activity__name_ returns a set of 12 data points, which is what we expect, since I have set the default value for `limit` to `12` in `app.py`.

backend/api/app.py
```diff
@@ -111,7 +111,10 @@ def get_category(category: str, version: str) -> Response:
 
 @app.route('/activity/<plugin>')
 def get_plugin_installs(plugin: str) -> Response:
-    return jsonify(get_installs(plugin))
+    limit = request.args.get('limit')
+    if not limit:
+        limit = '12'
```
manasaV3, line 116: Nit: We could do this as `limit = request.args.get('limit') if limit else '12'`

backend/api/model.py
```diff
@@ -374,20 +376,27 @@ def get_categories_mapping(version: str) -> Dict[str, List]:
...
+    :param limit: number of objects to return
     :return: list of objects
     """
     plugin_df = get_activity_dashboard_data(plugin)
-    plugin_df['MONTH'] = plugin_df['MONTH'].map(pd.Timestamp.timestamp) * 1000
+    plugin_df['MONTH_UTC'] = plugin_df['MONTH'].map(pd.Timestamp.timestamp) * 1000
+    last_day_of_prev_month = date.today().replace(day=1) - timedelta(days=1)
+    end_date = date.today().replace(day=1) - timedelta(days=last_day_of_prev_month.day)
+    start_date = end_date + relativedelta(months=-int(limit)+1)
+    end_date = end_date.strftime('%Y-%m-%d')
```
manasaV3, line 393: Nit: we could make `'%Y-%m-%d'` a constant.

backend/api/model.py
```diff
@@ -374,20 +376,27 @@ def get_categories_mapping(version: str) -> Dict[str, List]:
...
 
     :param plugin: plugin name
+    :param limit: number of objects to return
     :return: list of objects
     """
     plugin_df = get_activity_dashboard_data(plugin)
-    plugin_df['MONTH'] = plugin_df['MONTH'].map(pd.Timestamp.timestamp) * 1000
+    plugin_df['MONTH_UTC'] = plugin_df['MONTH'].map(pd.Timestamp.timestamp) * 1000
+    last_day_of_prev_month = date.today().replace(day=1) - timedelta(days=1)
+    end_date = date.today().replace(day=1) - timedelta(days=last_day_of_prev_month.day)
```
manasaV3, line 391: What are your thoughts on leveraging the `relativedelta(months=-1)` here as it has been done below? 
`end_date = date.today().replace(day=1) + relativedelta(months=-1)`?

backend/api/model.py
```diff
@@ -374,20 +376,27 @@ def get_categories_mapping(version: str) -> Dict[str, List]:
...
+    start_date = start_date.strftime('%Y-%m-%d')
+    plugin_df = plugin_df[(plugin_df['MONTH'] >= start_date) & (plugin_df['MONTH'] <= end_date)]
     installs_list = []
     for _, row in plugin_df.iterrows():
         obj = dict()
-        obj['x'] = int(row.MONTH)
+        obj['x'] = int(row.MONTH_UTC)
         obj['y'] = int(row.NUM_DOWNLOADS_BY_MONTH)
         installs_list.append(obj)
     return installs_list
```
manasaV3, line 402: To meet the criteria set [here](https://github.com/chanzuckerberg/napari-hub/issues/685#issuecomment-1270740467), should we add guard rails to ensure that months withouts any installs, have records with value set to 0?

backend/api/app.py
klai95: I get a syntax error on the second `limit1 variable when I do that, but `limit = request.args.get('limit') if request.args.get('limit') is not None else '12'` works. I am looking into a more elegant way to do it now.

backend/api/model.py
klai95: `obj['x'] = int(row.MONTH_UTC) if row.MONTH_UTC not none else 0`?

backend/api/model.py
klai95: I think that is a good idea. In that case, I no longer need line 391: `last_day_of_prev_month = date.today().replace(day=1) - timedelta(days=1)`

backend/api/app.py
manasaV3: Makes sense, I was accessing the variable before it got created. 😅

backend/api/model.py
manasaV3: I believe this is not checking for the guardrail criteria. This check would verify if the `MONTH_UTC` value in the data frame is not `None`.

backend/api/model.py
```diff
@@ -374,20 +376,27 @@ def get_categories_mapping(version: str) -> Dict[str, List]:
...
     and attribute y is the number of installs
 
     :param plugin: plugin name
+    :param limit: number of objects to return
     :return: list of objects
     """
+    date_format = '%Y-%m-%d'
     plugin_df = get_activity_dashboard_data(plugin)
-    plugin_df['MONTH'] = plugin_df['MONTH'].map(pd.Timestamp.timestamp) * 1000
+    plugin_df['MONTH_UTC'] = plugin_df['MONTH'].map(pd.Timestamp.timestamp) * 1000
```
richaagarwal, line 390: Minor, but for readability and cleanliness: 

Another way of doing this entire section would be to sort the dataframe by month (e.g. `plugin_df.sort_values(by='MONTH', inplace=True, ascending=False)`, and then slice the dataframe to the values you want, in this case it would be `plugin_df = plugin_df[1:(n+1)]` since we want the most recent n months, not including the current month.  This will work regardless of the day of month you're on.

backend/api/app.py
```diff
@@ -111,7 +111,8 @@ def get_category(category: str, version: str) -> Response:
 
 @app.route('/activity/<plugin>')
 def get_plugin_installs(plugin: str) -> Response:
-    return jsonify(get_installs(plugin))
+    limit = request.args.get('limit') if request.args.get('limit') is not None else '12'
```
richaagarwal, line 114: You should be able to do `request.args.get('limit', 12`) to set the default

backend/api/model.py
richaagarwal: Update to the above: Depending on the approach we go with for missing data, we may need to enumerate the months and set those accordingly, so this feedback may not be as relevant in that case. 

For enumerating the past _n_ months, you could do something like iterate over the range _n_ and then get the desired month that way using relativedelta.

backend/api/model.py
```diff
@@ -372,75 +373,37 @@ def get_categories_mapping(version: str) -> Dict[str, List]:
...
+    end_date = date.today().replace(day=1) + relativedelta(months=-1)
```
manasaV3, line 388: We could create a new variable, `start_of_current_month_date = date.today().replace(day=1)` and use that [here](https://github.com/chanzuckerberg/napari-hub/pull/713/files#diff-0ab08ad5c775cdda3530d5690e3f32dc4b670e91f86849f4d17260fda27196bcR390), as we are recomputing that below from the `end_date`.

backend/api/model.py
```diff
@@ -372,75 +373,37 @@ def get_categories_mapping(version: str) -> Dict[str, List]:
...
+    start_date = end_date + relativedelta(months=-int(limit)+1)
+    dates = list((pd.date_range(start_date, end_date + relativedelta(months=+1), freq='1M') - pd.offsets.MonthBegin(1)))
+    end_date = end_date.strftime(date_format)
+    start_date = start_date.strftime(date_format)
```
manasaV3, line 392: It is good practice to not reassign variables with different values, as it makes it harder to follow the code. 
In this case, we could assign this to a new variable, say `start_date_str` and `end_date_str`.

backend/api/model.py
```diff
@@ -372,75 +373,37 @@ def get_categories_mapping(version: str) -> Dict[str, List]:
...
+    :param limit: number of objects to return
     :return: list of objects
     """
+    date_format = '%Y-%m-%d'
     plugin_df = get_activity_dashboard_data(plugin)
-    plugin_df['MONTH'] = plugin_df['MONTH'].map(pd.Timestamp.timestamp) * 1000
+    plugin_df['MONTH_UTC'] = plugin_df['MONTH'].map(pd.Timestamp.timestamp) * 1000
+    end_date = date.today().replace(day=1) + relativedelta(months=-1)
+    start_date = end_date + relativedelta(months=-int(limit)+1)
+    dates = list((pd.date_range(start_date, end_date + relativedelta(months=+1), freq='1M') - pd.offsets.MonthBegin(1)))
```
manasaV3, line 390: We can simplify this to: `dates = pd.date_range(start=start_date, periods=limit, freq='MS')`

backend/api/model.py
```diff
@@ -372,75 +373,37 @@ def get_categories_mapping(version: str) -> Dict[str, List]:
...
         obj['y'] = int(row.NUM_DOWNLOADS_BY_MONTH)
         installs_list.append(obj)
-    return installs_list
+        dates.remove(row.MONTH)
+    for date_with_no_installs in dates:
+        obj = dict()
+        obj['x'] = int(date_with_no_installs.replace(tzinfo=timezone.utc).timestamp()) * 1000
+        obj['y'] = 0
+        installs_list.append(obj)
+    return sorted(installs_list, key=lambda d: d['x'])
```
manasaV3, line 406: What are your thoughts on using a map/dictionary to simplify this logic? 

We could create a map from the data in `plugin_df` such that `numberOfInstallsByMonth = {timestamp: numberOfInstalls}`. 

While iterating over the dates, we could get from the map, defaulting to zero, if there exists no entry for that date. This would also remove the need to sort the result. 

Untested code snippet:

```
installs_by_month = {}
for row in plugin_df.iterrows():
	installs_by_month[row.MONTH.strftime(date_format)] = row.NUM_DOWNLOADS_BY_MONTH

result = []
for date in pd.date_range(start=start_date, periods=limit, freq='MS'):
	result.append({'x': int(date.timestamp()), 'y':  installs_by_month.get(date.strftime(date_format))})

return result
```

klai95: Just to call out since I am in the process of testing, we do need `_` in `for _, row in plugin_df.iterrows():`, because not including it results in the following error: `AttributeError: 'tuple' object has no attribute 'NUM_DOWNLOADS_BY_MONTH'`

backend/api/model.py
klai95: I have changed `periods=limit` to `periods=int(limit)` to avoid the tests from breaking

backend/api/model.py
klai95: I have modified the code to `int(cur_date.timestamp()) * 1000`

backend/api/model.py
```diff
@@ -424,23 +426,34 @@ def update_activity_data():
...
     """
+    if not limit.isdigit():
+        return []
+    date_format = '%Y-%m-%d'
     plugin_df = get_activity_dashboard_data(plugin)
-    plugin_df['MONTH'] = plugin_df['MONTH'].map(pd.Timestamp.timestamp) * 1000
-    installs_list = []
+    plugin_df['MONTH_UTC'] = plugin_df['MONTH'].map(pd.Timestamp.timestamp) * 1000
+    end_date = date.today().replace(day=1) + relativedelta(months=-1)
+    start_date = end_date + relativedelta(months=-int(limit) + 1)
```
manasaV3, line 444: iirc did we want to do `-1 * (int(limit) + 1)` ?

backend/api/model.py
```diff
@@ -424,23 +426,34 @@ def update_activity_data():
...
+    dates = pd.date_range(start=start_date, periods=int(limit), freq='MS')
+    end_date = end_date.strftime(date_format)
+    start_date = start_date.strftime(date_format)
+    plugin_df = plugin_df[(plugin_df['MONTH'] >= start_date) & (plugin_df['MONTH'] <= end_date)]
```
manasaV3, line 448: With the new logic, do we still need this?

backend/api/model.py
```diff
@@ -424,23 +426,34 @@ def update_activity_data():
...
+    :param limit: number of objects to return
     :return: list of objects
     """
+    if not limit.isdigit():
+        return []
+    date_format = '%Y-%m-%d'
     plugin_df = get_activity_dashboard_data(plugin)
-    plugin_df['MONTH'] = plugin_df['MONTH'].map(pd.Timestamp.timestamp) * 1000
-    installs_list = []
+    plugin_df['MONTH_UTC'] = plugin_df['MONTH'].map(pd.Timestamp.timestamp) * 1000
```
manasaV3, line 442: I believe we are no longer using this column.

backend/api/model.py
```diff
@@ -424,23 +426,34 @@ def update_activity_data():
...
     and attribute y is the number of installs
-
     :param plugin: plugin name
+    :param limit: number of objects to return
     :return: list of objects
     """
+    if not limit.isdigit():
+        return []
+    date_format = '%Y-%m-%d'
     plugin_df = get_activity_dashboard_data(plugin)
```
richaagarwal, line 441: I would recommend filtering the dataframe returned in `get_activity_dashboard_data` by adding an optional `limit` parameter to that function as well. This way, you're only holding the necessary amount of data in memory, and it helps narrow what `get_installs` is responsible for.

richaagarwal: (Note that the logic for that will be a lot less complex as well, as you can just return a slice of the dataframe as I noted before)

backend/api/model.py
klai95: I just checked. That is correct, thanks!

klai95: `-1 * (int(limit) + 1)` returns

`DatetimeIndex(['2021-09-01', '2021-10-01', '2021-11-01', '2021-12-01',
               '2022-01-01', '2022-02-01', '2022-03-01', '2022-04-01',
               '2022-05-01', '2022-06-01', '2022-07-01', '2022-08-01'`
               
whereas 

`-int(limit) + 1` returns 
`DatetimeIndex(['2021-11-01', '2021-12-01', '2022-01-01', '2022-02-01',
               '2022-03-01', '2022-04-01', '2022-05-01', '2022-06-01',
               '2022-07-01', '2022-08-01', '2022-09-01', '2022-10-01'],
              dtype='datetime64[ns]', freq='MS')`

manasaV3 (approved):

backend/api/s3.py
```diff
@@ -80,5 +82,15 @@ def get_activity_dashboard_data(plugin) -> Dict:
...
+    plugin_df.sort_values(by='MONTH', inplace=True, ascending=False)
+    if limit is None:
+        return plugin_df
+    # It's possible for first row of dataframe to point to previous month's data, instead of current month's data.
+    # Therefore, `latest_month_installed` is used to determine how dataframe is sliced.
+    date_format = '%Y-%m-%d'
+    current_month = date.today().replace(day=1)
+    latest_month_installed = plugin_df.iloc[0].MONTH.strftime(date_format)
+    plugin_df = plugin_df.iloc[1:(int(limit)+1)] if latest_month_installed == current_month.strftime(date_format) else plugin_df.iloc[0:int(limit)]
+    for month_offset in range(len(plugin_df), int(limit)):
```
richaagarwal, line 94: What is this loop doing?

backend/api/model.py
```diff
@@ -424,23 +423,22 @@ def update_activity_data():
...
+        return []
+    plugin_df = get_activity_dashboard_data(plugin, limit)
+    result = []
     for _, row in plugin_df.iterrows():
-        obj = dict()
-        obj['x'] = int(row.MONTH)
-        obj['y'] = int(row.NUM_DOWNLOADS_BY_MONTH)
-        installs_list.append(obj)
-    return installs_list
+        result.append({'x': int(row.MONTH.timestamp()) * 1000, 'y': row.NUM_DOWNLOADS_BY_MONTH})
```
richaagarwal, line 440: I'm not sure what the loop below is doing (see this comment - https://github.com/chanzuckerberg/napari-hub/pull/713/files#r1016086766), but I think it would make sense to move that logic here. We should only have to loop over the dataframe once to do what we need.

backend/api/s3.py
richaagarwal: Update: It took me a while to realize this is iterating over any "gap" months, but I find the code a little hard to follow. I'm also not sure if it handles the case where the "gap" data is not sequential, e.g. if a plugin has 5 installs one month, zero installs the next month, and 10 installs the following month.

klai95: Yeah, that was my original thinking / implementation with creating the `dates` list. The goal would be to filter rows where it falls within the start_date and end_date, and then we can fill in default values for the "gap" months. Even if we sort the dataframe and get the next `limit` rows in the dataframe, it may be getting rows that are out of range, because of the issue with the "gap" months. Would it make sense to have the original logic here?

backend/api/model.py
klai95: This loop is returning the list of objects to return to the frontend in the way that Jeremy expects the data. See https://github.com/chanzuckerberg/napari-hub/issues/638#issuecomment-1225935611 for the exact specification. Also, at this point, `plugin_df` only has `limit` number of items in the list, and this is simply converting the data based on the frontend specification. 

I don't think we would move that logic to `s3.py`, because we would lose access to the dataframe. `get_installs_stats` needs only the dataframe and not this piece of logic.

backend/api/s3.py
richaagarwal: Yup, that makes sense -- sorry for leading you astray with the slicing! I think the original `dates` logic makes sense here. 

There should be a way to look up the specific row you want (by the date) without iterating over all the values in the dataframe, so that you don't have to create the dictionary as you do in this section - https://github.com/chanzuckerberg/napari-hub/blob/4764870a75fed1dcc425efe7c5eaaa9e5d2a41e0/backend/api/model.py#L442

But that's pretty minor, and other than that, I agree your original logic makes sense!

klai95: Sounds good, and no worries at all! I will look into that optimization you suggested as well as make the necessary changes!

backend/api/model.py
richaagarwal: Sorry, by "below" I meant the for loop in `s3.py`.  I agree we shouldn't move this logic there; I meant we could maybe move the logic for filling in the zeroes in `s3.py` up here. I think my original feedback on returning a slice of data in `s3.py` no longer applies, so it probably makes sense to keep all the business logic in `model.py`.

klai95: Oh okay got it! I will make that change. Thanks for the clarification!

backend/api/model.py
```diff
@@ -424,23 +426,31 @@ def update_activity_data():
...
-        obj['x'] = int(row.MONTH)
-        obj['y'] = int(row.NUM_DOWNLOADS_BY_MONTH)
-        installs_list.append(obj)
-    return installs_list
+    end_date = date.today().replace(day=1) + relativedelta(months=-1)
+    start_date = end_date + relativedelta(months=-int(limit) + 1)
+    dates = pd.date_range(start=start_date, periods=int(limit), freq='MS')
+    plugin_df = plugin_df[(plugin_df['MONTH'] >= start_date.strftime(date_format)) & (plugin_df['MONTH'] <= end_date.strftime(date_format))]
+    result = []
+    for cur_date in dates:
```
manasaV3, line 447: As we are only filling the result list for values in dates, I believe we don't need to iterate over plugin_df to filter to entries between start and end dates. 

Also, because we are retaining it as a data_frame after the filtering  in [here](https://github.com/chanzuckerberg/napari-hub/pull/713/files#diff-0ab08ad5c775cdda3530d5690e3f32dc4b670e91f86849f4d17260fda27196bcR449) we are iterating over the data every time to filter it. I believe filtering over a dataframe is a O(n) if we aren't leveraging a unique index. Making the complexity O(n<sup>2</sup>). 

If we mapped the data to a dictionary as we had done before, the computation of fetching a downloads for an install would be O(1). 

```
installsByMonth = {}
for _, rows in plugin_df[(plugin_df['MONTH'] >= start_date.strftime(date_format)) & (plugin_df['MONTH'] <= end_date.strftime(date_format))]:
    installsByMonth[rows.MONTH] = row.NUM_DOWNLOADS_BY_MONTH
```

manasaV3 (approved):

backend/api/model.py
klai95: At the point of `row = plugin_df[plugin_df['MONTH'] == cur_date]`, plugin_df is no longer of length n though, but rather (at default 12) that we will mostl likely set a hard limit for in the future. Even as n grows, where n is the number of rows of the dataframe we get from `get_activity_dashboard_data`, which essentially is the number of (month, year) rows for installs we get per plugin, wouldn't it still essentially be an O(1) operation as n scales? 

I can see that the filtering step is an O(n) operation, where n is how I have defined it above, but after the filtering, the length of the dataframe is more or less O(c), where c is a constant set at default of 12 and we can put a hard cap on the value to a reasonable constant.

What you suggest definitely makes sense, but I just want to get some clarity and context on what I've brought up before moving forward with a particular implementation!

manasaV3: @klai95 Thanks for breaking that down into more detail, I agree that for our limit of 12 this is not a major concern at the moment.  :)

I do want to preface this feedback is not a blocker for this PR to be merged, but more of consideration for further optimization. 

In my complexity computation, I had assumed the worst case where _limit = number of entries for that plugin_, and landed in the **O(n<sup>2</sup>)**. But, I agree that would not be the case for our current system. 

I think I am not following, how that time complexity becomes **O(1)** even when the limit is small as it changes with the limit and doesn't remain constant.  This is my understating of the current flow. Please let me know if there is something, I have misunderstood. 

Let's say we have an arbitrary limit `c`. 
Step 1: That would create a dataframe of at most `c` rows from the `n` rows of total data.  So, at the end of this operation, we would have a time complexity = **O(n)** where n = number of records in the data frame initially. This would result in a dataframe of max length `c`, as you had mentioned. 
Step 2: We are then iterating over `c` entries in the dates and filtering for each date on a dataframe with at most `c` rows. The filtering now becomes a **O(c)** operation.  Making the iterating block a **O(c<sup>2</sup>)** operation. 

And the overall complexity,  **O(n) + O(c<sup>2</sup>)**

While c might be small enough at the moment, we should bear in mind that, it is a non linear complexity as the c value increases. For example, if we wanted to use this function for a case where we wanted to visualize the installs by months in the last 5 years, the increase in time complexity will not be linear for it.

klai95: @manasaV3 To clarify, you are correct in saying that the overall complexity is O(n) + O(c^2). I said it would be O(n) + O(c) just because of my assumption that if the number of rows is between 12 and 24, we can effectively treat it just as O(c) since it is such a small number as n grows. I do agree with you that, in the future, it may be a non-negligible factor that affects the latency of the system if we were to ever query a few years worth of data, so I will be making the change again, thanks!!

PR #718, opened 2022-10-31
Secure backend jobs with API key

manasaV3:
Based on: https://github.com/chanzuckerberg/napari-hub/issues/719

Implementation difference:
We would need a shared secret between the caller (the event bridge) and the callee (the lambda) to secure the API calls. Instead of using the secret manager, a randomly generated UUID is used as a secret. The secret is known only to the required resources.

The advantage of this implementation is that lesser infrastructure needs to be maintained. It can also be set up to be updated at a regular cadence if the need arises.


richaagarwal (approved):

manasaV3:
Added relevant documentation: https://github.com/chanzuckerberg/napari-hub/wiki/Backend-Architecture#implementation-details-of-securing-data-update-workflow

Issue #720, opened 2022-10-31
Optimizing API endpoint for activity data


manasaV3:
### Summary
Currently, the frontend is making multiple calls to fetch data for loading several components of activity. This is not optimal, especially as all those data are related and we render them in the same place. This can be optimized for by having a single endpoint that returns all the data. 
### Intended Outcome
Proposed request path:

`/metrics/<plugin>?include="timeline","stats"&limit=25`

Proposed response structure:

```
{
    "activity": {
        "timeline": [
            {
                "timestamp": "10-1-2020",
                "installs": 2
            },
            {
                "timestamp": "11-1-2020",
                "installs": 0
            },
            {
                "timestamp": "12-1-2020",
                "installs": 12
            },
            ...
        ],
        "stats": {
            "totalInstalls": 500,
            "totalMonths": 15,
            "installInLast30Days": 100,
            ...
        }
    }
}
```

### How will it work?

Based on current data storage structures, we could fetch data from multiple different sources, filter the data by plugin and aggregate the response. This can be further optimized by updating how we store data.

richaagarwal:
I see this is in review, but does a PR for the work in [this comment](https://github.com/chanzuckerberg/napari-hub/pull/741/files#r1022229640) need to be created first?

manasaV3:
I believe it was set to review for https://github.com/chanzuckerberg/napari-hub/pull/741 

Now that https://github.com/chanzuckerberg/napari-hub/pull/749 has been merged, I can make a new PR to complete the cleanup. I can mark this issue closed after that. :)

Issue #709, opened 2022-10-24
API: return installs count for the prior 30 days


neuromusic:
Add new API endpoint to return installs count for the past 30 days (not including today).

### P0 implementation details
- Add an additional query to cron job (that will be introduced in https://github.com/chanzuckerberg/napari-hub/issues/664) to retrieve daily install counts (for past 30 days only) from Snowflake and write those to an additional file in the S3 folder.  
- Add a new API endpoint (similar to `/activity/<plugin>/stats`) that returns the installs count for the past 30 days.  This should just be a sum of what's in the S3 file since the filtering is happening as part of the query above. Align with Jeremy/front-end on expected variable naming for this.

### Dependencies
- https://github.com/chanzuckerberg/napari-hub/issues/664

manasaV3:
## Implementation strategies for last 30 days install metric

- Run query along with daily job to fetch data on installs by month
- Store data in persistent storage

### Fetching the data:

The daily job that runs to capture the number of installs by month can be leveraged to also update the installs in last 30 days metric as well. 

**Query:**

```
SELECT 
    file_project, 
    COUNT(*) as num_downloads
FROM
    (
        SELECT 
            file_project,
            CONCAT(
                DETAILS,
                DETAILS_INSTALLER, DETAILS_INSTALLER_NAME, DETAILS_INSTALLER_VERSION,
                DETAILS_PYTHON,
                DETAILS_IMPLEMENTATION_NAME, DETAILS_IMPLEMENTATION_VERSION,
                DETAILS_DISTRO, DETAILS_DISTRO_NAME, DETAILS_DISTRO_VERSION, DETAILS_DISTRO_ID, DETAILS_DISTRO_LIBC, DETAILS_DISTRO_LIBC_LIB, DETAILS_DISTRO_LIBC_VERSION,
                DETAILS_SYSTEM, DETAILS_SYSTEM_NAME, DETAILS_SYSTEM_RELEASE,
                DETAILS_CPU,
                DETAILS_OPENSSL_VERSION,
                DETAILS_SETUPTOOLS_VERSION,
                DETAILS_RUSTC_VERSION
            ) AS details_all,
            to_date(timestamp) AS timestamp
        FROM 
            imaging.pypi.downloads
        WHERE 
            (details_all NOT LIKE '%aws%' AND details_all NOT LIKE '%-azure%' AND details_all NOT LIKE '%gcp%' AND details_all NOT LIKE '%amzn%')  
[...]

PR #726, opened 2022-11-02
Implement API and data job for last 30 days of PyPI installs

manasaV3:
Relates to: https://github.com/chanzuckerberg/napari-hub/issues/709

#### Happy path - Valid plugin name: 
https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/dev-last30daysInstall/activity/affinder/recentStats

#### Invalid plugin name:
https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/dev-last30daysInstall/activity/test/recentStats

Relates to: 
https://github.com/chanzuckerberg/napari-hub/issues/720

#### Happy path - Valid plugin name: 
https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/dev-last30daysInstall/metrics/affinder

#### Invalid plugin name:
https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/dev-last30daysInstall/metrics/test


backend/api/model.py
```diff
@@ -92,6 +94,7 @@ def get_frontend_manifest_metadata(plugin, version):
     interpreted_metadata = parse_manifest(raw_metadata)
     return interpreted_metadata
 
+
```
klai95, line 97: Additional newline seems to have been inserted

backend/api/model.py
```diff
@@ -121,7 +125,7 @@ def get_manifest(plugin: str, version: str = None) -> dict:
     elif version is None:
         version = plugins[plugin]
     plugin_metadata = get_cache(f'cache/{plugin}/{version}-manifest.json')
-    
+
```
klai95, line 128: I think you may need to revert the change here too

backend/api/model.py
```diff
@@ -108,6 +111,7 @@ def discover_manifest(plugin: str, version: str = None):
         Payload=json.dumps(lambda_event),
     )
 
+
```
klai95, line 114: Additional newline here too

backend/api/model.py
```diff
@@ -377,52 +381,44 @@ def update_activity_data():
...
+        SELECT 
+            file_project, DATE_TRUNC('month', timestamp) as month, count(*) as num_downloads
+        FROM
+            imaging.pypi.labeled_downloads
+        WHERE 
+            download_type = 'pip'
+            AND project_type = 'plugin'
+        GROUP BY file_project, month
+        ORDER BY file_project, month
+        """
```
klai95, line 394: nit: should the bottom triple quote align with query? Just want to make sure

backend/api/model.py
```diff
@@ -377,52 +381,44 @@ def update_activity_data():
...
+            file_project, count(*) as num_downloads
+        FROM
+            imaging.pypi.labeled_downloads
+        WHERE 
+            download_type = 'pip'
+            AND project_type = 'plugin'
+            AND timestamp > DATEADD(DAY, -30, CURRENT_DATE)
+        GROUP BY file_project     
+        ORDER BY file_project
+        """
```
klai95, line 419: same question as above

backend/api/model.py
```diff
@@ -458,3 +454,41 @@ def get_installs_stats(plugin: str) -> Any:
...
     return obj
+
+def get_recent_installs_stats(plugin: str) -> Any:
+    """
+    This should return a dict, with number of Installs in the last 30 days
+
+    :param plugin: plugin name
+    :return: dict
+    """
+    response = {'installInLast30Days': get_recent_activity_dashboard_data().get(plugin, 0)}
```
richaagarwal, line 465: This is very minor, but I think the plural `installs` would make more sense in the naming here, so `installsInLast30Days`.

richaagarwal:
I'd suggest giving the pull request a more readable name, e.g. "Implement API endpoint for last 30 days of PyPI installs"

backend/api/model.py
manasaV3: I added the new line to be consistent with recommendations from [here](https://peps.python.org/pep-0008/#blank-lines) and also most of the other methods had two lines separating them.
Lemme know, if we are following a different formatting. 😅

backend/api/model.py
manasaV3: Yeah, there should be indent on the query.

backend/api/model.py
manasaV3: Yeah, there should be indent on the query.

backend/api/dao/s3.py
```diff
@@ -0,0 +1,53 @@
...
+import os
+from typing import Dict, Any
+
+import boto3
+from botocore.exceptions import ClientError
+
+from utils.timer import Timer
+
+
+class S3DAO:
```
richaagarwal, line 12: I don't know too much about the Data Access Object pattern in Java specifically, but I don't think DAO needs to be a part of the naming here.  Something more descriptive might be `HubS3` if we want to define our own set of functions to interface with boto. That said, I could imagine it would get a little redundant if we had to provide a wrapper method here each time we want to call a new boto utility. 

I think we can improve the functions available in https://github.com/chanzuckerberg/napari-hub/blob/main/backend/api/s3.py such that they are more reusable and enable easier logging, but am curious about the need for our own class.

backend/api/dao/s3.py
```diff
@@ -0,0 +1,53 @@
...
+    def __init__(self):
+        self.__bucket = os.environ.get('BUCKET')
+        self.__bucket_path = os.environ.get('BUCKET_PATH', '')
+        __endpoint_url = os.environ.get('BOTO_ENDPOINT_URL', None)
+        self.__s3_client = boto3.client('s3', endpoint_url=__endpoint_url)
+
+    def get_activity_timeline_data(self) -> Any:
+        try:
+            return self.__get_object_body_from_s3('activity_dashboard_data/plugin_installs.csv').decode('utf-8')
+        except Exception as e:
```
richaagarwal, line 23: It can be useful to catch particular exceptions we expect, and `raise` those we don't. I would caution against only logging here.

backend/api/dao/s3.py
```diff
@@ -0,0 +1,53 @@
...
+            logging.error(e)
+            return None
+
+    def get_recent_activity_dashboard_data(self) -> Dict:
+        path = 'activity_dashboard_data/plugin_recent_installs.json'
+        try:
```
richaagarwal, line 29: For error handling/logging, this could probably be captured in a decorator function, that you can then apply to the desired functions, to avoid redundant code. Lots of resources on decorators in python -- https://higee.io/exception-handling-decorators-with-proper-logging-d0f5b54f7ba8 has some examples specifically for exception handling.

backend/api/dao/s3.py
manasaV3: I wanted to have something that was following Single Responsibility Principle, where the only responsibility would be to interface with s3 and have no additional logic to it. 

It was easier to implement that for the activity pieces as the logic was easier to cleanly separate out. I wanted to clean up the other methods in api.s3 with more thought, so I left them as is in that module albeit the timer. :)

backend/api/dao/s3.py
manasaV3: I was planning on making the timer a decorator in the near future, but I am happy to do that here along with the logging. 😁

backend/api/s3.py
```diff
@@ -82,3 +82,11 @@ def get_activity_dashboard_data(plugin) -> Dict:
     plugin_df['MONTH'] = pd.to_datetime(plugin_df['MONTH'])
     return plugin_df
 
+
+def get_recent_activity_dashboard_data():
```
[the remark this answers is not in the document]
manasaV3: I was planning on retaining that logic in the model, to support the SRP. What are your thoughts on that?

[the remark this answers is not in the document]
richaagarwal: That sounds good to me!

backend/api/dao/s3.py
richaagarwal: Sounds good - doesn't have to be in this PR!

backend/api/dao/snowflake_db.py
```diff
@@ -0,0 +1,83 @@
...
+        SELECT 
+            file_project, DATE_TRUNC('month', timestamp) as month, count(*) as num_downloads
+        FROM
+            imaging.pypi.labeled_downloads
+        WHERE 
+            download_type = 'pip'
+            AND project_type = 'plugin'
+        GROUP BY file_project, month
+        ORDER BY file_project, month
+    """
```
klai95, line 20: Should there be a newline between the two queries?

backend/api/app.py
```diff
@@ -126,6 +127,16 @@ def get_plugin_installs_stats(plugin: str) -> Response:
     return jsonify(get_installs_stats(plugin))
 
 
+@app.route('/activity/<plugin>/recent_stats')
+def get_recent_plugin_installs_stats(plugin: str) -> Response:
+    return jsonify(get_recent_installs_stats(plugin))
+
+
+@app.route('/metrics/<plugin>')
```
klai95, line 135: Just to get more context, what are some examples of `<metrics>`?

manasaV3:
Closing this in favour of https://github.com/chanzuckerberg/napari-hub/pull/741

Issue #723, opened 2022-11-01
Render public release in the installs timerseries chart


codemonkey800:
The public release data point renders an additional dashed line and label to indicate the number of installs during the month it was released. The frontend prototype determines the public release based on the `null` data points:

https://github.com/chanzuckerberg/napari-hub/blob/90c7aa6d1396b82d9b1b9d0198494ef5d6d0cb51/frontend/src/components/AreaChart/AreaChart.tsx#L86-L89

```json
[
  { "x": 1627801200000, "y": null },
  { "x": 1630479600000, "y": null },
  { "x": 1633071600000, "y": null },
  { "x": 1635750000000, "y": null },
  { "x": 1638345600000, "y": 179 },
  { "x": 1641024000000, "y": 164 },
  { "x": 1643702400000, "y": 406 },
  { "x": 1646121600000, "y": 695 },
  { "x": 1648796400000, "y": 338 },
  { "x": 1651388400000, "y": 906 },
  { "x": 1654066800000, "y": 514 },
  { "x": 1656658800000, "y": 457 }
]
```

<img width="610" alt="image" src="https://user-images.githubusercontent.com/2176050/199335182-ae7e9b5f-c5b9-461d-b896-ae518cd73451.png">

however, this may behave weirdly if there are gaps in the data :

<img width="589" alt="image" src="https://user-images.githubusercontent.com/2176050/199335291-0ed4662e-204d-42da-9bbd-7bf76fdedb96.png">

another edge case is showing the hub start date at a later time:

![image](https://user-images.githubusercontent.com/2176050/199335678-1d9685d7-570f-45d7-bf00-a13e74992ce6.png)

---

Right now, the frontend makes the following assumptions about the data:

1. It's always 12 points, starting from the current month to the 12 months ago
2. The data points are sorted by the `x` coordinate, or the timestamp
3. There are no null data points after public release. This is used to indicate when to show the public release indicator

However, it seems like these assumptions aren't necessarily true, so we should try to figure out a better way to show the public release indicator

Issue #732, opened 2022-11-04
Activity dashboard empty states


codemonkey800:
Empty states need to be implemented for the activity dashboard and area chart components

<img width="726" alt="image" src="https://user-images.githubusercontent.com/2176050/200068777-79e40c64-f2e9-4437-b0f4-6868e76cd24c.png">

<img width="470" alt="image" src="https://user-images.githubusercontent.com/2176050/200068802-dd500dff-797f-4373-acb5-8620bc208827.png">

richaagarwal:
@codemonkey800 Hmm when I go to a new plugin on staging, e.g. https://staging.napari-hub.org/plugins/napari-psf-analysis, I don't see any activity section. Is this expected, or is there corresponding back-end work that needs to happen as a result of your changes?

richaagarwal:
@codemonkey800 Shouldn't this plugin https://staging.napari-hub.org/plugins/napari-svetlana have the "There is not yet enough data to display for this plugin" for the monthly installs section?

richaagarwal:
Ah I would suggest using the release date to calculate this, but yes it seems back-end and front-end should align here on what data needs to be passed. Let's discuss more on Monday!

manasaV3:
I agree that using the release date is ideal, but we don't store the release date in the activity data we currently cache and fetch on API calls. Utilizing the min on the timestamp of installs will not accurately represent the release date in cases where a plugin didn't have installs in the first few months after release. 
From an earlier discussion, I believe there was a consensus that exposing the release date instead of the total months in the stats would be valuable. This would alleviate the problem but might require a bit of rework in the back end. 

As the front end already has the release date, we could leverage that and add a check before processing to render the timeline. This could be a stop-gap before we implement the release date being returned as a part of the stats. 

This could also be implemented using the total months metric that gets returned right now. But @klai95 and I noticed a bug with this. We are currently using the minimum timestamp of installs to compute the total months, which would not be accurate in the edge case discussed above.

richaagarwal:
> I agree that using the release date is ideal, but we don't store the release date in the activity data we currently cache and fetch on API calls. Utilizing the min on the timestamp of installs will not accurately represent the release date in cases where a plugin didn't have installs in the first few months after release. From an earlier discussion, I believe there was a consensus that exposing the release date instead of the total months in the stats would be valuable. This would alleviate the problem but might require a bit of rework in the back end.
> 
[...]

PR #729, opened 2022-11-03
Recent installs section

codemonkey800:
## Description

Closes #708
Depends on #728

Implements the recent installs section for the activity dashboard. The 

## Demo

image

PR #733, opened 2022-11-04
Monthly installs section

### LINKED ISSUE #731: Update area chart component to match design
Current implementation is based on the prototype. It should be updated to match the new design:

image

codemonkey800:
## Description

Closes #731 
Depends on #729

Refactors the UI for the area chart to match the designs for the monthly installs section. This also partially fixes https://github.com/chanzuckerberg/napari-hub/issues/723 because this PR adds data normalization and filtering such that we only show data points after the published line.

## Demos

https://napari-hub-monthly-installs.vercel.app/plugins/napari-svg?enable-feature=activityDashboard

#### >= 12 months

image

#### < 12 months

image

#### 1 month

image

## Next Steps

- Implement https://github.com/chanzuckerberg/napari-hub/issues/732

### DIFF
--- a/frontend/src/components/ActivityDashboard/RecentInstalls.tsx
+++ b/frontend/src/components/ActivityDashboard/RecentInstalls.tsx
@@ -40,7 +40,7 @@ export function RecentInstalls() {
 
       <span className="mr-2">{t('activity:recentInstalls.inPast')}</span>
 
-      <div className="!font-medium">{t('activity:duration.30days')}</div>
+      <span className="!font-medium">{t('activity:duration.30days')}</span>
     </Text>
   );
 }


liaprins-czi:
Thanks for sharing this, it's looking so good to see it coming together!

There are just a few things to update that I saw from design perspective:

- [ ] The aspect ratio fo the chart is a little too tall. This would make there be too tall of a gap between the bottom of the other metrics (like number of installs) and any future metrics we'd add to the dashboard, below this top row of installs-based metrics + chart. Is it possible to maintain the same width of the chart, but make itshorter? Here is the aspect ratio of the design in Figma, if we can have it this same height that'd be ideal!
Screenshot 2022-11-07 at 9 13 15 AM

Current implementation of chart:
Screenshot 2022-11-07 at 9 13 31 AM


- [ ] When teh screen width is narrowed, the activity dashboard metrics + chart don't scale to fit the space as expected. I don't think I was very explicit about this in the designs, so I will mock up like I did for the [Collections breakpoints](https://www.figma.com/file/KZPeWr37HJFlIbVQbpSC71/Collections?node-id=107%3A448). But basically the metrics + chart should switch between 1-, 2-, 3-columns of content in the exact same way and at the same breakpoints as the [Collections design](https://www.figma.com/file/KZPeWr37HJFlIbVQbpSC71/Collections?node-id=107%3A448). Right now they are not widening to fill the space the same way. I will start mocking up these more thoroughly now and will let you know when ready (should be today).


- [ ] Yeah good point about the y-axis when the data only has zeroes! I think in that case we could just have the y-axis scale from 0 to 1, and could still have ~4-5 tick mark labels for consistency (like 0.2, 0.4, 0.6, 0.8, 1.0) if that is possible?

liaprins-czi:
UPDATE: @codemonkey800 Here are the breakpoint mockups for the Activity dashboard; I did the same breakpoints as for Collections, and mocked up what the screen should look like AT each breakpoint, and also right before (1px narrower) the breakpoint hits:
https://www.figma.com/file/d3t84J9Uzmj0GZnYROzdQX/Activity-Dashboard?node-id=1729%3A2340
(Scroll to the right or zoom out to see them all!)

liaprins-czi (changes requested):
(Just copy-pasting these original comments I'd made on this issue before I could add them as reviews)

Thanks for sharing this, it's looking so good to see it coming together!

There are just a few things to update that I saw from design perspective:

- [ ]  The aspect ratio fo the chart is a little too tall. This would make there be too tall of a gap between the bottom of the other metrics (like number of installs) and any future metrics we'd add to the dashboard, below this top row of installs-based metrics + chart. Is it possible to maintain the same width of the chart, but make itshorter? Here is the aspect ratio of the design in Figma, if we can have it this same height that'd be ideal (225px x 119px)!

Screenshot 2022-11-07 at 9 13 15 AM
Current implementation of chart (225px x 150px):
Screenshot 2022-11-07 at 9 13 31 AM

- [ ]  When teh screen width is narrowed, the activity dashboard metrics + chart don't scale to fit the space as expected. I don't think I was very explicit about this in the designs, so here's a link to much more specific design at each of the relevant breakpoints (follows the same pattern as Collections breakpoints). I mocked up what the screen should look like AT each breakpoint, and also right before (1px narrower) the breakpoint hits:
https://www.figma.com/file/d3t84J9Uzmj0GZnYROzdQX/Activity-Dashboard?node-id=1729%3A2340
(Scroll to the right or zoom out to see them all!)

- [ ]  Yeah good point about the y-axis when the data only has zeroes! I think in that case we could just have the y-axis scale from 0 to 1, and could still have ~4-5 tick mark labels for consistency (like 0.2, 0.4, 0.6, 0.8, 1.0) if that is possible?

manasaV3 (dismissed):

liaprins-czi (changes requested):
Thanks for fixing the area chart width so it scales with the columns as the page changes width. There are a few other things that shoudl change to better match the designs, but not sure if you were planning them as part of a separate PR, but I will list them here!
- [x] The other metrics should also scale the same way that the area chart responds to the screen width changing ([link to designs](https://www.figma.com/file/d3t84J9Uzmj0GZnYROzdQX/Activity-Dashboard?node-id=1729%3A5587&t=6SWUduTBBRShwLHC-4)). Currently they maintain their narrower width when the screen gets narrower and switches to having less columsn (3 to 2, then 2 to 1), but they should widen to match the new column width.
Current implementation when 2 columns:
Screenshot 2022-11-16 at 11 19 42 AM

Current implementation when 1 column:
Screenshot 2022-11-16 at 11 19 54 AM


- [ ] The tick labels text of the area chart is too large when the UI switches to 2 columns (<875px); they shoudl be the same size that they are at >=875px. Also the little dots within the area chart marking each data point shoudl be the same size (but if the dots are difficult / time-consuming to fix it's not a big deal!)
<875px:
Screenshot 2022-11-16 at 11 27 19 AM

>=875px:
Screenshot 2022-11-16 at 11 27 24 AM


- [ ] The first two metrics are rolling up the metrics into shortened forms like "1.1k", but we decided that we would list the actual whole number for these two install metrics, not shorten them with the "k" abbreviation style. However, I think it's good that the y-axis of the area chart _does_ use this shortened "1k" etc style, so it can fit larger numbers more concisely (so don't change it there!)


- [x] The months that are labeled in the x-axis should always be "Jan", "Apr", Jul", "Oct", and these month's labels should shift along the y-axis as needed to match where their months are. We want these months specifically to be labeled always because they map to the end of the four "quarters" of a year. (Sorry I didn't clarify this better in the final designs!)

liaprins-czi (approved):
Looks like everythign I requested is fixed! Thanks, nice work! 🙌 🐹

PR #735, opened 2022-11-04
Fix hover for anchor headings

### LINKED ISSUE #649: HTML anchors don't work anymore
#### Description
When you use a link with an HTML anchor, e.g. pointing users to installation instructions, the user is not sent to the anchor position.

If you look carefully, you will note that the anchor works for a split-second, but then the user is sent to the top of the page unfortunately.

#### Steps/Code to Reproduce

When clicking on this link:
https://www.napari-hub.org/plugins/devbio-napari#installation

#### Expected Results

You should see this:
image

#### Actual Results

But you see this:
image

codemonkey800:
## Description

Closes #649

Should fix the anchor headings hover effect

## Demo

fix-anchor-heading-hover

### DIFF
--- a/frontend/src/components/AnchorHeading/AnchorHeading.tsx
+++ b/frontend/src/components/AnchorHeading/AnchorHeading.tsx
@@ -20,7 +20,7 @@ export function AnchorHeading({
     element,
     {
       id,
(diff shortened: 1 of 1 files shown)


manasaV3 (approved):
LGTM! 👍

PR #741, opened 2022-11-08
Implement API and data job for last 30 days of PyPI installs

manasaV3:
Relates to: https://github.com/chanzuckerberg/napari-hub/issues/709

Happy path - Valid plugin name:
https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/dev-recent-stats/activity/affinder/recent_stats

Invalid plugin name:
https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/dev-recent-stats/activity/test/recent_stats

Relates to:
https://github.com/chanzuckerberg/napari-hub/issues/720

Happy path - Valid plugin name:
https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/dev-recent-stats/metrics/affinder

Invalid plugin name:
https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/dev-recent-stats/metrics/test

### DIFF
--- a/backend/api/app.py
+++ b/backend/api/app.py
@@ -133,7 +133,7 @@ def get_plugin_recent_installs_stats(plugin: str) -> Response:
 
 @app.route('/metrics/<plugin>')
 def get_plugin_metrics(plugin: str) -> Response:
-    return jsonify(get_metrics_for_plugin(plugin), request.args.get('limit', '12'))
+    return jsonify(get_metrics_for_plugin(plugin, request.args.get('limit', '12')))
 
 
 @app.route('/collections')
--- a/backend/api/app.py
+++ b/backend/api/app.py
@@ -133,7 +133,7 @@ def get_plugin_recent_installs_stats(plugin: str) -> Response:
 
 @app.route('/metrics/<plugin>')
 def get_plugin_metrics(plugin: str) -> Response:
-    return jsonify(get_metrics_for_plugin(plugin))
+    return jsonify(get_metrics_for_plugin(plugin), request.args.get('limit', '12'))
 
 
 @app.route('/collections')
--- a/backend/api/app.py
+++ b/backend/api/app.py
@@ -8,7 +8,7 @@ from flask_githubapp.core import GitHubApp
 from api.collections import get_collections, get_collection
 from api.model import get_public_plugins, get_index, get_plugin, get_excluded_plugins, update_cache, \
     move_artifact_to_s3, get_category_mapping, get_categories_mapping, get_manifest, get_installs, get_installs_stats, \
-    update_activity_data
+    update_activity_data, get_recent_installs_stats, get_metrics_for_plugin
 from api.shield import get_shield
 from utils.utils import send_alert, reformat_ssh_key_to_pem_bytes
 
@@ -124,6 +124,16 @@ def get_plugin_installs(plugin: str) -> Response:
 @app.route('/activity/<plugin>/stats')
 def get_plugin_installs_stats(plugin: str) -> Response:
     return jsonify(get_installs_stats(plugin))
+
+
+@app.route('/activity/<plugin>/recent_stats')
+def get_plugin_recent_installs_stats(plugin: str) -> Response:
+    return jsonify(get_recent_installs_stats(plugin))
+
+
+@app.route('/metrics/<plugin>')
+def get_plugin_metrics(plugin: str) -> Response:
+    return jsonify(get_metrics_for_plugin(plugin))
 
 
 @app.route('/collections')
--- a/backend/api/model.py
+++ b/backend/api/model.py
@@ -9,7 +9,7 @@ from collections import defaultdict
 import pandas as pd
 from utils.github import get_github_metadata, get_artifact
 from utils.pypi import query_pypi, get_plugin_pypi_metadata
-from api.s3 import get_cache, cache, get_activity_dashboard_data, write_activity_data
+from api.s3 import get_cache, cache, get_activity_dashboard_data, write_activity_data, get_recent_activity_data_from_s3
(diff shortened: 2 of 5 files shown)


manasaV3:
Closing this in favour of: https://github.com/chanzuckerberg/napari-hub/pull/741

backend/api/s3.py
```diff
@@ -63,22 +63,36 @@ def cache(content: Union[dict, list, IO[bytes]], key: str, mime: str = None):
...
 
 
-def write_activity_data(csv_string: str):
-    s3_client.put_object(Body=csv_string, Bucket=bucket, Key=os.path.join(bucket_path, "activity_dashboard_data/plugin_installs.csv"))
+def _get_complete_path(path):
+    return os.path.join(bucket_path, path)
 
 
-def get_activity_dashboard_data(plugin) -> Dict:
+def write_activity_data(data: str, path: str):
```
richaagarwal, line 70: Minor, but since this is now parametrizing the data and path, would it be more descriptive to name this something more generic?

backend/api/app.py
```diff
@@ -126,6 +126,16 @@ def get_plugin_installs_stats(plugin: str) -> Response:
     return jsonify(get_installs_stats(plugin))
 
 
+@app.route('/activity/<plugin>/recent_stats')
+def get_plugin_recent_installs_stats(plugin: str) -> Response:
+    return jsonify(get_recent_installs_stats(plugin))
+
+
+@app.route('/metrics/<plugin>')
```
richaagarwal, line 134: Curious about the naming here (metrics) vs. the naming above (activity)

backend/api/model.py
```diff
@@ -453,18 +452,76 @@ def get_installs(plugin: str, limit: str) -> List[Any]:
     return result
 
 
-def get_installs_stats(plugin: str) -> Any:
-    """
-    This should return an object, with numerical attributes totalInstallCount and totalMonths
+def _get_activity_dashboard_data(plugin):
```
richaagarwal, line 455: Is this meant to be a private method? If so, it should use a double underscore, and it should be the reverse - e.g. functionality is defined here and `get_activity_dashboard_data` is the one that is public and should be called.  That said, since this isn't a class at the moment, I'm curious what the purpose of a private/public method distinction would be.

richaagarwal: Update: I see that `get_activity_dashboard_data` is defined in another file and being imported here, so I don't think we should be redefining it here as a private or internal method, but I am curious what the goal is.

backend/api/app.py
manasaV3: I named it metrics, as it could be extended to support future metrics around citation, GitHub data, etc with the name holding true. 

I believe this will also require additional caching updates for api gateway because currently, we don't support that for this prefix.

backend/api/model.py
manasaV3: I figured we could mark methods in the module that are for internal purposes with the `_` to demarcate that they should not be imported. I based it on [this](https://peps.python.org/pep-0008/#method-names-and-instance-variables) 

> Use one leading underscore only for non-public methods and instance variables. 

@richaagarwal if the preference is to go with a double underscore, I can update the methods. :)

I do agree this specific private method should not be a wrapper around the method from `s3` module. Not sure what my thought process was when I created this. 😅

richaagarwal: Sounds good! Yes, the other methods are fine as a single underscore / indicating internal usage, was just confused about this one in particular :)

backend/api/model.py
```diff
@@ -461,10 +472,54 @@ def get_installs_stats(plugin: str) -> Any:
...
+            data[row[0]] = row[1]
+
+    write_data(json.dumps(data), "activity_dashboard_data/recent_installs.json")
+
+
+def _get_recent_activity_data(plugin) -> int:
+    return get_recent_activity_data_from_s3().get(plugin, 0)
+
+
+def get_recent_installs_stats(plugin: str) -> Dict:
```
richaagarwal, line 508: Is the idea for this function to go away once https://github.com/chanzuckerberg/napari-hub/issues/744 is complete?

backend/api/s3.py
```diff
@@ -63,22 +63,36 @@ def cache(content: Union[dict, list, IO[bytes]], key: str, mime: str = None):
...
+    s3_client.put_object(Body=data, Bucket=bucket, Key=_get_complete_path(path))
+
+
+def _get_from_s3(path):
+    return s3_client.get_object(Bucket=bucket, Key=_get_complete_path(path))['Body'].read().decode('utf-8')
+
+
+def get_activity_dashboard_data(plugin):
     """
     Get the content of activity_dashboard.csv file on s3.
```
richaagarwal, line 80: I know this was already inaccurate and missed in a prior code review, but let's update the docstring here so it's accurate (don't necessarily have to refer to the filename though).

backend/api/model.py
```diff
@@ -461,10 +472,54 @@ def get_installs_stats(plugin: str) -> Any:
...
+    cursor_list = _execute_query(query)
+    data = {}
+    for cursor in cursor_list:
+        for row in cursor:
+            data[row[0]] = row[1]
+
+    write_data(json.dumps(data), "activity_dashboard_data/recent_installs.json")
+
+
+def _get_recent_activity_data(plugin) -> int:
```
richaagarwal, line 504: Curious why this needs to be wrapped into an internal function.

backend/api/model.py
manasaV3: Yeah. :)

richaagarwal (approved):
Looks great! Thank you for cleaning up the queries

klai95 (commented):
LGTM!!

PR #749, opened 2022-11-15
Frontend use optimized activity api

### LINKED ISSUE #744: Use optimized API for activity dashboard
Once  #741 is merged, we should refactor the frontend to use the new API so we can reduce the number of API calls the frontend makes

The following will need to be updated:

- Mock server
- hub API client
- React hooks

codemonkey800:
## Description

Closes #744 

Refactors the activity dashboard to use the plugin metrics endpoint. The actual added LOC is 144, the generated fixtures data adds 1,793 new lines 😆

### DIFF
--- a/frontend/scripts/create-metrics-data.js
+++ b/frontend/scripts/create-metrics-data.js
@@ -24,14 +24,12 @@ for (const plugin of plugins) {
 
   const timeline = [];
   const now = dayjs();
-  for (let i = 1; i <= Math.min(totalMonths, 12); i += 1) {
+  for (let i = Math.min(totalMonths, 12); i >= 1; i -= 1) {
     timeline.push({
       timestamp: now.subtract(i, 'month').toString(),
       installs: Math.floor(installsInLast30Days * random(true)),
     });
   }
-
-  timeline.reverse();
 
   metrics[plugin] = {
     activity: {


frontend/scripts/create-metrics-data.js
```diff
@@ -0,0 +1,49 @@
...
+  const totalInstalls = installsInLast30Days * totalMonths;
+
+  const timeline = [];
+  const now = dayjs();
+  for (let i = 1; i <= Math.min(totalMonths, 12); i += 1) {
+    timeline.push({
+      timestamp: now.subtract(i, 'month').toString(),
+      installs: Math.floor(installsInLast30Days * random(true)),
+    });
+  }
```
manasaV3, line 32: What are your thoughts on making a non-first or non-last timeline record with zero installs and one of the random plugins not having any installs?

frontend/scripts/create-metrics-data.js
```diff
@@ -0,0 +1,49 @@
...
+
+  timeline.reverse();
+
+  metrics[plugin] = {
+    activity: {
+      timeline,
```
manasaV3, line 38: I didn't know the key could be automatically assigned to the object like this. Pretty cool. 😎

frontend/scripts/create-metrics-data.js
```diff
@@ -0,0 +1,49 @@
...
+  const timeline = [];
+  const now = dayjs();
+  for (let i = 1; i <= Math.min(totalMonths, 12); i += 1) {
+    timeline.push({
+      timestamp: now.subtract(i, 'month').toString(),
+      installs: Math.floor(installsInLast30Days * random(true)),
+    });
+  }
+
+  timeline.reverse();
```
manasaV3, line 34: Nit: We could instead reverse the for-loop to start at the goal and decrement it every iteration.

frontend/src/fixtures/metrics.json
```diff
@@ -0,0 +1,1793 @@
+{
+  "napari-test-plugin": {
+    "activity": {
+      "timeline": [
+        {
+          "timestamp": "Sun, 14 Nov 2021 23:24:34 GMT",
```
manasaV3, line 6: The backend returns the timestamp as an epoch timestamp, so should these also be of the same data type? 

PS Not sure if this matters, but the date is also set to the 1st of every month. :)

frontend/src/fixtures/metrics.json
```diff
@@ -0,0 +1,1793 @@
+{
```
manasaV3, line 1: Would this data be generated every iteration of the test?

manasaV3 (approved):

PR #753, opened 2022-11-15
Update plugin metadata copy to render in sentence case

### LINKED ISSUE #705: Updated copy per UXR recs
On sidebar of plugin page & plugin card

- "Release date" should become "Last updated"
- "First released" should remain the same

codemonkey800:
## Description

Closes #705

Refactors how we render the metadata on the plugin search result card so that we show the metadata labels in sentence case rather than lowercase 

## Demos

image


liaprins-czi (approved):
Looks good to me! Thanks for the quick update!

PR #757, opened 2022-11-15
Use split.io for frontend feature flagging

codemonkey800:
## Description

- Sets up the architecture for fetching feature flags from split.io for the hub frontend
  - This works by fetching split.io on the server and passing it as state to the frontend for server-side and client-side rendering
  - The flags need to be available for both phases of rendering to prevent UI flash on the frontend
- This also adds a new util `getServerSidePropsHandler` for  adding feature flags and i18n translations to an SSR page
- Refactors all SSR pages to use the above util

## Demo

In the demo I show off turning off the collections and activity dashboard feature in Split.io and show how quick it is to turn off a feature on a deployment

https://user-images.githubusercontent.com/2176050/207124234-a5b7da78-a7ce-46ea-a4e4-7bcbaa6b70ab.mp4


manasaV3 (approved):
lgtm :shipit:

PR #758, opened 2022-11-16
Fix preview i18n for category chips

codemonkey800:
## Description

Fixes issue with i18n not rendering correctly on preview page. Instead, it shows the i18n keys `showMore` and `collapse` instead of the rendered text 

### Before

image

image

### After

image

image


manasaV3 (approved):
:shipit:

PR #763, opened 2022-11-17
Fix hub spot regression

### LINKED ISSUE #761: [Subscription] Subscription button is not working even though subscriptions are going thru
#### Description
Applet not changing when subscription button is pressed

#### Steps/Code to Reproduce
1. Go to end of napari-hub page
2. Click on Subscription text field
3. Enter correct email data
4. Click on <Subscribe> button

#### Expected Results
Subscription message must be displayed

#### Actual Results
Seems button is not responding as message does not show up even though backend is working as subscription are being received

https://user-images.githubusercontent.com/63799148/202566210-0f7e4ec4-acc2-4301-bcf3-58138120f982.mov

codemonkey800:
Closes #761

There was a regression with the hub spot script where the form wasn't submitting. Not sure what the exact issue is with the script, but this fixes it for now by using the button instead of the form to submit

### DIFF
--- a/frontend/src/components/SignupForm/SignupForm.test.tsx
+++ b/frontend/src/components/SignupForm/SignupForm.test.tsx
@@ -15,7 +15,8 @@ describe('<SignupForm />', () => {
       <>
         <div id={FORM_CONTAINER_ID}>
           <form>
-            <input />
+            <input type="email" />
+            <input type="submit" />
           </form>
         </div>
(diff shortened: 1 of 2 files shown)


richaagarwal (dismissed):
Thanks for taking care of this so quickly!

manasaV3 (approved):

klai95 (approved):

PR #772, opened 2022-11-28
Fix monthly installs empty state

codemonkey800:
## Description

I did a bit of digging into the code and we should be able to fix https://github.com/chanzuckerberg/napari-hub/issues/732#issuecomment-1325616379 by checking if every data point is null

### DIFF
--- a/frontend/src/components/ActivityDashboard/MonthlyInstalls/MonthlyInstalls.tsx
+++ b/frontend/src/components/ActivityDashboard/MonthlyInstalls/MonthlyInstalls.tsx
@@ -41,7 +41,7 @@ export function MonthlyInstalls() {
     return <Skeleton height="100%" variant="rectangular" />;
   }
 
-  const isEmpty = chartData.length === 0;
+  const isEmpty = chartData.every((point) => point.y === null);
 
   return (
     <section>


richaagarwal (approved):

klai95 (approved):

frontend/src/components/ActivityDashboard/MonthlyInstalls/MonthlyInstalls.tsx
```diff
@@ -41,7 +41,7 @@ export function MonthlyInstalls() {
     return <Skeleton height="100%" variant="rectangular" />;
   }
 
-  const isEmpty = chartData.length === 0;
+  const isEmpty = chartData.every((point) => point.y === null);
```
manasaV3, line 44: Would this also check hold true for `point.y === 0`? :)

richaagarwal: Ah good point, the backend returns zeroes not null values.

manasaV3: That makes sense, thanks for clarifying that. 
I have been learning so many cool things about javascript, I was excited to know if it treated `0` and `null` similarly. :D

Issue #785, opened 2022-12-06
Activity dashboard does not capture the download of recOrder-napari plugin


chili-chiu:
#### Description

[recOrder-napari](https://www.napari-hub.org/plugins/recOrder-napari) reported 20k downloads on its description page using PePy, but 0 downloads are reported on the Activity page.

#### Steps/Code to Reproduce

Compare [recOrder-napari](https://www.napari-hub.org/plugins/recOrder-napari) Activity page and its [PePy page](https://pepy.tech/project/recOrder-napari).  

#### Expected Results

Expected the downloads captured by the Activity page are consistent with the PePy record.

#### Actual Results

The Activity page is not detecting the downloads.

PR #786, opened 2022-12-06
Handling case-sensitivity in plugin name for activity

manasaV3:
Relates to: https://github.com/chanzuckerberg/napari-hub/issues/785

### Summary
We transform the plugin name used for querying against data to lowercase to fix the issue. To ensure consistency, we also convert the plugin name to lowercase when we fetch and save the data.

### DIFF
--- a/backend/api/_tests/test_activity_dashboard.py
+++ b/backend/api/_tests/test_activity_dashboard.py
@@ -5,65 +5,65 @@ from api import model
 from datetime import datetime
 from dateutil.relativedelta import relativedelta
 
-base = datetime.today().date().replace(day=1)
-date_list = [base - relativedelta(months=x) for x in range(12) if x % 2 == 0]
-mock_installs = list(range(1, len(date_list) + 1))
-mock_df = pd.DataFrame({'MONTH': pd.to_datetime(date_list), 'NUM_DOWNLOADS_BY_MONTH': mock_installs})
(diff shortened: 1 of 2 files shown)


backend/api/_tests/test_activity_dashboard.py
```diff
@@ -5,65 +5,65 @@
 from datetime import datetime
 from dateutil.relativedelta import relativedelta
 
-base = datetime.today().date().replace(day=1)
-date_list = [base - relativedelta(months=x) for x in range(12) if x % 2 == 0]
-mock_installs = list(range(1, len(date_list) + 1))
-mock_df = pd.DataFrame({'MONTH': pd.to_datetime(date_list), 'NUM_DOWNLOADS_BY_MONTH': mock_installs})
-empty_df = pd.DataFrame(columns=['MONTH', 'NUM_DOWNLOADS_BY_MONTH'])
-mock_plugin_recent_installs = {'string-1': 25, 'foo': 10, 'bar': 30}
+BASE = datetime.today().date().replace(day=1)
```
klai95, line 8: Just for some context, what is the reason behind naming all variables in caps?

manasaV3: The python convention for marking a variable to be constant is to name it in uppercase. As all those variables should be treated as constants, I updated their name to reflect the convention. 

ref: https://peps.python.org/pep-0008/#constants

klai95 (approved):

richaagarwal (approved):

PR #795, opened 2022-12-16
Adding begins with checks to git url match

manasaV3:
Relates to: https://github.com/gitpython-developers/GitPython/issues/1515

According to https://security.snyk.io/vuln/SNYK-PYTHON-GITPYTHON-3113858:
This is only relevant when enabling the `ext` transport protocol.

We already validate the URLs that we pass on the `git.clone` to be `https` protocol with a regex match
https://github.com/chanzuckerberg/napari-hub/blob/main/backend/utils/github.py#L28
https://github.com/chanzuckerberg/napari-hub/blob/main/backend/preview/preview.py#L90

### Changes:
Making the regex match more explicit to validate that the url begins with `https://github.com`

### DIFF
--- a/backend/preview/preview.py
+++ b/backend/preview/preview.py
@@ -1,6 +1,5 @@
 from typing import Union
 from git import Repo
-from pkginfo import Wheel
 from collections import defaultdict
 import datetime
 import subprocess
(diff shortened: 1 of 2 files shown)


klai95 (commented):
LGTM!

PR #796, opened 2022-12-17
Resolve relative images when caching plugin data

### LINKED ISSUE #541: add support for relative links to images
### Who it's for

Users and plugin developers

### Problem it solves

Whether a plugin developer uses `.napari-hub/DESCRIPTION.md` or their PyPI description (which often copies `README.md`), Github supports relative links for images, but then they are not displayed on the napari hub.

### How it will work

When the plugin developer has provided relative links, we will attempt to resolve them if they have provided links to a valid Github repo.

codemonkey800:
## Description

Closes #541

This PR adds relative image resolution for plugin markdown descriptions. This works by following the image resolution rules outlined here: https://github.com/chanzuckerberg/napari-hub/issues/140#issuecomment-1084842222

This also handles the edge case regarding repo default branch by using the `default_branch` property from the GitHub repo API.

The only edge case that isn't handled is versioning, but since we currently do not support viewing versioned plugins on the hub, we can defer handling versioned images for later

### Relative images starting with `./`

https://www.napari-hub.org/plugins/napari-dzi-zarr
https://dev-rel-img-frontend.dev.imaging.cziscience.com/plugins/napari-dzi-zarr

### Relative images starting with `../`

https://www.napari-hub.org/plugins/napari-bleach-correct
https://dev-rel-img-frontend.dev.imaging.cziscience.com/plugins/napari-bleach-correct

### DIFF
--- a/backend/api/model.py
+++ b/backend/api/model.py
@@ -201,11 +201,14 @@ def build_plugin_metadata(plugin: str, version: str) -> Tuple[str, dict]:
     metadata = get_plugin_pypi_metadata(plugin, version=version)
     if not metadata:
         return plugin, metadata
-    github_repo_url = metadata.get('code_repository')
-    if github_repo_url:
-        metadata = {**metadata, **get_github_metadata(github_repo_url)}
-    if 'description' in metadata and github_repo_url:
-        repo = get_github_repo(github_repo_url)
+
+    code_repo_url = metadata.get('code_repository')
+    # TODO Add support for other repos like Bitbucket / GitLab
+    if code_repo_url and 'github.com' in code_repo_url:
+        metadata = {**metadata, **get_github_metadata(code_repo_url)}
+
+    if 'description' in metadata and code_repo_url:
+        repo = get_github_repo(code_repo_url)
         branch = get_github_default_branch(repo)
 
         if repo and branch:


klai95 (approved):
Looks good to me!

backend/utils/markdown.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def extendMarkdown(self, md: Markdown):
+        md.treeprocessors.register(
+            # Extension
+            ResolveRelativeImageTreeProcesser(md, self.repo, self.branch),
+            # Extension name
+            'resolve-image-extension',
+            # Priorty
+            0,
+        )
```
manasaV3, line 42: What are your thoughts on using parameter names here in place of comments? 
 ```
 md.treeprocessors.register(
            item=ResolveRelativeImageTreeProcesser(md, self.repo, self.branch),
            name='resolve-image-extension',
            priority=0,
        )
```

File: backend/utils/markdown.py [hunk 2]
```diff
@@ -0,0 +1,47 @@
...
+    def run(self, root):
+        for img in root.iter('img'):
+            src = img.get('src')
+
+            for pattern in ('./', '../'):
+                if src.startswith(pattern):
+                    url = get_raw_github_url(
+                        self.repo,
+                        self.branch,
+                        src.replace(pattern, ''),
```
manasaV3, line 24: Would replace also work for `../` as it would have to point to the parent folder?

backend/utils/markdown.py
```diff
@@ -0,0 +1,47 @@
+from markdown import Markdown
+from markdown.extensions import Extension
+from markdown.treeprocessors import Treeprocessor
+from markdownify import markdownify
+
+def get_raw_github_url(repo, branch, file):
+    return f"https://raw.githubusercontent.com/{repo}/{branch}/{file}"
```
manasaV3, line 7: Nit: As we are using single quotes for strings in all the other places in the file, could we also have this use single quotes?

backend/utils/markdown.py
manasaV3: It turns out we can leverage the `normpath` from the `os.path` [module](https://docs.python.org/3.9/library/os.path.html#os.path.normpath) for handling the above issue. I was thinking through this, and here is an untested snippet that may work for our case.

If we parsed the URL into its components, we could use the `base_path` to update our path.

```
from urllib.parse import urlparse, urlunparse
url = f'https://raw.githubusercontent.com/{repo}/{branch}/'
scheme, netloc, base_path, *_ = urlparse(url)
```

This way, we could retain the other components (if any) from the relative path. 
```
if src.startswith(pattern):
    *_, path, params, query, fragment = urlparse(src)
    complete_path = os.path.normpath(os.path.join(base_path, path))
    absolute_path = urlunparse(scheme=scheme, 
                               netloc=netloc, 
                               path=complete_path, 
                               params=params, 
                               query=query, 
                               fragment=fragment)
    img.set('src', absolute_path)
```

backend/utils/markdown.py
```diff
@@ -0,0 +1,47 @@
...
+
+class ResolveRelativeImageTreeProcesser(Treeprocessor):
+    def __init__(self, md, repo: str, branch: str):
+        super().__init__(md)
+        self.repo = repo
+        self.branch = branch
+
+    def run(self, root):
+        for img in root.iter('img'):
```
manasaV3, line 16: Should we also consider adding this for the anchor tags?

backend/utils/markdown.py
manasaV3: Totally agree that we should have an automated check for formatting... I am currently working on a pep8 lint checker that will run on PR requests...

PR #797, opened 2022-12-20
Updating cron frequency by environment

manasaV3:
Relates to: https://github.com/chanzuckerberg/napari-hub/issues/804

This updates the data refresh frequency to 1 hour for staging and one day for dev environments.

### DIFF
--- a/.happy/terraform/modules/ecs-stack/main.tf
+++ b/.happy/terraform/modules/ecs-stack/main.tf
@@ -54,7 +54,7 @@ locals {
   backend_function_name = "${local.custom_stack_name}-backend"
   plugins_function_name = "${local.custom_stack_name}-plugins"
 
-  plugin_update_schedule = var.env == "dev" ? "at(${formatdate("yyyy-mm-ddThh:mm:ss", timeadd(timestamp(), "15m"))})" : var.env == "staging" ? "rate(1 hours)" : "rate(5 minutes)"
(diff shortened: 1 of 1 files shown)


.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -53,6 +53,8 @@ locals {
   frontend_url = var.frontend_url != "" ? var.frontend_url: try(join("", ["https://", module.frontend_dns.dns_prefix, ".", local.external_dns]), var.frontend_url)
   backend_function_name = "${local.custom_stack_name}-backend"
   plugins_function_name = "${local.custom_stack_name}-plugins"
+
+  plugin_update_schedule = var.env == "dev" ? "rate(1 day)" : var.env == "staging" ? "rate(1 hours)" : "rate(5 minutes)"
```
richaagarwal, line 57: Very cool!  How are you envisioning dev data gets created upon a new dev branch? We wouldn't want to wait 24 hours.  In practice, there's not a real need for the cron jobs to be running at all in dev environments; but we do need data initialization within 5 minutes (or sooner).

manasaV3: I did a sanity check. As this uses `rate` it executes it immediately once the rule is created, and the subsequent trigger rate is defined by the period. So, we should be okay with this for our dev environment.

richaagarwal: Oh, perfect! Thanks for confirming.

richaagarwal (approved):

klai95 (approved):

PR #816, opened 2023-01-05
Adding behavioural tests for API

manasaV3:
Relates to https://github.com/chanzuckerberg/napari-hub/issues/818

#### Summary:
Introduces bdd tests for the GET API endpoints that fetch data from s3. It also adds a GitHub workflow to execute the test after a deployment to the dev environment. 

To allow the workflow to test the specific dev environment that was deployed to, the push remote dev workflow uploads an artifact of the prefix. The new workflow downloads the artifact and passes that to the behavior tests.


backend/features/steps/plugin_steps.py
```diff
@@ -0,0 +1,74 @@
+import json
+from behave import given, then
```
richaagarwal, line 2: I think you'll need to add behave to `backend/reqiurements.txt`

manasaV3: I hadn't added it there as it was only needed for testing and is not needed at run time. I included it in the setup [here](https://github.com/chanzuckerberg/napari-hub/pull/816/files#diff-f646c28b7a5bf1000289f659fc46254610c145f0feeaf0979788b75e8654e3a9R24) instead.

kne42: could you add it to a test requirements file then, in case local users want to test it?

manasaV3: Added it.. :)

.github/workflows/deploy.yml
```diff
@@ -71,3 +71,16 @@ jobs:
...
           env: "staging"
           operation: "create-or-update"
+    update-artifacts:
+      name: Setup artifacts on environment deployed to
+      runs-on: ubuntu-latest
+      steps:
+        - name: Set workflow artifact file
+          run: |
+            mkdir -p ./workflow_artifact
+            echo STACK_NAME > ./workflow_artifact/stack_name
```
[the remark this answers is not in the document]
manasaV3: Yes, Thanks for catching that.. :)

klai95 (dismissed):
Excited to see integration tests for the API! Having a more robust backend system that validates the response structure and the data is a huge plus!

klai95 (approved):

Issue #819, opened 2023-01-06
View latest commits made to plugin


klai95:
### Summary
- As a plugin user, when I'm viewing a plugin's activity, I want to be able to see when the latest commit occurred for the plugin (or its core package)

### Intended Outcomes
- The SQL query has been written to execute the requirement of this ticket.
- Backend has implemented code to run said SQL query on a daily basis to ensure data is updated.
- Backend has surfaced relevant endpoints for said data for frontend to display on the `Activity` tab.

### Note
- The SQL query and Snowflake output generated by SQL query are commented on this ticket.

### Links
- [Epic](https://github.com/chanzuckerberg/napari-hub/issues/807)
- [Design](https://www.figma.com/file/d3t84J9Uzmj0GZnYROzdQX/Activity-Dashboard?node-id=1912%3A4299)
- [Document](https://docs.google.com/document/d/1PG3hld19fdZSC7c57gBfg_MouOGcd-GC6GVj0EaRq-4/edit)
- [PR](https://github.com/chanzuckerberg/napari-hub/pull/832)

klai95:
SQL query for this ticket:
```
 select repo, date_trunc('month', to_date(commit_author_date)) as month, count(*) as num_commits
 from imaging.github.commits
 where repo_type = 'plugin'
 group by 1,2

klai95:
Snowflake output generated by SQL query:

<img width="1441" alt="Screen Shot 2023-01-06 at 2 55 44 PM" src="https://user-images.githubusercontent.com/70177777/211098467-c157d564-9616-44d5-8c9e-e839a094ff42.png">

klai95:
As verified by Ben in [this Slack conversation](https://czi-sci.slack.com/archives/C025V5TPDAL/p1672862449724019) with the Trust Team, there is no concern for posting SQL queries to the napari-hub repo.

EricMaxWang:
Note from sprint planning meeting on Mon, 1/9:

- consideration for shifting granularity of this 'latest commit' to be "within the last week" vs. "within the past day"
- update description w/ consideration of a lag in 24 hr data freshness


CONFIRMED:
- [Lia] "within the last week"

klai95:
PR for this ticket has been created: https://github.com/chanzuckerberg/napari-hub/pull/830

manasaV3:
To solve for plugin name differing from the repository name:

- Load cache/index.json into memory
- Create a dictionary with the key as code_repository and value plugin name
[...]

PR #832, opened 2023-01-13
View Plugin's Latest Commit 

klai95:
#closes https://github.com/chanzuckerberg/napari-hub/issues/819

### DIFF
--- a/backend/api/_tests/test_activity_dashboard.py
+++ b/backend/api/_tests/test_activity_dashboard.py
@@ -18,7 +18,7 @@ MOCK_PLUGIN_LATEST_COMMIT = '2023-01-01 00:00:00'
 
 class TestActivityDashboard(unittest.TestCase):
 
-    @patch.object(model, 'get_latest_commit', return_value=MOCK_PLUGIN_LATEST_COMMIT)
+    @patch.object(model, 'get_latest_commit', return_value=None)
     @patch.object(model, 'get_recent_activity_data', return_value={})
     @patch.object(model, 'get_install_timeline_data', return_value=EMPTY_DF.copy())
     def test_get_metrics_empty(self, mock_get_install_timeline_data, mock_get_recent_activity_data, mock_get_latest_commit):
--- a/backend/api/_tests/test_activity_dashboard.py
+++ b/backend/api/_tests/test_activity_dashboard.py
@@ -13,12 +13,12 @@ EMPTY_DF = pd.DataFrame(columns=['MONTH', 'NUM_DOWNLOADS_BY_MONTH'])
 PLUGIN_NAME = 'StrIng-1'
 PLUGIN_NAME_CLEAN = 'string-1'
 MOCK_PLUGIN_RECENT_INSTALLS = {PLUGIN_NAME_CLEAN: 25, 'foo': 10, 'bar': 30}
-MOCK_PLUGIN_LATEST_COMMIT = 1672531200
+MOCK_PLUGIN_LATEST_COMMIT = '2023-01-01 00:00:00'
 
 
 class TestActivityDashboard(unittest.TestCase):
 
-    @patch.object(model, 'get_latest_commit', return_value=None)
+    @patch.object(model, 'get_latest_commit', return_value=MOCK_PLUGIN_LATEST_COMMIT)
     @patch.object(model, 'get_recent_activity_data', return_value={})
     @patch.object(model, 'get_install_timeline_data', return_value=EMPTY_DF.copy())
     def test_get_metrics_empty(self, mock_get_install_timeline_data, mock_get_recent_activity_data, mock_get_latest_commit):
@@ -77,7 +77,7 @@ class TestActivityDashboard(unittest.TestCase):
         return [{timestamp_key: to_timestamp(i), installs_key: to_installs(i)} for i in range(start_range, 0)]
 
     @staticmethod
-    def _generate_expected_metrics(timeline=None, total_installs=0, installs_in_last_30_days=0, latest_commit=None):
+    def _generate_expected_metrics(timeline=None, total_installs=0, installs_in_last_30_days=0, latest_commit=MOCK_PLUGIN_LATEST_COMMIT):
         return {
             'usage': {
                 'timeline': timeline if timeline else [],
@@ -88,7 +88,7 @@ class TestActivityDashboard(unittest.TestCase):
             },
             'maintenance': {
                 'stats': {
-                    'latest_commit': latest_commit
+                    'latest_commit': datetime.strptime(latest_commit, "%Y-%m-%d %H:%M:%S").strftime("%s")
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


backend/api/s3.py
```diff
@@ -89,6 +89,14 @@ def get_install_timeline_data(plugin):
     return plugin_df
 
 
+def get_latest_commits() -> Dict:
+    try:
+        return json.loads(_get_from_s3("activity_dashboard_data/latest_commits.json"))
+    except Exception as e:
+        logging.error(e)
```
manasaV3, line 96: Good idea on logging. :)

backend/api/model.py
```diff
@@ -479,8 +503,12 @@ def get_metrics_for_plugin(plugin: str, limit: str) -> Dict:
...
     }
-    activity_data = {
+    latest_commits = get_latest_commits()
+    usage_data = {
         'timeline': timeline,
-        'stats': complete_stats
+        'stats': complete_stats,
+    }
+    maintenance_data = {
+        'latest commit': latest_commits[plugin],
```
manasaV3, line 512: Having a key without space is the best practice for API response.

backend/api/model.py
```diff
@@ -470,17 +471,54 @@ def _update_recent_activity_data(number_of_time_periods=30, time_granularity='DA
...
+            repo_to_plugin_dict[plugin_obj['code_repository'].replace('https://github.com/', '')] = plugin_obj['name']
+    # the latest commit is fetched as a tuple of the format (repo, timestamp)
+    cursor_list = _execute_query(query, "GITHUB")
+    data = {}
+    for cursor in cursor_list:
+        for row in cursor:
+            repo = row[0]
+            if repo in repo_to_plugin_dict:
+                plugin = repo_to_plugin_dict[repo]
+                data[plugin] = pd.to_datetime(row[1]).strftime(format="%Y-%m-%d %H:%M:%S")
```
manasaV3, line 500: Is there a reason we use pandas here instead of the default python datetime module?

backend/api/model.py
```diff
@@ -470,17 +471,54 @@ def _update_recent_activity_data(number_of_time_periods=30, time_granularity='DA
...
     timeline = [] if _is_not_valid_limit(limit) else _process_for_timeline(data, int(limit))
-    complete_stats = {
+
+    usage_stats = {
         'totalInstalls': install_stats.get('totalInstalls', 0),
         'installsInLast30Days': get_recent_activity_data().get(plugin, 0)
     }
-    activity_data = {
+    maintenance_stats = {
+        'latest_commit': get_latest_commit(plugin)
```
manasaV3, line 515: Not a blocker: From the principle of single responsibility, it is better to have the s3 module fetch and write objects to s3. We could have this module get the data for the plugin.

backend/api/_tests/test_activity_dashboard.py
```diff
@@ -13,50 +13,60 @@
...
     @patch.object(model, 'get_recent_activity_data', return_value=MOCK_PLUGIN_RECENT_INSTALLS)
     @patch.object(model, 'get_install_timeline_data', return_value=MOCK_DF.copy())
-    def test_get_metrics_nonempty(self, mock_get_install_timeline_data, mock_get_recent_activity_data):
+    def test_get_metrics_nonempty(self, mock_get_install_timeline_data, mock_get_recent_activity_data, mock_get_latest_commit):
         expected = self._generate_expected_metrics(
             timeline=self._generate_expected_timeline(-3),
             total_installs=sum(MOCK_INSTALLS),
-            installs_in_last_30_days=25
+            installs_in_last_30_days=25,
+            latest_commit="2023-01-01 00:00:00",
```
manasaV3, line 38: As `MOCK_PLUGIN_LATEST_COMMIT` is the exact value, we can use the constant here.

backend/api/model.py
klai95: When I use the default python datetime module, the format of the timestamp comes out to be something like `"2022-11-20+19:18:40", and I want the format to be consistent with what we already have

backend/api/model.py
klai95: Sorry, could you clarify what you mean? `get_latest_commit` is a method that is in s3.py used to fetch the data we need, and I am passing the plugin name in model.py

manasaV3: It is a good design principle to allow things to have a single functionality. For example, in this case, the s3 module should only be a way to write to s3 and get data from s3. It shouldn't have any knowledge about what the actual data is. It would just fetch the data from s3 and pass it on to this module, which would do the filtering, aka fetching the latest commit by a plugin name. 

I understand that many existing pieces of code do not follow that principle. So, I am not going to block on that. We could consider this in our future designs. :)

backend/api/model.py
manasaV3: Sorry, if I a missing something obvious here. Could we not use `strftime("%Y-%m-%d %H:%M:%S")`? 

The `strftime(..)` is also available for datetime.
https://docs.python.org/3/library/datetime.html#datetime.date.strftime

backend/api/model.py
klai95: That makes sense! I agree that we should do this in the future, since it is a good principle!

backend/api/model.py
```diff
@@ -470,17 +472,54 @@ def _update_recent_activity_data(number_of_time_periods=30, time_granularity='DA
...
+            repo_to_plugin_dict[plugin_obj['code_repository'].replace('https://github.com/', '')] = plugin_obj['name']
+    # the latest commit is fetched as a tuple of the format (repo, timestamp)
+    cursor_list = _execute_query(query, "GITHUB")
+    data = {}
+    for cursor in cursor_list:
+        for row in cursor:
+            repo = row[0]
+            if repo in repo_to_plugin_dict:
+                plugin = repo_to_plugin_dict[repo]
+                data[plugin] = pd.to_datetime(row[1]).strftime(format="%Y-%m-%d %H:%M:%S")
```
manasaV3, line 501: Why are we storing the data in the format of `"%Y-%m-%d %H:%M:%S"`? 
It is similar to the format that we are storing the `last_commit` field in the database. The only use case where we use the data must once again transform it at run time. 

It is better to transform it to an epoch timestamp here and store it. It will also make it easier to handle the case where the  `get_latest_commit(..)` returns `None`.

manasaV3 (changes requested):

manasaV3 (approved):

manasaV3:
This PR introduces both the data fetch and the API. The data job only runs once a day. So, before we push that to production, we should ensure the required files are available in S3 in the prod stage, so the API does not break.

backend/api/s3.py
```diff
@@ -89,6 +89,14 @@ def get_install_timeline_data(plugin):
     return plugin_df
 
 
+def get_latest_commit(plugin: str) -> Dict:
```
manasaV3, line 92: Missed this in my earlier review. This method never returns a dict.

klai95: Curious what is the ask here now that the pr is already merged in? I was just going to make this change in the current PRs I have.

backend/api/model.py
```diff
@@ -479,8 +503,12 @@ def get_metrics_for_plugin(plugin: str, limit: str) -> Dict:
...
+    usage_data = {
         'timeline': timeline,
-        'stats': complete_stats
+        'stats': complete_stats,
+    }
+    maintenance_data = {
+        'latest commit': latest_commits[plugin],
     }
-    return {'activity': activity_data}
+    return {'usage': usage_data, 'maintenance': maintenance_data}
```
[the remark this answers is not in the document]
richaagarwal: Thanks for flagging this @codemonkey800, and sorry I missed this comment earlier. 

Just posting now for posterity: I know we'll be making corresponding frontend changes soon and before the next release. However, in the future, we should make the necessary frontend changes as part of the PR, or have a set of back-end and front-end PRs that get merged in together, so that we can avoid breaking functionality on staging.

PR #836, opened 2023-01-18
Latest commit UI

### LINKED ISSUE #825: Latest commit section
image

codemonkey800:
## Description

Closes #825

Implements latest commit UI section and refactors code to be reusable across the Activity and Maintenance section 

## Demo

image

### DIFF
--- a/frontend/i18n/en/activity.json
+++ b/frontend/i18n/en/activity.json
@@ -26,6 +26,9 @@
   "totalCommits": {
     "repoCreated": "since plugin repo created"
   },
+  "recentCommit": {
+    "latest": "Latest commit made"
+  },
   "duration": {
     "30days": "30 days",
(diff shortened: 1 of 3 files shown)


frontend/i18n/en/activity.json
```diff
@@ -26,6 +26,9 @@
   "totalCommits": {
     "repoCreated": "since plugin repo created"
   },
+  "recentCommit": {
```
klai95, line 29: Is the reason it's called `recentCommit` because we may have commits from different interval of time, instead of just the one that is the latest?

klai95: That makes sense! We can keep it as is, just want to make sure!

manasaV3 (dismissed):

klai95 (approved):

Issue #820, opened 2023-01-06
View commit activity over the past year


klai95:
### Summary
- As a plugin user, when I'm viewing a plugin's activity, I want to be able to see how commit activity has varied over the past year.

### Intended Outcomes
- The SQL query has been written to execute the requirement of this ticket.
- Backend has implemented code to run said SQL query on a daily basis to ensure data is updated.
- Backend has surfaced relevant endpoints for said data for frontend to display on the `Activity` tab.

### Note
- The SQL query and Snowflake output generated by SQL query are commented on this ticket.

### Links
- [Epic](https://github.com/chanzuckerberg/napari-hub/issues/807)
- [Design](https://www.figma.com/file/d3t84J9Uzmj0GZnYROzdQX/Activity-Dashboard?node-id=1912%3A4299)
- [Document](https://docs.google.com/document/d/1PG3hld19fdZSC7c57gBfg_MouOGcd-GC6GVj0EaRq-4/edit)

klai95:
SQL query for this ticket:
``` 
select repo, date_trunc('month', to_date(commit_author_date)) as month, count(*) as num_commits
 from imaging.github.commits
 where repo_type = 'plugin'
 group by 1,2

klai95:
Snowflake output generated by SQL query:

<img width="1440" alt="Screen Shot 2023-01-06 at 3 03 15 PM" src="https://user-images.githubusercontent.com/70177777/211099594-829f4c21-b3f9-4346-be59-daae676a4fd0.png">

klai95:
As verified by Ben in [this Slack conversation](https://czi-sci.slack.com/archives/C025V5TPDAL/p1672862449724019) with the Trust Team, there is no concern for posting SQL queries to the napari-hub repo.

Issue #817, opened 2023-01-05
View total number of commits made to plugin since repo's creation


klai95:
### Summary
- As a plugin user, when I'm viewing a plugin's activity, I want to be able to see the total number of commits made to the plugin (and its core package) since the repo was created.

### Intended Outcomes
- The SQL query has been written to execute the requirement of this ticket.
- Backend has implemented code to run said SQL query on a daily basis to ensure data is updated.
- Backend has surfaced relevant endpoints for said data for frontend to display on the `Activity` tab.

### Note
- This ticket can be addressed by summing all values of `num_commits` for each repo to yield the total number of commits made to the plugin (and its core package) since each repo has been created.
- The SQL query and Snowflake output generated by SQL query are commented on this ticket.

### Links
- [Epic](https://github.com/chanzuckerberg/napari-hub/issues/807)
- [Design](https://www.figma.com/file/d3t84J9Uzmj0GZnYROzdQX/Activity-Dashboard?node-id=1912%3A4299)
- [Document](https://docs.google.com/document/d/1PG3hld19fdZSC7c57gBfg_MouOGcd-GC6GVj0EaRq-4/edit)

klai95:
select REPO, SUM(NUM_COMMITS) as TOTAL_COMMITS
from
(
 select repo, date_trunc('month', to_date(commit_author_date)) as month, count(*) as num_commits
 from imaging.github.commits
 where repo_type = 'plugin'
 group by 1,2
)
group by Repo 
order by TOTAL_COMMITS desc

klai95:
Snowflake output from SQL query:
<img width="1443" alt="Screen Shot 2023-01-05 at 8 34 32 PM" src="https://user-images.githubusercontent.com/70177777/210918251-f474d0b9-59a4-417e-bf56-59194e11dcc4.png">

klai95:
As verified by Ben in [this Slack conversation](https://czi-sci.slack.com/archives/C025V5TPDAL/p1672862449724019) with the Trust Team, there is no concern for posting SQL queries to the napari-hub repo.

PR #848, opened 2023-01-27
View Plugin's Total Commit

klai95:


klai95:
API contract
```
{
  "maintenance": {
    "stats": {
      "latest_commit_timestamp": 1674591907000,
      "total_commits": 286
    }
  },
  "usage": {
    "stats": {
      "installs_in_last_30_days": 7,
      "total_installs": 70
    },
    "timeline": [
      {
        "installs": 0,
        "timestamp": 1640995200000
      },
      {
        "installs": 0,
        "timestamp": 1643673600000
      },
      {
        "installs": 0,
        "timestamp": 1646092800000
      },
      {
        "installs": 0,
        "timestamp": 1648771200000
      },
      {
        "installs": 0,
        "timestamp": 1651363200000
      },
      {
        "installs": 20,
        "timestamp": 1654041600000
      },
      {
        "installs": 6,
        "timestamp": 1656633600000
      },
      {
        "installs": 2,
        "timestamp": 1659312000000
      },
      {
        "installs": 4,
        "timestamp": 1661990400000
      },
      {
        "installs": 26,
        "timestamp": 1664582400000
      },
      {
        "installs": 4,
        "timestamp": 1667260800000
      },
      {
        "installs": 1,
        "timestamp": 1669852800000
      }
    ]
  }
}

**Comment by klai95** [comment]
Moved https://github.com/chanzuckerberg/napari-hub/issues/817 to `In Code Review`

**Comment by manasaV3** [comment]
This PR introduces both the data fetch and the API. The data job only runs once a day. So, before we push that to production, we should ensure the required files are available in S3 in the prod stage, so the API does not break.

**Comment by manasaV3** [comment]
This PR introduces both the data fetch and the API. The data job only runs once a day. So, before we push that to production, we should ensure the required files are available in S3 in the prod stage so the API does not break.

File: backend/api/model.py [hunk 1]
```diff
@@ -501,6 +507,33 @@ def _update_latest_commits():
...
+            (
+                SELECT repo, date_trunc('month', to_date(commit_author_date)) as month, count(*) as num_commits
+                FROM imaging.github.commits
+                WHERE repo_type = 'plugin'
+                GROUP BY 1,2
+            )
+        GROUP BY repo
+        ORDER BY total_commits desc    
+    """
+    repo_to_plugin_dict = _get_repo_to_plugin_dict()
```
manasaV3, line 525: Instead of fetching and processing the data twice, could we create the dict it in the parent method once and pass it to all the methods that need it?

backend/api/model.py
```diff
@@ -501,6 +507,33 @@ def _update_latest_commits():
...
+    cursor_list = _execute_query(query, "GITHUB")
+    data = {}
+    for cursor in cursor_list:
+        for row in cursor:
+            repo = row[0]
+            if repo in repo_to_plugin_dict:
+                plugin = repo_to_plugin_dict[repo]
+                data[plugin] = int(row[1])
+    write_data(json.dumps(data), "activity_dashboard_data/total_commits.json")
```
manasaV3, line 534: Do we need this file or can we do a summation on the monthly timeline?

backend/api/model.py
```diff
@@ -512,7 +545,8 @@ def get_metrics_for_plugin(plugin: str, limit: str) -> Dict:
         'installs_in_last_30_days': get_recent_activity_data().get(plugin, 0)
     }
     maintenance_stats = {
-        'latest_commit_timestamp': get_latest_commit(plugin)
+        'latest_commit_timestamp': get_latest_commit(plugin),
+        'total_commits': get_total_commit(plugin)
```
manasaV3, line 549: Adding a comma to the entry is a good practice. It will allow us to add newer entries without changing other lines.

backend/api/s3.py
```diff
@@ -97,10 +97,17 @@ def get_latest_commit(plugin: str) -> Dict:
         return None
 
 
+def get_total_commit(plugin: str) -> Dict:
```
manasaV3, line 100: I will leave this here in case we wanted to implement this: https://github.com/chanzuckerberg/napari-hub/pull/832#discussion_r1087080363

backend/api/model.py
klai95: We need this file. The monthly timeline only has data for the past year

klai95:
@codemonkey800 Commenting here to align on api contract on frontend with respect to the changes in this PR!

backend/api/model.py
```diff
@@ -501,6 +507,33 @@ def _update_latest_commits():
...
+    query = f"""
+        SELECT repo, sum(num_commits) as total_commits from
+            (
+                SELECT repo, date_trunc('month', to_date(commit_author_date)) as month, count(*) as num_commits
+                FROM imaging.github.commits
+                WHERE repo_type = 'plugin'
+                GROUP BY 1,2
+            )
+        GROUP BY repo
+        ORDER BY total_commits desc
```
manasaV3, line 523: Could we simplify this query?

backend/api/model.py
manasaV3: I did not realize that. But, if we wanted to support a limit component for that as well, having just a year's data might not be sufficient.

klai95: Could you clarify what you mean by your comment above? The current implementation of this PR maps each plugin to its total commits since the inception of the repo for all plugins in `total_commits.json`, which is the ask of https://github.com/chanzuckerberg/napari-hub/issues/817. 

If you are talking about the changes necessary in https://github.com/chanzuckerberg/napari-hub/issues/820 instead from the other PR, from the requirement of the ticket, it wouldn't need more than a year's worth of data, since we are viewing commit activity for the past year, and therefore we shouldn't need to support a limit component as of right now. If we need to do so based on the needs of some future tickets, we can always make that change!

backend/api/model.py
klai95: From my understanding, the only other way we can execute this query while maintaining the correct functionality is to use the `join` clause instead of nesting as the current one does, since one runs into a nesting error when trying to perform something like `sum(count(*))`. Curious what your thoughts are, as using the join clause isn't exactly simplifying the query.

backend/api/s3.py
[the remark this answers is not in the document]
klai95: Good catch; will change the return type to `Any`

backend/api/model.py
klai95: To be specific, when you say `the parent method`, are you referring to `_get_repo_to_plugin_dict`?

klai95:
Currently, there is an error in `Backend Tests` with respect to `preview/_tests/test_preview_meta.py:84: AssertionError`

I have noted the concern in the `napari-hub-eng` Slack channel, but I would like to surface the information here as well to clarify that the failed test has nothing to do with the code changes in this PR.

backend/api/model.py
manasaV3: What if we did:
```
SELECT repo, sum(1) as total_commits
FROM imaging.github.commits
WHERE repo_type = 'plugin'
GROUP BY 1
ORDER BY total_commits desc  
```

backend/api/model.py
manasaV3: No, the parent method I am referencing is whichever is calling `_update_total_commits()`. In this case, it would be `update_activity_data`.

backend/api/model.py
manasaV3: We can always rewrite code for future requirements. But, it is good practice to design things in a way that requires minimal rework for the future. 

Also, this doesn't maintain parity with the existing activity piece with a similar structure. In total_installs computation, we do a summation of the installs we retrieved for a plugin. 

While I will not block on this, it is worth considering.

backend/api/s3.py
```diff
@@ -89,18 +89,25 @@ def get_install_timeline_data(plugin):
...
-def get_latest_commit(plugin: str) -> Dict:
+def get_latest_commit(plugin: str) -> Any:
     try:
         return json.loads(_get_from_s3("activity_dashboard_data/latest_commits.json"))[plugin]
     except Exception as e:
         logging.error(e)
         return None
 
 
+def get_total_commit(plugin: str) -> Any:
```
manasaV3, line 100: We are repeating the same boilerplate in multiple places. This can be reduced to a helper function that handles failure on fetching from s3 and parsing into json.

backend/api/s3.py
```diff
@@ -89,18 +89,21 @@ def get_install_timeline_data(plugin):
...
+        return {}
 
 
-def get_recent_activity_data() -> Dict:
-    try:
-        return json.loads(_get_from_s3("activity_dashboard_data/recent_installs.json"))
-    except Exception as e:
-        logging.error(e)
-        return {}
+def get_latest_commit(plugin: str) -> Dict:
```
manasaV3, line 100: I believe the return type for this was also supposed to be `Any`.

manasaV3 (approved):
:shipit:

backend/api/model.py
richaagarwal: @klai95 I know this work has already been merged, but I agree with Manasa's feedback here and is worthwhile to open up an additional PR to address this before our next release. I think this line in a separate PR - https://github.com/chanzuckerberg/napari-hub/pull/849/files#diff-0ab08ad5c775cdda3530d5690e3f32dc4b670e91f86849f4d17260fda27196bcR547 - is the origin of this, since limiting it to 12 months of data then necessitates needing to keep track of a separate file in this PR.  So, it is good to think through these set of changes holistically and minimize duplication where possible.

PR #871, opened 2023-01-30
Handling prefix mount in dev environment

manasaV3:
Closes: https://github.com/chanzuckerberg/napari-hub/issues/872

With the custom domain of the API gateway's dev environment, we get the prefix of the stack as a part of the URL. This PR adds wsgi middleware to allow the route mapping to ignore the prefix if present. 

We can now access the dev api as follows: 
- https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/dev-manasa-test/
- https://api.dev.napari-hub.org/dev-manasa-test/

Relates to: https://github.com/chanzuckerberg/sci-imaging-infra/pull/77
Blocks: https://github.com/chanzuckerberg/napari-hub/pull/816

### DIFF
--- a/backend/api/custom_wsgi.py
+++ b/backend/api/custom_wsgi.py
@@ -1,0 +1,11 @@
+
+def script_path_middleware(script_path):
+    def wrapper(wsgi_app):
+        def new_app(environ, start_response):
+            if environ['PATH_INFO'].startswith(script_path):
+                environ['PATH_INFO'] = environ['PATH_INFO'][len(script_path):]
+                environ['SCRIPT_NAME'] += script_path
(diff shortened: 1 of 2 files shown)


backend/api/custom_wsgi.py
```diff
@@ -0,0 +1,11 @@
+
+def script_path_middleware(script_path):
```
klai95, line 2: I think a few comments describing the purpose of the method would help a lot! I know it is clear from the description of the PR what the code achieves, but having it directly in the codebase helps clarify it even more! :)

klai95 (approved):

PR #873, opened 2023-01-31
Fix code_repository key access error

klai95:
Fixes the internal server error caused by directly accessing a key that may not exist


manasaV3 (approved):
:shipit:

Issue #838, opened 2023-01-18
Integrate backend for maintenance section of activity dashboard


codemonkey800:
When #824, #825, and #826 are closed, we should be ready to start integration the backend with the frontend as soon as the backend is ready.

Right now the UI uses static data to render everything, so we'll have to replace all of the static data with hooks that make API calls to the backend.

klai95:
The response for activity dashboard endpoint will be structured in the following way:
`{'usage': usage_data, 'maintenance': maintenance_data}`

PR #883, opened 2023-02-03
Activity maintenance frontend integration

codemonkey800:
## Description

https://github.com/chanzuckerberg/napari-hub/issues/838

Replaces fixture data with actual maintenance data from metrics hook.

## Demos

image

### DIFF
--- a/frontend/src/hooks/useFormattedDuration.ts
+++ b/frontend/src/hooks/useFormattedDuration.ts
@@ -11,9 +11,10 @@ export enum DateBucketType {
 }
 
 export function useDateBucketType(
-  date: dayjs.Dayjs,
+  dateConfig: dayjs.ConfigType,
   now = dayjs(),
 ): DateBucketType {
+  const date = dayjs(dateConfig);
   const days = now.diff(date, 'day');
 
   if (days < 7) {


frontend/src/components/ActivityDashboard/ActivityMaintenanceSection/RecentCommit.tsx
```diff
@@ -1,14 +1,15 @@
...
-  const isLoading = false;
-  const date = dayjs().subtract(11, 'months');
+  const { plugin } = usePluginState();
+  const { data: metrics, isLoading } = usePluginMetrics(plugin?.name);
+  const stats = metrics?.maintenance.stats;
 
   return (
     <RecentStats
-      date={date}
+      date={stats?.latest_commit_timestamp}
```
klai95, line 12: Just want to clarify: if the latest commit from within a week, it would say "latest commit made a week ago", and if it's beyond a week but within two weeks, it would say "latest commit made 2 weeks ago" right?

klai95 (dismissed):
Besides the comment I left, everything else looks good! Thanks Jeremy! 🎉

richaagarwal:
Exciting!  Let's see if we need to update the language "since plugin repo created" since we're not retrieving that date currently, pending internal discussion, before merging.

frontend/src/components/ActivityDashboard/ActivityMaintenanceSection/TotalCommits.tsx
```diff
@@ -1,17 +1,17 @@
...
+  const { data: metrics, isLoading } = usePluginMetrics(plugin?.name);
+  const stats = metrics?.maintenance.stats;
 
   return (
     <TotalStats
-      count={commits}
+      count={stats?.total_commits ?? 0}
       countI18nKey="activity:commits"
-      date={date}
+      date={plugin?.first_released}
```
richaagarwal, line 14: After digging into it a little more, I think we should go ahead and use the plugin repo's creation date and keep to the existing design.  So, we would want to set the creation date in https://github.com/chanzuckerberg/napari-hub/blob/main/frontend/src/utils/repo.ts and use that date here instead.

richaagarwal:
Exciting! I left one comment here that I think should be addressed before we merge: https://github.com/chanzuckerberg/napari-hub/pull/883/files#r1096268677

Also, please link to the related GitHub issue in the description

richaagarwal (dismissed):
🔥

frontend/src/components/ActivityDashboard/ActivityMaintenanceSection/RecentCommit.tsx
richaagarwal: @codemonkey800 Hmm I think there may be an existing bug with the first case, "less than a week ago", since at this line - https://github.com/chanzuckerberg/napari-hub/blob/main/frontend/src/hooks/useFormattedDuration.ts#L50 - we don't return anything for the less than a week case. I noticed while on the activity tab for https://dev-shared-frontend.dev.imaging.cziscience.com/plugins/napari-sim-processor, that it says "Latest commit made a week ago" despite the `latest_commit_timestamp` returned in https://6vwmwypt38.execute-api.us-west-2.amazonaws.com/dev-shared/metrics/napari-sim-processor indicating a date of 2/2/2023. 

We could open up a separate issue for it, but probably easiest to fix & verify as part of this PR?

klai95:
Just want to clarify, is the failure of `Frontend Tests / tsc (pull_request)` expected?

frontend/src/components/ActivityDashboard/RecentStats.tsx
```diff
@@ -43,7 +45,15 @@ export function RecentStats({
       )}
 
       {date && !durationI18nKey && (
-        <FormattedDuration date={date} isLoading={isLoading} />
+        <>
```
richaagarwal, line 48: Is there a reason this isn't specified in https://github.com/chanzuckerberg/napari-hub/blob/main/frontend/src/hooks/useFormattedDuration.ts#L50?

klai95 (approved):

frontend/src/components/ActivityDashboard/RecentStats.tsx
richaagarwal: Got it - so if I understand the logic correctly, if it's not less than a week ago, it would default to over a week ago?  Right now we say "a week ago" (and according to your earlier comment), so just wanted to confirm the "over" change is intended.

PR #888, opened 2023-02-06
Ingest all data for commit_activity.json & compute total_commits based on it

klai95:
`Related Issue`
https://github.com/chanzuckerberg/napari-hub/issues/820

`Description`
Based on the feedback from Manasa and Richa in https://github.com/chanzuckerberg/napari-hub/pull/848#discussion_r1092356074 and https://github.com/chanzuckerberg/napari-hub/pull/848#discussion_r1096389113, this follow-up PR re-structures how the backend logic handles querying and returning plugin commit activity in order to reduce the amount of rework in design and engineering for the future

`S3`
To view updates made to `commit_activity.json` and `latest_commits.json`, click on https://s3.console.aws.amazon.com/s3/buckets/napari-hub-staging?region=us-west-2&prefix=activity_dashboard_data/. Note that `total_commits.json` was not updated due to the changes introduced in this PR. See `(2)` in the `Changes` section for more context.

`Changes`
Before the changes introduced in this PR, `commit_activity.json` on `s3` only has the 12 month's of data for all plugins. With this current PR , the changes introduced include:
(1) `commit_activity.json` contains all-time commit activity for all plugins instead of only for 12 months
(2) `total_commits` is calculated by summing up all commits for each plugin directly using `commit_activity.json` to maintain parity with the implementation of usage metrics, and thus there is no longer a need for the existence and usage of `total_commits.json`

### DIFF
--- a/backend/api/_tests/test_activity_dashboard.py
+++ b/backend/api/_tests/test_activity_dashboard.py
@@ -21,22 +21,20 @@ MOCK_PLUGIN_COMMIT_ACTIVITY = [(1, 1), (2, 2), (3, 3), (4, 4), (5, 5), (6, 6), (
 class TestActivityDashboard(unittest.TestCase):
 
     @patch.object(model, 'get_latest_commit', return_value=None)
-    @patch.object(model, 'get_total_commit', return_value=None)
     @patch.object(model, 'get_commit_activity', return_value=None)
     @patch.object(model, 'get_recent_activity_data', return_value={})
     @patch.object(model, 'get_install_timeline_data', return_value=EMPTY_DF.copy())
-    def test_get_metrics_empty(self, mock_get_install_timeline_data, mock_get_recent_activity_data,  mock_get_commit_activity, mock_get_total_commit, mock_get_latest_commit):
+    def test_get_metrics_empty(self, mock_get_install_timeline_data, mock_get_recent_activity_data,  mock_get_commit_activity, mock_get_latest_commit):
         expected = self._generate_expected_metrics(
             timeline=self._generate_expected_timeline(-3, to_installs=lambda i: 0)
         )
-        self._verify_results('3', expected, mock_get_install_timeline_data, mock_get_recent_activity_data, mock_get_commit_activity, mock_get_total_commit, mock_get_latest_commit)
+        self._verify_results('3', expected, mock_get_install_timeline_data, mock_get_recent_activity_data, mock_get_commit_activity, mock_get_latest_commit)
 
     @patch.object(model, 'get_latest_commit', return_value=MOCK_PLUGIN_LATEST_COMMIT)
-    @patch.object(model, 'get_total_commit', return_value=MOCK_PLUGIN_TOTAL_COMMIT)
     @patch.object(model, 'get_commit_activity', return_value=MOCK_PLUGIN_COMMIT_ACTIVITY)
     @patch.object(model, 'get_recent_activity_data', return_value=MOCK_PLUGIN_RECENT_INSTALLS)
     @patch.object(model, 'get_install_timeline_data', return_value=MOCK_DF.copy())
-    def test_get_metrics_nonempty(self, mock_get_install_timeline_data, mock_get_recent_activity_data, mock_get_commit_activity, mock_get_total_commit, mock_get_latest_commit):
+    def test_get_metrics_nonempty(self, mock_get_install_timeline_data, mock_get_recent_activity_data, mock_get_commit_activity, mock_get_latest_commit):
         expected = self._generate_expected_metrics(
             timeline=self._generate_expected_timeline(-3),
             total_installs=sum(MOCK_INSTALLS),
@@ -45,14 +43,13 @@ class TestActivityDashboard(unittest.TestCase):
             total_commit=MOCK_PLUGIN_TOTAL_COMMIT,
             commit_activity=MOCK_PLUGIN_COMMIT_ACTIVITY
         )
-        self._verify_results('3', expected, mock_get_install_timeline_data, mock_get_recent_activity_data, mock_get_commit_activity, mock_get_total_commit, mock_get_latest_commit)
+        self._verify_results('3', expected, mock_get_install_timeline_data, mock_get_recent_activity_data, mock_get_commit_activity, mock_get_latest_commit)
 
     @patch.object(model, 'get_latest_commit', return_value=MOCK_PLUGIN_LATEST_COMMIT)
-    @patch.object(model, 'get_total_commit', return_value=MOCK_PLUGIN_TOTAL_COMMIT)
     @patch.object(model, 'get_commit_activity', return_value=MOCK_PLUGIN_COMMIT_ACTIVITY)
     @patch.object(model, 'get_recent_activity_data', return_value=MOCK_PLUGIN_RECENT_INSTALLS)
     @patch.object(model, 'get_install_timeline_data', return_value=MOCK_DF.copy())
-    def test_get_metrics_nonempty_zero_limit(self, mock_get_install_timeline_data, mock_get_recent_activity_data, mock_get_commit_activity, mock_get_total_commit, mock_get_latest_commit):
+    def test_get_metrics_nonempty_zero_limit(self, mock_get_install_timeline_data, mock_get_recent_activity_data, mock_get_commit_activity, mock_get_latest_commit):
         expected = self._generate_expected_metrics(
             total_installs=sum(MOCK_INSTALLS),
             installs_in_last_30_days=25,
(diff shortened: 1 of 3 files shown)


klai95:
By connecting to the `napari-hub-staging` bucket on `s3`, I have verified that `commit_activity.json` now includes all-time data for plugins, and thus showing that the logic is correct for `(1)` as detailed in the description of the PR. As for validating `(2)`, this is API response from testing the PR locally:
```
{'usage': {'timeline': [{'timestamp': 1643673600000, 'installs': 44}, {'timestamp': 1646092800000, 'installs': 134}, {'timestamp': 1648771200000, 'installs': 77}, {'timestamp': 1651363200000, 'installs': 122}, {'timestamp': 1654041600000, 'installs': 262}, {'timestamp': 1656633600000, 'installs': 94}, {'timestamp': 1659312000000, 'installs': 76}, {'timestamp': 1661990400000, 'installs': 87}, {'timestamp': 1664582400000, 'installs': 103}, {'timestamp': 1667260800000, 'installs': 107}, {'timestamp': 1669852800000, 'installs': 123}, {'timestamp': 1672531200000, 'installs': 148}], 'stats': {'total_installs': 2555, 'installs_in_last_30_days': 168}}, 'maintenance': {'timeline': [{'timestamp': 1643695200000, 'commits': 86}, {'timestamp': 1646114400000, 'commits': 18}, {'timestamp': 1648789200000, 'commits': 101}, {'timestamp': 1651381200000, 'commits': 49}, {'timestamp': 1654059600000, 'commits': 48}, {'timestamp': 1656651600000, 'commits': 4}, {'timestamp': 1659330000000, 'commits': 6}, {'timestamp': 1662008400000, 'commits': 8}, {'timestamp': 1664600400000, 'commits': 12}, {'timestamp': 1669874400000, 'commits': 8}, {'timestamp': 1672552800000, 'commits': 14}, {'timestamp': 1675231200000, 'commits': 4}], 'stats': {'latest_commit_timestamp': 1675354867000, 'total_commits': 815}}}

**Comment by klai95** [comment]
@codemonkey800 Just for further context, this PR does not introduce any naming or structural changes to the API response, and therefore frontend has the up-to-date changes that align with said API response.

File: backend/api/model.py [hunk 1]
```diff
@@ -94,6 +95,7 @@ def get_frontend_manifest_metadata(plugin, version):
     interpreted_metadata = parse_manifest(raw_metadata)
     return interpreted_metadata
 
+
```
richaagarwal, line 98: Are these extra lines introduced due to a linting check or extraneous?

backend/api/model.py
```diff
@@ -544,19 +521,17 @@ def _update_commit_activity(repo_to_plugin_dict):
...
     cursor_list = _execute_query(query, "GITHUB")
     data = {}
     for cursor in cursor_list:
         for row in cursor:
             repo = row[0]
             if repo in repo_to_plugin_dict:
                 plugin = repo_to_plugin_dict[repo]
-                data.setdefault(plugin, []).append({'timestamp': int(pd.to_datetime(row[1]).strftime("%s")) * 1000, 'commits': int(row[2])})
+                data.setdefault(plugin, []).append(
+                    {'timestamp': int(pd.to_datetime(row[1]).strftime("%s")) * 1000, 'commits': int(row[2])})
```
richaagarwal, line 534: Having the indices hardcoded here makes the code less readable and more susceptible to breaking changes in the future.  There should be ways to initialize the row with field names.

richaagarwal:
@klai95 Could you give this PR a more descriptive title?

backend/api/model.py
klai95: Due to a linting check - I did not add these lines

backend/api/model.py
```diff
@@ -9,7 +9,8 @@
 import pandas as pd
 from utils.github import get_github_metadata, get_artifact
 from utils.pypi import query_pypi, get_plugin_pypi_metadata
-from api.s3 import get_cache, cache, write_data, get_install_timeline_data, get_latest_commit, get_total_commit, get_commit_activity, get_recent_activity_data
+from api.s3 import get_cache, cache, write_data, get_install_timeline_data, get_latest_commit, \
```
DragaDoncila, line 12:
```suggestion
from api.s3 import get_cache, cache, write_data, get_install_timeline_data, get_latest_commit,
```

We should be able to omit the `\` if there's no whitespace after the comma

backend/api/model.py
```diff
@@ -544,19 +521,23 @@ def _update_commit_activity(repo_to_plugin_dict):
...
-            AND month >= dateadd(month, -12, DATE_TRUNC(month, CURRENT_DATE())) 
-            AND month < DATE_TRUNC(month, CURRENT_DATE())
         GROUP BY 1,2
     """
-    repo_to_plugin_dict = _get_repo_to_plugin_dict()
     cursor_list = _execute_query(query, "GITHUB")
     data = {}
+    fields = ['repo', 'month', 'num_commits']
     for cursor in cursor_list:
         for row in cursor:
```
DragaDoncila, line 530: We should be able to just split it out directly in the for statement and avoid the kind of confusing `dict_row` and the multiple `get` statements.

```suggestion
        for repo, month, num_commits in cursor:
```

backend/api/model.py
```diff
@@ -566,16 +547,22 @@ def get_metrics_for_plugin(plugin: str, limit: str) -> Dict:
...
+    timeline = []
+    maintenance_timeline = []
+    if _is_valid_limit(limit):
+        limit = int(limit)
+        timeline = _process_for_timeline(data, limit)
+        maintenance_timeline = commit_activity[-limit:]
 
     usage_stats = {
         'total_installs': install_stats.get('totalInstalls', 0),
         'installs_in_last_30_days': get_recent_activity_data().get(plugin, 0)
```
DragaDoncila, line 561: If `get_recent_activity_data()` is an API call/other similar request, I would move it above this final formatting bit. I think most people wouldn't expect heavy calls by this point in the function. Potentially same with `get_latest_commit` and summing the total commits below.

DragaDoncila (commented):
Left some pretty minor comments but it's looking good!

backend/api/model.py
klai95: Yeah, I think that definitely makes sense! If I remember correctly, @manasaV3 told me at some point that `installs_in_last_30_days': get_recent_activity_data().get(plugin, 0)` would be preferred, as opposed to having `get_recent_activity_data().get(plugin, 0)` defined as a variable and then passing it in to `usage_stat`, but want to open it up for discussion!

backend/api/model.py
kne42: agreed, provided that there is not missing data and that is why you're using `get` instead of indexing

if there is then i would suggest you clean the data beforehand

backend/api/_tests/test_activity_dashboard.py
```diff
@@ -97,7 +95,7 @@ def _generate_expected_timeline(start_range,
         return [{timestamp_key: to_timestamp(i), installs_key: to_installs(i)} for i in range(start_range, 0)]
 
     @staticmethod
-    def _generate_expected_metrics(timeline=None, total_installs=0, installs_in_last_30_days=0, latest_commit=None, total_commit=None, commit_activity=None):
+    def _generate_expected_metrics(timeline=None, total_installs=0, installs_in_last_30_days=0, latest_commit=None, total_commit=None, commit_activity=[]):
```
kne42, line 98: nit since you created a constant anyways
```suggestion
    def _generate_expected_metrics(timeline=None, total_installs=0, installs_in_last_30_days=0, latest_commit=None, total_commit=None, commit_activity=MOCK_PLUGIN_COMMIT_ACTIVITY_EMPTY):
```

backend/api/model.py
```diff
@@ -420,8 +421,8 @@ def _update_activity_timeline_data():
     write_data(csv_string, "activity_dashboard_data/plugin_installs.csv")
 
 
-def _is_not_valid_limit(limit):
-    return not limit.isdigit() or limit == '0'
+def _is_valid_limit(limit):
```
kne42, line 424: nit: i dont see a reason for a helper function when it is only used once and for a one-liner; i would rather opt to use a comment inline

backend/api/model.py
klai95: Feedback has been addressed!

backend/api/_tests/test_activity_dashboard.py
klai95: Feedback has been addressed!

backend/api/model.py
klai95: There shouldn't be any missing data, since Ben cleans the data beforehand. Otherwise, thanks for the feedback and it has been addressed!

backend/api/model.py
klai95: Feedback has been addressed!

backend/api/model.py
DragaDoncila: Yeah I see arguments for either. I don't personally mind as either work fine, so I think you should go with whatever you prefer here. My main concern is not having lots of critical functionality late in the function where it "looks" like all you're doing is formatting some data for response. Functionally of course it makes no difference 😊

DragaDoncila (approved):
This looks good to me! Thanks Kevin 🎉

kne42 (approved):
LGTM, great summary writeup btw!

PR #892, opened 2023-02-08
Create fixtures from API response

charles-testco:
This PR creates fixtures for E2E test based on responses from API. The result is a JSON file that can be used to test against.


charles-testco:
> oh looks like one of the test shards failed 🤔 might be flaky?

Hopefully it should be resolved by the next PRs

PR #906, opened 2023-02-14
Show empty state for no repo

codemonkey800:
## Description

#890

Shows empty state for maintenance section if the repo is not defined or a non-GitHub repo.

## Demos

image

PR #908, opened 2023-02-14
Update no code repo copy

codemonkey800:
## Description

Follow up of https://github.com/chanzuckerberg/napari-hub/pull/906#discussion_r1106150352 to update copy for when the maintenance section of the activity dashboard has no repo or is not on GitHub

## Demos

image

### DIFF
--- a/frontend/i18n/en/activity.json
+++ b/frontend/i18n/en/activity.json
@@ -11,7 +11,7 @@
     "allData": "No data could be fetched for this plugin. <hubLink href='$t(learnMoreLink)'>Learn more.</hubLink>",
     "monthly": "There is not yet enough data to display for this plugin",
     "preview": "Activity not available for previews",
-    "noCodeRepo": "Metrics not available, this plugin does not have a code repository specified."
+    "noCodeRepo": "Metrics not available, this plugin does not have a code repository specified or the repository is not on GitHub."
   },
   "monthlyInstalls": {
     "title": "Monthly installs in past year",


klai95 (approved):

frontend/i18n/en/activity.json
```diff
@@ -11,7 +11,7 @@
     "allData": "No data could be fetched for this plugin. <hubLink href='$t(learnMoreLink)'>Learn more.</hubLink>",
     "monthly": "There is not yet enough data to display for this plugin",
     "preview": "Activity not available for previews",
-    "noCodeRepo": "Metrics not available, this plugin does not have a code repository specified."
+    "noCodeRepo": "Metrics not available, this plugin does not have a code repository specified or the repository is not on GitHub."
```
richaagarwal, line 14: Hmm is this accurate though? I think right now we are only checking if the code repository is specified or not, not where it's sourced from, right?  Do we know what would happen if a non-github-repo were specified?

@bnelson-czi if a code repository were to point to some other source, would you still retrieve data for it?

richaagarwal: Ooh got it, thanks for pointing me to that!  I started a related thread in Slack in case we do want to consider non-github repos but this looks good for now then 👍🏼

richaagarwal (approved):

frontend/i18n/en/activity.json
bnelson-czi: We check if the code repository is specified AND if it's a github url. If not, we fetch no data on that plugin. I'm open to expanding our data ingestion to other platforms if this is a reoccurring issue.

Issue #909, opened 2023-02-15
Documentation and Project Site links on Plugin pages are clickable but nothing happens


haesleinhuepf:
#### Description

Since recently, it became much harder to access documentation and code from the napari-hub. The buttons do not work. I have the feeling one has to click very precisely  on the text to make the link work. Clicking on the button does not cause any action.

#### Steps/Code to Reproduce

* Go to this plugin for example: https://www.napari-hub.org/plugins/napari-spreadsheet
* Click the documentation / project site links/buttons

#### Expected Results

The documentation and/or project site opens.

#### Actual Results

Nothing happens.

https://user-images.githubusercontent.com/12660498/218978497-7c375be7-0bd8-464a-a17d-3f31aaad4e83.mp4

Hint:
I'm working with Google Chrome on Windows 10

PR #910, opened 2023-02-15
Fix click on plugin links

codemonkey800:
## Description

#909

Issue was that padding was specified on link container rather than link, so a portion of the container was not clickable. The only way to open the link would be by clicking on the text directly, this fixes it so that clicking on the entire menu item will open the link.

## Demos

https://user-images.githubusercontent.com/2176050/219170365-6cdee57f-feb0-4133-ba77-e6505ae83356.mp4


richaagarwal (approved):
Looks good to me. Do we have an idea of what introduced this bug?

Issue #900, opened 2023-02-09
Add Open Graph meta tags to each page


codemonkey800:
Open Graph meta tags allow us to enrich the appearance of links on social media. We currently have Open Graph tags implemented for collections: https://github.com/chanzuckerberg/napari-hub/pull/553

But we do not have it for other pages. The main things we can add to the Open Graph tags are:

- Title
- Type
- URL
- Image
- Description

https://github.com/chanzuckerberg/napari-hub/blob/36f7ce20c04b5e5862d60dbda7fdf6dc9d5fd84a/frontend/src/pages/collections/%5Bsymbol%5D.tsx#L84-L102

EricMaxWang:
@codemonkey800 split ticket into different tickets for: metadata vs. image service

Issue #895, opened 2023-02-09
Create optimized titles and descriptions for each page


codemonkey800:
Pages on the napari hub currently have titles and descriptions that are not optimal for SEO. The titles for each page have the following formats:

| Page                                                             | Page Title                                                        |
| ---------------------------------------------------------------- | ----------------------------------------------------------------- |
| [Home](https://www.napari-hub.org)                               | `napari hub \| search`                                            |
| [Plugin](https://www.napari-hub.org/plugins/napari-ism)          | `napari hub \| plugins \| napari-ism by Alessandro Zunino`        |
| [About](https://www.napari-hub.org/about)                        | `napari hub \| About`                                             |
| [FAQ](https://www.napari-hub.org/faq)                            | `napari hub \| FAQ`                                               |
| [Privacy](https://www.napari-hub.org/privacy)                    | `napari hub \| Privacy`                                           |
| [Contact](https://www.napari-hub.org/contact)                    | `napari hub \| Contact`                                           |
| [Collections Home](https://www.napari-hub.org/collections)       | `napari hub \| collections`                                       |
| [Collection](https://www.napari-hub.org/collections/alfa-cohort) | `napari hub \| collections \| Segmentation Plugins by the Alf...` |

The current suggestion is to:

1. Use dash instead of pipe for titles
2. Place target keywords at beginning of title
3. Limit titles to 51 - 60 characters
4. Limit descriptions to 150 - 160 characters

For example:

- https://www.napari-hub.org/plugins/stardist-napari
- Current meta title:  `napari hub | plugin | stardist-napari by Uwe Schmidt, Martin Weigert`
- Suggested meta title:  `stardist napari by Uwe Schmidt, Martin Weigert - plugins`

We can change the title / description to match the format specified in Danielle's [Hub metadata doc](https://docs.google.com/document/d/1pfkk6_8vY_8YcU4budv7AnZ_KLEQawHY67VJfT7YtR0/edit#heading=h.n0obsym12ejl).

EricMaxWang:
'Ready' pending edits to @dgmccart's metadata doc. Kevin will work on other two tickets first.

klai95:
PR for this issue: https://github.com/chanzuckerberg/napari-hub/pull/961

PR #913, opened 2023-02-16
Open graph meta tags

codemonkey800:
## Description

#900

Adds open graph meta tags to every page 

## Demos

image

### DIFF
--- a/frontend/src/pages/collections/[symbol].tsx
+++ b/frontend/src/pages/collections/[symbol].tsx
@@ -10,7 +10,7 @@ import { ErrorMessage } from '@/components/ErrorMessage';
 import { PageMetadata } from '@/components/PageMetadata';
 import { useLoadingState } from '@/context/loading';
 import { CollectionData } from '@/types/collections';
-import { getHubUrlByEnv } from '@/utils';
+import { createUrl } from '@/utils';
 import { hubAPI } from '@/utils/HubAPIClient';
 import { getServerSidePropsHandler } from '@/utils/ssr';
 import { getZodErrorMessage } from '@/utils/validate';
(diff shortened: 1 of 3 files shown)


klai95:
@codemonkey800 Is this the PR that needs to be merged before https://github.com/chanzuckerberg/napari-hub/issues/895 can be worked on?

frontend/src/pages/collections/[symbol].tsx
```diff
@@ -75,32 +75,20 @@ export default function Collections({ collection, error }: Props) {
 
   return (
     <>
-      <PageMetadata description={collection?.summary} />
+      <PageMetadata
+        description={collection?.summary}
+        title={collection?.title}
+        url={
+          collection
+            ? `https://napari-hub.org/collections/${collection.title}`
```
Comment (manasaV3) [inline frontend/src/pages/collections/[symbol].tsx:83]: Could this url be relative so we can point to the collection page in the same environment?

Also, does the title referenced here point to the name of collection, like "Segmentation Plugins by the Alfa Cohort" or the `id` like "alfa-cohort"? Because we link to the collections by their id, like https://www.napari-hub.org/collections/alfa-cohort, but we might want the metadata title to be the descriptive name.

frontend/src/utils/url.ts
```diff
@@ -69,3 +71,17 @@ export function setUrlHash(hash: string | undefined | null) {
...
+
+  if (PROD) {
+    base = 'https://napari-hub.org';
+  }
+
+  if (STAGING) {
+    base = 'https://staging.napari-hub.org';
+  }
+
+  return createUrl(path, base).href;
```
manasaV3, line 86: Would the DEV having a `localhost` work?

frontend/src/components/PageMetadata/PageMetadata.tsx
```diff
@@ -1,44 +1,92 @@
...
-import { usePageMetadata } from './usePageMetadata';
-
-interface Props {
-  description?: string;
-  keywords?: string[];
-}
+import { PageMetadataProps, usePageMetadata } from './usePageMetadata';
 
 /**
  * Renders meta tags for the current page. If keywords are provided, they'll be
```
klai95, line 12: Since new code has been added, just wondering if the comment section here is still reflective of the changes?

manasaV3 (approved):
:shipit:

Issue #852, opened 2023-01-30
Create infrastructure for modularizing ETL workflows


manasaV3:
This will create the required infrastructure pieces to move out the components for fetching data for installs from the backend lambda into a separate lambda.

Create a new lambda with an event trigger of SQS. Create a new event bridge rule, that schedules a message with structure `{'workflow': 'install'}` to the SQS that invokes the new lambda. Set the max retry for the SQS to 5 for production and 1 for dev and staging. Also, create a parameter store. 

The new components created in terraform should have a prefix. If the environment is dev, the prefix should be the stack name. If the environment is staging or prod, the environment can be the prefix. Pass the prefix to the lambda.

The ETL workflow lambda should have permission to perform `query` and `update` actions on the activity dynamo table.

PR #917, opened 2023-02-23
Adding data workflows ETL infrastructure

manasaV3:
## Description

Addresses https://github.com/chanzuckerberg/napari-hub/issues/852

#### docker-compose.yml
The docker-compose file provides the configuration updates for creating and managing the docker image for the new lambda.

#### data-workflows
The skeleton for the lambda codebase is saved under the data-workflows folder.

#### Lambda
The changes create the data workflow lambda infrastructure for processing the new workflows to write data to the dynamo. It creates the skeleton for the new data workflow module that can have its image pushed to ECR. The lambda has also been provided with some of the basic expected permissions. 

#### SQS
It also creates the data_workflow SQS and configures message publishing to it as a part of the existing event bridge rule trigger for activity. Permissions have been added to allow the rule to publish messages to the queue and for the lambda to read and delete messages. The queue has also been added as an event source to invoke the lambda.  

#### Parameter Store
The changes also create a parameter store to maintain the last_execution_timestamp to be used by subsequent executions. The required access permission to read and update it has been provided to the data workflow lambda. To prevent terraform from overwriting the data, the lifecycle policy has been configured to ignore the updates to the value and not overwrite it.  

## Pre-deployment Tasks

- [ ] Ensure that the new environment has the ECR repository created for data workflows images. This can be done by triggering the sci-imaging workflow manually for that environment.
- [ ] Verify in the secrets manager the relevant secret has been updated with the repo URL for data workflow


.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -182,6 +185,63 @@ module plugins_lambda {
...
+  environment             = {
+    "SNOWFLAKE_PASSWORD" = local.snowflake_password
+    "SNOWFLAKE_USER"     = local.snowflake_user
+    "STACK_NAME"         = local.custom_stack_name
+  }
+
+  log_retention_in_days   = 14
+  timeout                 = 300
+  memory_size             = 256
+  ephemeral_storage_size  = 512
```
klai95, line 210: Are all the values for these variables default ones? There are many values assigned to various variables in this file, so curious how the specific values were decided.

manasaV3: The `log_retention_in_days`, `memory_size`, and `ephemeral_storage_size` are set to be consistent with the other stacks. 

We could increase the log_retention period for all stacks as a separate change outside this PR. But that should be a part of a larger discussion with the team.

The `memory_size` for this stack could be reduced to 128MB, and we can increase that later if needed. 

The ephemeral storage is not charged for the 512MB, so we could leave that as is unless we have a specific reason to modify that. 

The `timeout` is set to 5 minutes, as that felt like a safe enough period for the execution to complete safely.

.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -182,6 +185,63 @@ module plugins_lambda {
...
+  visibility_timeout_seconds  = 300
+  tags                        = var.tags
+}
+
+resource aws_sqs_queue_policy data_workflows_queue_policy {
+  queue_url = aws_sqs_queue.data_workflows_queue.url
+  policy = data.aws_iam_policy_document.data_workflows_sqs_policy.json
+}
+
+resource "aws_lambda_event_source_mapping" "data_workflow_sqs_event_source_mapping" {
```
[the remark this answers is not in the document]
manasaV3: This has no significance. It uses quotes, as the similar component above used as a copy-pasta source had quotes on them. Fixing it for consistency. :)

potating-potato (approved):
in general it looks good to me functionally, but I didn't look into all details making sure it would work exactly, please test it on dev to make sure it is working properly

manasaV3:
> in general it looks good to me functionally, but I didn't look into all details making sure it would work exactly, please test it on dev to make sure it is working properly

Thank you for the review. I didn't find any issues when testing this in dev env. :)

Issue #896, opened 2023-02-09
Use canonical links


codemonkey800:
https://ahrefs.com/blog/canonicalization/

Canonical links is a way for us to specify the primary URL for all pages that can be derived from a single authoritative URL. This is important for the following scenarios:

1. Faceted navigation on home page: Given the search page is driven by search, filters, sort, and pagination, there can be a large combination of possible URLs because of the different possible query parameters. For example:
https://www.napari-hub.org/?search=napari&sort=relevance&license=oss&supportedData=2D&page=1
2. Plugin page category highlighting: When filtering by npe2 metadata, the hub will add additional query parameters to the URL to highlight the filtered metadata. For example: 
https://www.napari-hub.org/plugins/napari-stracking?workflowStep=Object+tracking
3. Prevent aliased deployments from being indexed on Google, like https://public-frontend.prod.imaging.cziscience.com

Because of this, it's important we specify a canonical URL for each page so that it points directly to the napari hub.

klai95:
@codemonkey800 Could you also provide a screenshot of the current behavior? I think it would make it more clear visually what the problem is.

PR #919, opened 2023-03-02
Use canonical links

klai95:
`Ticket`
- https://github.com/chanzuckerberg/napari-hub/issues/896

`Description`
- This PR adds canonical link to the frontend codebase so that we can specify the primary URL for all pages that can be derived from a single authoritative URL

`Screenshots`

Screen Shot 2023-03-02 at 6 47 01 PM

Screen Shot 2023-03-02 at 6 47 15 PM

### DIFF
--- a/frontend/src/pages/_app.tsx
+++ b/frontend/src/pages/_app.tsx
@@ -95,15 +95,14 @@ function App({ Component, pageProps }: AppProps) {
       }
 
       <Head>
+        <link rel="canonical" href={canonicalLink} />
+
         {/*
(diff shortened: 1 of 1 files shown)


klai95:
As seen in the `Screenshots` section of https://github.com/chanzuckerberg/napari-hub/pull/919#issue-1607525648, the canonical links for both the index/home page and plugin page do not include any query parameters, such as `https://www.napari-hub.org/` and `https://www.napari-hub.org/plugins/napari-test-plugin`.

Issue #853, opened 2023-01-30
Create activity install  ETL workflow


manasaV3:
In the new lambda's handler’s check if the invoking sqs message is for activity. If yes, execute the ETL.

- Fetch the `last_updated_timestamp` from the parameter store.
- Query snowflake view for all records inserted with ingestion timestamp after the `last_updated_timestamp` and until `current_timestamp`
- Transform the results into dynamo record of type DAY. Batch write all the records to dynamo.
- For all the plugins with new install activity, query snowflake for the install activity at a month level granularity for the months that had new records inserted. 
- Transform the results into dynamo record of type MONTH. Batch write all the records to dynamo.
- Query snowflake for the total installs for all the plugins with new install activity. 
- Transform the results into dynamo record of type TOTAL. Batch write all the records to dynamo.
- Update the parameter store with the `current_timestamp`

Some of data that could be useful to have logged:
- Time taken for each snowflake query
- Time taken for dynamo writes
- Number of records in each batch write
- Exceptions encountered

klai95:
Where can I find the schema for this ticket?

richaagarwal:
@klai95 This is a part of the "install activity migration" work outlined in #845, the schema is referenced in https://github.com/chanzuckerberg/napari-hub/issues/867

PR #926, opened 2023-03-07
ETL workflow for install activity

manasaV3:
Relates to #853 

### Summary:
This PR introduces the ETL workflow needed to identify and update the dynamo with the install activity data since the last run. It also adds logging for metrics surrounding the external IO operations.

#### Extraction
The timestamp of the last run is fetched from the parameter store. The snowflake database is then queried to get the plugin names with the earliest install activity added between the last run and now. Since their earliest install activity, the day, month, and total granularity data is recomputed for all those plugins. 

#### Transformation
The records fetched are transformed into the dynamo records of relevant types

#### Load
The records are batch written to the dynamo table.

### Miscellaneous
On successful completion of the workflow, the parameter store is updated with the end_timestamp used in the workflow.

### TO-DO:
- ~Logging~
- ~Unit Tests~


data-workflows/snowflake_adapter.py
```diff
@@ -0,0 +1,100 @@
...
+    return ' OR '.join([f"""LOWER(file_project) = '{name}' AND timestamp >= {_format_timestamp(timestamp_mapper(ts))}"""
+                        for name, ts in plugins_by_timestamp.items()])
+
+
+def _format_timestamp(timestamp):
+    datetime_obj = timestamp if type(timestamp) is datetime else datetime_from_millis(timestamp)
+    return f"""TO_TIMESTAMP('{datetime_obj.strftime("%Y-%m-%d %H:%M:%S")}')"""
+
+
+def _cursor_to_plugin_timestamp_mapper(accumulator: dict[str, datetime], row: List) -> dict[str, datetime]:
```
klai95, line 69: What are the contents of `row`?

data-workflows/activity/install_activity_model.py
```diff
@@ -0,0 +1,82 @@
...
+    InstallActivityType.MONTH: lambda timestamp: timestamp.replace(day=1),
+}
+
+
+class InstallActivity(Model):
+    class Meta:
+        prefix = os.getenv('STACK_NAME')
+        region = os.getenv('AWS_REGION')
+        table_name = f'{prefix}-install-activity'
+        region = region
```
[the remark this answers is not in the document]
manasaV3: Nothing says configuring settings like assigning the variable to itself. :)

data-workflows/snowflake_adapter.py
manasaV3: Yes, that is correct. :)

manasaV3:
> overall lgtm, but left one comment I was curious about before approving. also curious if you'll be adding a GitHub workflow for the data-workflows test in a separate PR

That is a great call out, yes. I will add that as a separate PR.

data-workflows/activity/install_activity_model.py
```diff
@@ -0,0 +1,86 @@
...
+    MONTH = 2
+    TOTAL = 3
+
+    def get_timestamp_format(self) -> Callable[[datetime], Union[int, None]]:
+        if self is InstallActivityType.TOTAL:
+            return lambda timestamp: None
+        return lambda timestamp: timestamp.replace(tzinfo=timezone.utc).timestamp() * 1000
+
+    def get_query_timestamp_projection(self) -> str:
+        return '1' if self is InstallActivityType.TOTAL else f'DATE_TRUNC(\'{self.name}\', timestamp)'
```
kne42, line 27:
```suggestion
        return '1' if self is InstallActivityType.TOTAL else f"DATE_TRUNC('{self.name}', timestamp)"
```

data-workflows/activity/install_activity_model.py
```diff
@@ -0,0 +1,86 @@
...
+    type_timestamp = UnicodeAttribute(range_key=True)
+    granularity = UnicodeAttribute(attr_name='type')
+    timestamp = NumberAttribute(null=True)
+    install_count = NumberAttribute()
+    last_updated_timestamp = NumberAttribute(default_for_new=get_current_timestamp)
+
+    def __eq__(self, other):
+        if isinstance(other, InstallActivity):
+            return (self.plugin_name, self.type_timestamp, self.granularity, self.timestamp, self.install_count) == \
+                   (other.plugin_name, other.type_timestamp, other.granularity, other.timestamp, other.install_count)
```
kne42, line 56:
```suggestion
            return ((self.plugin_name, self.type_timestamp, self.granularity, self.timestamp, self.install_count) == 
                    (other.plugin_name, other.type_timestamp, other.granularity, other.timestamp, other.install_count))
```

and format with e.g. `black`

data-workflows/activity/install_activity_model.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def get_query_timestamp_projection(self) -> str:
+        return '1' if self is InstallActivityType.TOTAL else f'DATE_TRUNC(\'{self.name}\', timestamp)'
+
+    def get_type_timestamp_formatter(self) -> Callable[[datetime], str]:
+        if self is InstallActivityType.TOTAL:
+            return lambda timestamp: 'TOTAL:'
+        elif self is InstallActivityType.DAY:
+            return lambda timestamp: f'DAY:{timestamp.strftime("%Y%m%d")}'
+        return lambda timestamp: f'MONTH:{timestamp.strftime("%Y%m")}'
```
kne42, line 34: as discussed, I think this kind of logic belongs more in a dictionary

data-workflows/activity/install_activity_model.py
klai95, line 27: Why do we return the value of '1' if `InstallActivityType.TOTAL`?

data-workflows/activity/install_activity_model.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def get_timestamp_query_formatter(self) -> Callable[[datetime], datetime]:
```
klai95, line 36: I understand the method on a high level, but could you provide context how you have decided to name it `get_timestamp_query_formatter`?

data-workflows/activity/install_activity_model.py
```diff
@@ -0,0 +1,86 @@
...
+            item = InstallActivity(plugin_name.lower(),
+                                   type_timestamp_format(timestamp),
+                                   granularity=activity_type.name,
+                                   timestamp=timestamp_format(timestamp),
+                                   install_count=activity['count'])
+            batch.save(item)
+            count += 1
+
+    batch.commit()
+    duration = (time.perf_counter_ns() - start) // 1000000
```
klai95, line 83: Since `time.perf_counter()` returns a value in seconds according to https://docs.python.org/3/library/time.html#time.perf_counter, I am wondering if `duration` could be calculated by the following:
```
start = time.perf_counter()
... // code between line 69 and line 82
duration = time.perf_counter() - start

File: data-workflows/activity/install_activity_model.py (hunk 3 above)
Comment (manasaV3) [reply data-workflows/activity/install_activity_model.py:27]: That is because we don't use any timestamp-specific functions in the projections of the query where this gets used. 
While we aggregate data at a month or a day level for the other two cases. 
Adding `1` would return `1` without adding more computation to the query.

File: data-workflows/activity/install_activity_model.py (hunk 5 above)
Comment (manasaV3) [reply data-workflows/activity/install_activity_model.py:34]: Oops! I had updated this to be an Enum with meaningful attributes set at initialization, so it made sense to have this method be inside the enum. Looks like those changes got lost. 😞

File: data-workflows/activity/install_activity_model.py (hunk 6 above)
Comment (manasaV3) [reply data-workflows/activity/install_activity_model.py:36]: As this provides a formatter for timestamp used in the subquery specific to the InstallActivityType.

File: data-workflows/activity/install_activity_model.py [hunk 8]
```diff
@@ -0,0 +1,92 @@
...
+    def __new__(cls, value, timestamp_formatter, type_timestamp_formatter):
+        install_activity_type = object.__new__(cls)
+        install_activity_type._value = value
+        install_activity_type.timestamp_formatter = timestamp_formatter
+        install_activity_type.type_timestamp_formatter = type_timestamp_formatter
+        return install_activity_type
+
+    DAY = (1, to_utc_timestamp_in_millis, lambda timestamp: f'DAY:{timestamp.strftime("%Y%m%d")}')
+    MONTH = (2, to_utc_timestamp_in_millis, lambda timestamp: f'MONTH:{timestamp.strftime("%Y%m")}')
+    TOTAL = (3, lambda timestamp: None, lambda timestamp: 'TOTAL:')
```
kne42, line 31: I'm still not a fan of overloading the enum with methods and complex values like this.

How about keeping this as a simple enum
```suggestion
    DAY = auto()
    MONTH = auto()
    TOTAL = auto()
```

then making separate dicts to map the types to the formats

also instead of using lambdas, I'd suggest using the built-in string formatting:
```python
type_timestamp_formats = {
   InstallActivityType.DAY: 'DAY:{0:"%Y%m%d"}',
   InstallActivityType.MONTH: 'MONTH:{0:"%Y%m"}',
   InstallActivityType.MONTH.TOTAL: 'TOTAL:',
}

type_format = type_timestamp_formats[my_activity_type].format(my_timestamp)
```

kne42:
Looking back, another option is to use a `namedtuple`:

```python
from collections import namedtuple
from functools import partial

InstallActivityType = namedtuple('InstallActivityType', ['type', 'timestamp_projection', 'timestamp_formatter', 'type_timestamp_formatter', 'timestamp_query_formatter'])
DAY = InstallActivityType('day', ...)
MONTH = InstallActivityType('month', "DATE_TRUNC('MONTH', timestamp)", to_utc_timestamp_in_millis, 'MONTH:{0:"%Y%m"}'.format, partial(datetime.replace, day=1)
TOTAL = ('total', ...)
```

data-workflows/activity/install_activity_model.py
```diff
@@ -0,0 +1,92 @@
...
+        return self.timestamp_formatter
+
+    def get_query_timestamp_projection(self) -> str:
+        return '1' if self is InstallActivityType.TOTAL else f"DATE_TRUNC('{self.name}', timestamp)"
+    
+    def get_type_timestamp_formatter(self) -> Callable[[datetime], str]:
+        return self.type_timestamp_formatter
+
+    def get_timestamp_query_formatter(self) -> Callable[[datetime], datetime]:
+        return (lambda ts: ts) if self is InstallActivityType.DAY else (lambda ts: ts.replace(day=1))
```
kne42, line 43: capturing explicitly here, you could use `functools.partial`:
```suggestion
        return (lambda ts: ts) if self is InstallActivityType.DAY else partial(datetime.replace, day=1)
```

data-workflows/activity/snowflake_adapter.py
```diff
@@ -0,0 +1,99 @@
...
+    if install_activity_type is InstallActivityType.TOTAL:
+        return f"""LOWER(file_project) IN ({','.join([f"'{plugin}'" for plugin in plugins_by_timestamp.keys()])})"""
+
+    timestamp_mapper = install_activity_type.get_timestamp_query_formatter()
+    return ' OR '.join([f"""LOWER(file_project) = '{name}' AND timestamp >= {_format_timestamp(timestamp_mapper(ts))}"""
+                        for name, ts in plugins_by_timestamp.items()])
+
+
+def _format_timestamp(timestamp):
+    datetime_obj = timestamp if type(timestamp) is datetime else datetime_from_millis(timestamp)
```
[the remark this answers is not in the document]
richaagarwal: I agree with Kevin here. The function `datetime_from_millis` doesn't provide any additional value over `datetime.fromtimestamp` other than converting milliseconds to seconds. That could be done as a separate line and would be more readable.

Similarly, I think this function overall (`_format_timestamp`) may not be necessary and may be an indication that some other code needs to be refactored.  For instance, the actual formatting is a single line that probably doesn't need its own function (`f"""TO_TIMESTAMP('{datetime_obj.strftime("%Y-%m-%d %H:%M:%S")}')"""`). 

It seems we're not sure what the data type of the variable being passed in is (`timestamp`) and need to convert it if it's not a datetime object. That logic should ideally happen as close to the source as possible, so that by the time we're formatting it, we know what data type we are working with.

data-workflows/activity/snowflake_adapter.py
```diff
@@ -0,0 +1,99 @@
...
+    return f"""TO_TIMESTAMP('{datetime_obj.strftime("%Y-%m-%d %H:%M:%S")}')"""
+
+
+def _cursor_to_plugin_timestamp_mapper(accumulator: dict[str, datetime], row: List) -> dict[str, datetime]:
```
richaagarwal, line 68: It took me a while to trace where a `timestamp_mapper` gets defined, to this function.  I also am not sure what this function is doing based on the function and variable naming.  Normally I'd recommend adding a docstring, but in this case I think maybe there's too many layers of abstraction from what we're trying to.

data-workflows/activity/install_activity_model.py
manasaV3: I feel keeping that a part of the enum closely couples the enum to this formatting. As this formatting is specific to each type of the enum. I think making it a map, adds distance between that relationship. What do you think?

data-workflows/activity/install_activity_model.py
```diff
@@ -0,0 +1,92 @@
...
+        install_activity_type.timestamp_formatter = timestamp_formatter
+        install_activity_type.type_timestamp_formatter = type_timestamp_formatter
+        return install_activity_type
+
+    DAY = (1, to_utc_timestamp_in_millis, lambda timestamp: f'DAY:{timestamp.strftime("%Y%m%d")}')
+    MONTH = (2, to_utc_timestamp_in_millis, lambda timestamp: f'MONTH:{timestamp.strftime("%Y%m")}')
+    TOTAL = (3, lambda timestamp: None, lambda timestamp: 'TOTAL:')
+
+    def get_timestamp_formatter(self) -> Callable[[datetime], Union[int, None]]:
+        return self.timestamp_formatter
```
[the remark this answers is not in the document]
manasaV3: I don't feel strongly about it either way. I can make this a method that returns the formatted value. :)

manasaV3:
>  Looking back, another option is to use a namedtuple:
I had considered this, but this will cause us to loose on the Enum representation. 😿

data-workflows/activity/install_activity_model.py
manasaV3: This is awesome. I didn't know we could do this in python as well.

data-workflows/activity/install_activity_model.py
```diff
@@ -0,0 +1,92 @@
...
+
+def to_utc_timestamp_in_millis(timestamp: datetime) -> int:
+    return int(timestamp.replace(tzinfo=timezone.utc).timestamp() * 1000)
+
+
+class InstallActivityType(Enum):
+
+    def __new__(cls, value, timestamp_formatter, type_timestamp_formatter):
+        install_activity_type = object.__new__(cls)
+        install_activity_type._value = value
```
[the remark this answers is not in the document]
richaagarwal: Adding to this, a good testing strategy is also to make sure tests fail before committing them - in a TDD environment it might be before implementation is complete, but you can also explicitly change a line in the relevant function after the fact to ensure a test fails in the expected way.

data-workflows/activity/install_activity_model.py
```diff
@@ -0,0 +1,89 @@
...
+class InstallActivityType(Enum):
+
+    def __new__(cls, value, timestamp_formatter, type_timestamp_formatter):
+        install_activity_type = object.__new__(cls)
+        install_activity_type._value_ = value
+        install_activity_type.timestamp_formatter = timestamp_formatter
+        install_activity_type.type_timestamp_formatter = type_timestamp_formatter
+        return install_activity_type
+
+    DAY = (1, to_utc_timestamp_in_millis, 'DAY:{0:%Y%m%d}')
```
richaagarwal, line 29: What do the 1, 2, and 3 here do?

data-workflows/utils/utils.py
```diff
@@ -0,0 +1,24 @@
...
+
+PARAMETER_NAME: str = f'/{os.getenv("STACK_NAME")}/napari-hub/data-workflows/config'
+
+
+def get_current_timestamp() -> int:
+    return round(time.time() * 1000)
+
+
+def get_last_updated_timestamp() -> int:
+    response = _get_ssm_client().get_parameter(Name=PARAMETER_NAME, WithDecryption=True)
```
klai95, line 15: nit: it seems like `_get_ssm_client()` is only being used in this file, so perhaps `boto3.client('ssm').get_parameter(Name=PARAMETER_NAME, WithDecryption=True)` is fine as is.

data-workflows/utils/utils.py
```diff
@@ -0,0 +1,24 @@
...
+    return json.loads(response['Parameter']['Value']).get('last_activity_fetched_timestamp')
+
+def set_last_updated_timestamp(timestamp) -> None:
+    value = json.dumps({'last_activity_fetched_timestamp': timestamp})
+    _get_ssm_client().put_parameter(Name=PARAMETER_NAME, Value=value, Overwrite=True)
```
klai95, line 20: nit: same feedback as https://github.com/chanzuckerberg/napari-hub/pull/926/files#r1143973946

data-workflows/handler.py
```diff
@@ -1,2 +1,29 @@
...
+    current_timestamp = utils.get_current_timestamp()
+    activity.processor.update_install_activity(last_updated_timestamp, current_timestamp)
+    utils.set_last_updated_timestamp(current_timestamp)
+
+
 def handle(event, context):
-    pass
+    _setup_logging()
+
+    for record in event.get('Records', []):
```
klai95, line 23: What are the structures of `event` and `record` in terms of the attributes they have?

data-workflows/utils/utils.py
manasaV3: As `_get_ssm_client()` method name by convention signifies it shouldn't be imported outside this module, I am a little confused why directly calling it would be better than the method.

data-workflows/handler.py
manasaV3: https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html#example-standard-queue-message-event

data-workflows/activity/install_activity_model.py
manasaV3: As Enums are always assigned numeric values when stored, I defined that value explicitly. It has no significance to our workflow. It could also be replaced by `auto()`.

data-workflows/utils/utils.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def _get_ssm_client():
```
richaagarwal, line 23: Another option would be to define a variable `SSM_CLIENT` up top since this only needs to be initialized once

data-workflows/utils/utils.py
richaagarwal: @manasaV3 I don't think that creating a new function for this does much to imply that we should not be using the `boto3` library outside this module, if that's your goal.  There are many other references to `boto3.client` elsewhere in the code. 

It is generally good practice to not initialize the client in multiple places, however this function doesn't help with that. See this comment https://github.com/chanzuckerberg/napari-hub/pull/926#discussion_r1144022687

manasaV3: @richaagarwal As you can see [here](https://github.com/chanzuckerberg/napari-hub/pull/926/commits/53e4b4f29e7b243040e00f6a244aa11cc723fbec#diff-3d195c90c58b0fbf3e3be9812e1589831ae1d85ea565a2a4d52c7035f6d57931R8), the ssm_client was initially initialized as recommended in [the comment](https://github.com/chanzuckerberg/napari-hub/pull/926#discussion_r1144022687). 

But as I increased the unit test coverage, I noticed that anywhere dependent on the util package also required setting up for `boto.client` as `ssm_client` would get initialized on the packages' import. I had it abstracted to a method. This, in turn, meant the `ssm_client` would only get initialized on explicit calls but multiple initializations. 

It has now been updated to a class where the initialization happens only once and is shared by the get* and set*. Currently, the same method calls both the getter and the setter, so the initialization gets done only once. But, if we have multiple places calling on it in the future, we would have to either make it a singleton class or handle it differently.

data-workflows/activity/install_activity_model.py
manasaV3: To ensure this is captured better, I have made this `auto()` and also added tests for the Enum. :)

klai95 (approved):
There is no more additional feedback on my end and it looks good to me! Awesome work and really appreciated your addressing all the feedback!

data-workflows/activity/install_activity_model.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class InstallActivityType(Enum):
+
+    def __new__(cls, timestamp_formatter, type_timestamp_formatter):
+        install_activity_type = object.__new__(cls)
+        install_activity_type._value_ = auto()
+        install_activity_type.timestamp_formatter = timestamp_formatter
+        install_activity_type.type_timestamp_formatter = type_timestamp_formatter
+        return install_activity_type
```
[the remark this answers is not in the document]
manasaV3: The `__new__` and `__init__` are called internally at different points. The `__new__` is the first step of instance creation. It's called first and is responsible for returning a new instance of your class. In contrast, `__init__` doesn't return anything.

PR #933, opened 2023-03-08
Run E2E tests only for frontend directory

codemonkey800:
## Description

#806 introduced a change where the frontend tests are run on every push. However, it makes more sense to run the E2E tests only when frontend changes are made, so this PR adds an event filter that checks if it's a push to `main` or to a pull request in the `frontend/*` directory

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -3,7 +3,17 @@
 
 name: E2E Tests
 
-on: push
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'frontend/**'
+  pull_request:
+    branches:
+      - '**'
+    paths:
+      - 'frontend/**'
 
 defaults:
   run:
@@ -142,7 +152,7 @@ jobs:
       - name: Upload artifacts
         if: ${{ failure() }}
         uses: actions/upload-artifact@v3.1.1
-        with: 
+        with:
           name: e2e-artifacts
           path: frontend/e2e/report
           retention-days: 1


richaagarwal:
@codemonkey800 Are there any scenarios where non-frontend changes might impact functionality tested as part of the e2e tests?

richaagarwal:
@codemonkey800 Yes, breaking changes to the schema as you've described could be one issue if not coordinated, as well as other unforeseen issues with data validation as we've encountered in the past.  It's also possible someone could do a large refactor and accidentally remove or rename endpoints.

Obviously when it's changes we are knowingly introducing we should do our best to ensure frontend and backend are aligned, but accidents/bugs happen & having tests would be one safeguard against this. 

I'm curious if this is motivated due to the time it takes for the CI checks to run, or any other concerns with having it run against backend changes as well?

richaagarwal:
@codemonkey800 Good point on infra changes - maybe we could just run on only frontend and backend dirs.  @manasaV3 @klai95 curious for your input as well!

klai95:
> @codemonkey800 Good point on infra changes - maybe we could just run on only frontend and backend dirs. @manasaV3 @klai95 curious for your input as well!

@codemonkey800 @richaagarwal I second on running on only frontend and backend dirs, as the frontend E2E test do take up quite some time (7-10 minutes), and at times I'd need to rerun them 1-2 times due to flaky tests.

richaagarwal:
ooh thanks for the additional context @codemonkey800 ! I was actually wondering about whether this hits backend code at all, so very good to know.  In that case, this PR looks good 👍🏼

richaagarwal (approved):

Issue #856, opened 2023-01-30
Update metric API usage.timeline to use dynamoDB


manasaV3:
If the incoming request is in the test, fetch the data for `usage.timeline` from dynamo, else fetch the results from s3.

- Query for records from activity_install with key condition expression `name=:plugin_name AND type_timestamp BETWEEN :start_month AND :end_date` and projection of installs and timestamp field.
```
aws dynamodb query \
--table-name dev-test_activity_install \
--key-condition-expression 'name = :plugin_name AND type_timestamp BETWEEN :start_date AND :end_month' \
--expression-attribute-values '{
":plugin_name": {"S": "foo"},
":start_month": {"S": "MONTH:202202"},
":end_month": {"S": "MONTH:202301"},
}'
--projection 'installs, timestamp'
```
In the above query, replace table_name with the correct table name by stage and stack name. Also, update the expression attribute value to have relevant values for all the fields. The `start_month` and `end_month` should be computed dynamically based on the current date.

- Then, transform the results from the dynamo to the expected format while adding records with a default value for installs for the missing months.

**Some of the data that could be useful to have logged:**

- Time taken for dynamo read
- If plugin has query results
- Total time taken for the computation of timeline

Issue #855, opened 2023-01-30
Update metric API usage.stats.installs_in_last_30_days to use dynamoDB


manasaV3:
If the incoming request is in the test, fetch the data for `usage.stats. installs_in_last_30_days` from dynamo, else fetch the results from s3.

- query for records from activity_install with key condition  expression 'name=:plugin_name AND type_timestamp BETWEEN :start_date AND :end_date’ and projection installs

```
aws dynamodb query \
--table-name dev-test_activity_install \
--key-conditions 'name = :plugin_name AND type_timestamp BETWEEN :start_date AND :end_date' \
--expression-attribute-values '{
":plugin_name": {"S": "foo"},
":start_date": {"S": "DAY:20221230"},
":end_date": {"S": "DAY:20230128"},
}'
--projection 'installs'
```

In the above query, replace table_name with the correct table name by stage and stack name. Also, update the expression attribute value to have relevant values for all the fields. The start_date and end_date should be computed dynamically based on the current date.

Aggregate the sum of the installs field across all the records fetched to find the value for `installs_in_last_30_days`. If there are no results, then the value for that is 0.

**Some of the data that could be useful to have logged:**

- Time taken for dynamo read
- If the plugin has query results
- Total time taken for the computation of installs_in_last_30_days

Issue #854, opened 2023-01-30
Update metric API usage.stats.total_installs to use dynamoDB


manasaV3:
If the incoming request is in the test, fetch the data for `usage.stats.total_installs` from dynamo else fetch the results from s3.

- getItem for records from activity_install  with key condition `name=:plugin_name AND type_timestamp=‘TOTAL:’` and projection `installs`

```
aws dynamodb getItem \
--table-name dev-test_activity_install \
--key-conditions 'name = :plugin_name AND type_timestamp = :type' \
--expression-attribute-values '{
":plugin_name": {"S": "foo"},
":type": {"S": "TOTAL:"}
}'
--projection 'installs'
```

In the above query, replace `foo` with the relevant plugin_name and table_name with the correct prefix. 

- If the result has a record, populate `total_installs` with the results from the records of type `TOTAL:`. Else set `total_installs` to `0`.

**Some of the data that could be useful to have logged:**

- Time taken for dynamo read
- If plugin has query results

PR #935, opened 2023-03-09
Updating metric API to use dynamoDB

manasaV3:
## Description
Addresses https://github.com/chanzuckerberg/napari-hub/issues/854, https://github.com/chanzuckerberg/napari-hub/issues/855, https://github.com/chanzuckerberg/napari-hub/issues/856

Queries dynamo DB for the relevant data and transforms them into the expected format.


backend/api/model.py
```diff
@@ -533,34 +535,42 @@ def _update_commit_activity(repo_to_plugin_dict):
     write_data(json.dumps(data), "activity_dashboard_data/commit_activity.json")
 
 
-def get_metrics_for_plugin(plugin: str, limit: str) -> Dict:
+def _get_usage_data(plugin, limit, in_test):
```
kne42, line 538: what's the `in_test` parameter for?

backend/api/model.py
```diff
@@ -533,34 +535,42 @@ def _update_commit_activity(repo_to_plugin_dict):
...
+        usage_stats = {
+            'total_installs': install_stats.get('totalInstalls', 0),
+            'installs_in_last_30_days': get_recent_activity_data().get(plugin, 0)
+        }
+        timeline = _process_for_timeline(data, limit) if limit else []
+
+    return {'timeline': timeline, 'stats': usage_stats, }
+
+
+def get_metrics_for_plugin(plugin: str, limit_str: str, in_test: bool) -> Dict:
```
kne42, line 557: could you add a docstring detailing what the parameters mean? the linked issues mention about being in a test but I lack context for what the test is for and what being in a test means

manasaV3: Added doc strings. :)

backend/api/install_activity.py
```diff
@@ -0,0 +1,80 @@
...
+def get_timeline(plugin: str, month_delta: int = 12) -> List[Dict[str, int]]:
+    """
+    Fetches plugin install data at a month level granularity from dynamo. For months that don't have data, we default to
+    zero.
+    Parameters:
+            plugin (str): Name of the plugin in lowercase for which usage data needs to be fetched.
+            month_delta (int): Specifies the number of months for results. (Defaults to 12)
+    Returns:
+            timeline (List[Dict[str, int]]): A list of entries for each month starting at the last month. It has the
+            structure {'timestamp': ..., 'installs': ...}
```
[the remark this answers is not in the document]
manasaV3: I think the zulip 's doc string mostly follows the [sphinx](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html) documentation style other than the return type. So, to be more consistent with a standard, I updated the doc strings to be stylized by sphinx recommendation for returns as well.

backend/api/install_activity.py
```diff
@@ -0,0 +1,74 @@
...
+
+def get_recent_installs(plugin: str, day_delta: int = 30) -> int:
+    """
+    Fetches plugin recent_install stats from dynamo.
+    :return int: sum of installs in the last day_delta timeperiod
+
+    :param str plugin: Name of the plugin in lowercase for which recent_install needs to be computed.
+    :param int day_delta: Specifies the number of days to include in the computation (Defaults to 30)
+    """
+    _day_type_format = lambda timestamp: f'DAY:{timestamp.strftime("%Y%m%d")}'
```
klai95, line 45: What is the naming convention you are using for `_day_type_format`? I understand that underscore is used to name helper methods.

backend/api/install_activity.py
```diff
@@ -0,0 +1,74 @@
...
+    today = datetime.date.today()
+    upper = _day_type_format(today)
+    lower = _day_type_format(today - relativedelta(days=day_delta))
+    results = InstallActivity.query(plugin, InstallActivity.type_timestamp.between(lower, upper))
+
+    return reduce(lambda acc, count: acc + count, [row.install_count for row in results], 0)
```
klai95, line 51: I know that in the docstring you state that the return value is the sum of installs in the last day_delta timeperiod of type `int`, but it isn't immediately obvious to me how this one-liner is achieving that goal

backend/api/install_activity.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def get_timeline(plugin: str, month_delta: int = 12) -> List[Dict[str, int]]:
```
klai95, line 54: If my understanding is correct, the data would not include the current month. If that is true, I think it would be helpful to mention that in the docstring.

backend/api/model.py
```diff
@@ -533,34 +536,56 @@ def _update_commit_activity(repo_to_plugin_dict):
...
+    """
+    Fetches plugin usage_data from s3 or dynamo based on the in_test variable
+    :returns (dict[str, Any]): A dict with the structure {'timeline': List, 'stats': Dict[str, int]}
+
+    :params str plugin: Name of the plugin in lowercase.
+    :params int limit: Sets the number of records to be fetched for timeline.
+    :params bool use_dyanmo: Fetch data from dynamo if True, else fetch from s3. (default= False)
+    """
+    if use_dynamo:
+        timeline = get_timeline(plugin, limit) if limit else []
```
klai95, line 549: What if `limit` is a negative number? There doesn't seem to be a guard for that condition

backend/api/install_activity.py
```diff
@@ -0,0 +1,74 @@
...
+
+    plugin_name = UnicodeAttribute(hash_key=True)
+    type_timestamp = UnicodeAttribute(range_key=True)
+    granularity = UnicodeAttribute(attr_name='type')
+    timestamp = NumberAttribute(null=True)
+    install_count = NumberAttribute()
+    last_updated_timestamp = NumberAttribute()
+
+
+def get_total_installs(plugin: str) -> int:
```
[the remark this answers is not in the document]
manasaV3: That is a great idea, I updated to include this.

backend/api/app.py
```diff
@@ -124,7 +124,17 @@ def update_activity() -> Response:
...
+    """
+    Fetches plugin metrics for usage, and maintenance
+    :return Response: A json object with entries for usage, and maintenance
+
+    :params str plugin: Name of the plugin in lowercase for which usage data needs to be fetched.
+    :query_params limit: Number of months to be fetched for timeline. (default=12).
+    :query_params use_dynamo_metric_usage: Fetch usage data from dynamo if True else fetch from s3. (default=False)
+    """
+    return jsonify(get_metrics_for_plugin(
+            plugin, request.args.get('limit', '12'), request.args.get('use_dynamo_metric_usage') == 'True'
```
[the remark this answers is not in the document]
manasaV3: I went with `True` as I was engrossed in the python ecosystem. But, as most languages use `true`, I have updated the check to be case insensitive. :)

backend/api/install_activity.py
manasaV3: map-reduce is a pattern where we transform a value and aggregate it in some form to an accumulator. This is the design pattern used here. 
The docs [here](https://docs.python.org/3/library/functools.html#functools.reduce) explains it a bit further, and might be worth looking into, to familiarize yourself with it.

backend/api/install_activity.py
manasaV3: Yes, you are right. I thought "over the last month_delta months" would clarify that. Let me know how we could reword it to make it clearer. :)

backend/api/model.py
manasaV3: That is a good catch. We didn't handle it in the existing code as well. So, adding that earlier in the code so the validation is done for all use cases to prevent unexpected behavior.

backend/api/metrics.py
```diff
@@ -0,0 +1,71 @@
...
+        day_type_format = 'DAY:{0:%Y%m%d}'
+        today = datetime.date.today()
+        upper = day_type_format.format(today)
+        lower = day_type_format.format(today - relativedelta(days=day_delta))
+        results = InstallActivity.query(plugin, InstallActivity.type_timestamp.between(lower, upper))
+
+        return reduce(lambda acc, count: acc + count, [row.install_count for row in results], 0)
+
+    @staticmethod
+    def get_timeline(plugin: str, month_delta: int = 12) -> List[Dict[str, int]]:
```
richaagarwal, line 54: I really like that the parameter is more explicitly named `month_delta` here instead of limit!  What are your thoughts on changing the other references to `limit`?

backend/api/app.py
```diff
@@ -183,5 +193,10 @@ def add_header(response):
     return response
 
 
+def _is_query_param_true(param_name: str):
+    value = request.args.get(param_name)
```
klai95, line 197: Is there a reason why we can't do the following?
```
value = request.args.get(param_name).lower()
return value == 'true'

File: backend/api/metrics.py [hunk 12]
```diff
@@ -0,0 +1,71 @@
...
+        """
+        Fetches plugin install count at a month level granularity from dynamo over the last month_delta months.
+        :returns List[Dict[str, int]]: Entries for the month_delta months
+
+        :param str plugin: Name of the plugin in lowercase for which timeline data needs to be fetched.
+        :param int month_delta: Number of months in timeline. (Defaults to 12)
+        """
+        month_type_format = 'MONTH:{0:%Y%m}'
```
klai95, line 62: What exact format you are following for `MONTH:{0:%Y%m}`? Mainly asking to understand how this format can be applied to other similar formats with higher or lower granularities

backend/api/model.py
```diff
@@ -2,14 +2,17 @@
 from datetime import datetime
 import json
 import os
-from typing import Tuple, Dict, List, Callable
+from typing import Tuple, Dict, List, Callable, Any
 from zipfile import ZipFile
 from io import BytesIO
 from collections import defaultdict
 import pandas as pd
+
```
klai95, line 10: nit: is this newline needed?

backend/api/app.py
```diff
@@ -183,5 +193,10 @@ def add_header(response):
     return response
 
 
+def _is_query_param_true(param_name: str):
```
richaagarwal, line 196: Not sure what we're expecting from the frontend, but instead of defining this function, we should be able to check using `bool(request.args.get(param_name)` instead.

data-workflows/activity/tests/test_snowflake_adapter.py
```diff
@@ -40,7 +40,7 @@ def get_plugins_with_installs_in_window_query():
 
 def get_plugins_install_count_since_timestamp_query(projection, subquery):
     return f"""
-            SELECT
```
klai95, line 43: Is this line modified on accident?

data-workflows/activity/snowflake_adapter.py
```diff
@@ -36,7 +36,7 @@ def get_plugins_with_installs_in_window(start_millis: int, end_millis: int) -> d
 def get_plugins_install_count_since_timestamp(plugins_by_earliest_ts: dict[str, datetime],
                                               install_activity_type: InstallActivityType) -> dict[str, List]:
     query = f"""
-            SELECT
```
klai95, line 39: same feedback as https://github.com/chanzuckerberg/napari-hub/pull/935/files#r1148139104

backend/api/app.py
```diff
@@ -124,7 +124,17 @@ def update_activity() -> Response:
...
+    """
+    Fetches plugin metrics for usage, and maintenance
+    :return Response: A json object with entries for usage, and maintenance
+
+    :params str plugin: Name of the plugin in lowercase for which usage data needs to be fetched.
+    :query_params limit: Number of months to be fetched for timeline. (default=12).
+    :query_params use_dynamo_metric_usage: Fetch usage data from dynamo if True else fetch from s3. (default=False)
+    """
+    return jsonify(get_metrics_for_plugin(
+            plugin, request.args.get('limit', '12'), _is_query_param_true('use_dynamo_metric_usage')
```
richaagarwal, line 136: In general, it is good practice to explicitly pass in the parameter names as it makes for more readable code and also future-proofs for refactors/changes to ordering/new optional variables etc.  So the arguments would then read `plugin=plugin, limit=..., use_dynamo_for_usage=...`.

backend/api/app.py
manasaV3: if `request.args.get(param_name)` were `None`, doing `.lower()` on it would raise an Error.

backend/api/app.py
manasaV3: The expectation for this query parameter is if it is passed, it can have the value "true" or "false". But, doing `bool("false")` computes to `False`. That is why I am doing a comparison to "true". 
Also, I moved it to a method instead of inlining it to allow reuse for the other cases that will be using the same logic.

data-workflows/activity/snowflake_adapter.py
manasaV3: Nit: Can you please link to the comment instead of the line that was commented on?

backend/api/app.py
richaagarwal: Sounds good. Another way to design this would be to look for a parameter, e.g. `if request.args.get('use_dynamo_metric_usage')`, such that if the param is included at all then we assume it is true. I know that would require updating the API contract with Jeremy but just wanted to mention it as an option.

data-workflows/activity/tests/test_snowflake_adapter.py
manasaV3: Yes, looks like this happened during the rebase. All changes to the data-workflow folder need to be reverted. Thanks for catching this.

klai95 (approved):
No more blocking feedback on my end! Nicely done!

Issue #922, opened 2023-03-06
Create maintenance activity ETL workflow


manasaV3:
Update the data workflow lambda's existing activity ETL.

### Workflow

- Use the `last_updated_timestamp` fetched from the parameter store and `current_timestamp` created in the workflow.
- Query snowflake `imaging.github.commits` for all records inserted with ingestion_timestamp after the `last_updated_timestamp` and until `current_timestamp`
- For all the updated plugins, query snowflake for the maintenance activity at a month-level granularity for the months with new records inserted. 
- Transform the results into dynamo record of type MONTH. Batch write all the records to dynamo.
- Query snowflake `imaging.github.commits` for the total commits for all the plugins with new maintenance activity. 
- Transform the results into dynamo record of type TOTAL. Batch write all the records to dynamo.
-  Query snowflake `imaging.github.commits` for the latest commit for all the plugins with new maintenance activity. 
- Transform the results into dynamo record of type LATEST. Batch write all the records to dynamo.

### Schema

plugin_name - range key (name of the plugin)
type_identifier - sort key

| Field name  |  Attribute type |  Description |  Example |
| ----------- | ----------- | ----------- | ----------- |
| plugin_name      | String  | Hash Key |  foo
| type_identifier   | String  | Range Key |  LATEST:bar/baz, MONTH:202302:bar/baz, TOTAL:bar/baz
| timestamp | Number | Holds value for the latest timestamp in milliseconds.  |  1678132844
| number_of_commits | Number | Holds value for the number_of_commits. Part of the MONTH record  |  10
| repo | String | Name of the repository  |  bar/baz
| last_updated_timestamp | Number | timestamp of when the record was last updated | 1678132839

The timestamp record holds the value for period, when the record of `type=MONTH` and the latest commits's timestamp value when `type=LATEST`

The new dynamo table should have a prefix. If the environment is dev, the prefix should be the stack name. If the environment is staging or prod, the environment can be the prefix. Pass the prefix to the backend lambda.

### Logging

Some of data that could be useful to have logged:
- Time taken for each snowflake query
- Time taken for dynamo writes
- Number of records in each batch write
- Exceptions encountered

manasaV3:
Similar to #853

klai95:
Similar to https://github.com/chanzuckerberg/napari-hub/pull/926

klai95:
Pull request for this ticket: https://github.com/chanzuckerberg/napari-hub/pull/939

PR #927, opened 2023-03-07
Create github-activity dynamoDB table

klai95:
Issue: https://github.com/chanzuckerberg/napari-hub/issues/921
Reference: https://github.com/chanzuckerberg/napari-hub/pull/911
Description: The code changes introduced in this PR create the `github-activity` table in dynamoDB.

PR #939, opened 2023-03-10
Create maintenance activity ETL workflow

klai95:
[Issue](https://github.com/chanzuckerberg/napari-hub/issues/922)
[Reference](https://github.com/chanzuckerberg/napari-hub/pull/926)
[RFC ](https://docs.google.com/document/d/1WOYxiFU7YVwugYeZmHUR5ynIz9QwRegAp9TCI5uH3CU/edit#heading=h.cj432tj1lb6z)


data-workflows/activity/model.py
```diff
@@ -0,0 +1,19 @@
+import os
+from pynamodb.models import Model
+from pynamodb.attributes import UnicodeAttribute, NumberAttribute
+
+
+class MaintenanceActivity(Model):
```
richaagarwal, line 6: I know this still in draft mode, but just wanted to provide some early feedback: Let's try to be consistent with naming, I know we are calling this table GitHub Activity here - https://github.com/chanzuckerberg/napari-hub/pull/927/files.

data-workflows/activity/model.py
```diff
@@ -0,0 +1,27 @@
...
+    LATEST = 1
+    MONTH = 2
+    TOTAL = 3
+
+
+class MaintenanceActivity(Model):
+    class Meta:
+        prefix = os.getenv('STACK_NAME')
+        region = os.getenv('AWS_REGION')
+        table_name = f'{prefix}-install-activity'
```
richaagarwal, line 18: Table name should be updated here

klai95:
I ran the maintenance activity ETL workflow and wrote the data to the `dev-test-github-activity` for testing purposes. The contents of the table can be seen here: https://us-west-2.console.aws.amazon.com/dynamodbv2/home?region=us-west-2#item-explorer?table=dev-test-github-activity&maximize=true

data-workflows/activity/snowflake_adapter.py
```diff
@@ -97,15 +142,15 @@ def _cursor_to_timestamp_by_plugin_mapper(accumulator: dict[str, datetime], curs
...
     cursor record and added to the accumulator keyed on plugin name.
-    The cursor contains the fields plugin, timestamp, and count
+    The cursor contains the fields plugin, column_1, and column_2
     :param dict[str, List] accumulator: Accumulator that will be updated with new data
     :param SnowflakeCursor cursor:
     :returns: Accumulator after data from cursor has been added
    """
-    for plugin, timestamp, count in cursor:
-        accumulator.setdefault(plugin, []).append({'timestamp': timestamp, 'count': count})
+    for plugin, column_1, column_2 in cursor:
```
richaagarwal, line 152: Is there a reason you've renamed `timestamp` and `count` to `column_1` and `column_2`?

klai95: Yes there is. I have already discussed and aligned this matter with Manasa, but basically the reason is that for the `LATEST` workflow `column_2` contains the timestamp, whereas for the `MONTH` workflow `column_1` contains the timestamp.

If the above change was not made, I would be assigning `activity['count']` as the latest commit timestamp value, which would be very confusing. Therefore, that is why I have renamed `timestamp` and `count` more generally. 

For further context, `get_query_projection` in `data-workflows/activity/github_activity_model.py` shows the different query projections returned depending on `GitHubActivityType`. We can see that `column_2` of `GitHubActivityType.LATEST` contains latest commit as timestamp value, but `column_2` of of `GitHubActivityType.TOTAL` contains num_commits as numerical value.

richaagarwal: I may be missing something, but wouldn't it be possible to reverse the ordering of the fields in this line https://github.com/chanzuckerberg/napari-hub/pull/939/files#diff-428b7b3e68dae07e5b5d9a5bc45651fd070646d4f2ebb25b13db4a4ae6bbcb84R38 so that we don't have to make changes to the naming here?

data-workflows/activity/github_activity_model.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def format_to_timestamp(self, timestamp: datetime) -> Union[int, None]:
+        return self.timestamp_formatter(timestamp)
+
+    def format_to_type_identifier(self, identifier: str) -> str:
+        return self.type_identifier_formatter.format(identifier)
+
+    def get_query_projection(self) -> str:
+        if self is GitHubActivityType.LATEST:
+            return '1, to_timestamp(max(commit_author_date)) as latest_commit'
```
richaagarwal, line 38: In addition to my comment regarding the ordering of these fields, I'm wondering why we're saving `1` as the number of commits when the metric is for latest commit. I don't think we need to set the commit count at all, as it's not relevant to the metric, and makes reviewing the data a little more confusing than it needs to be.

data-workflows/activity/github_activity_model.py
```diff
@@ -0,0 +1,110 @@
...
+            plugin_name = repo_name.split('/')[1]
+            if plugin_name in excluded_plugins:
+                continue
+            elif plugin_name in processed_public_plugins:
+                # get value of plugin name (case-sensitive) since repo_name is in lower case
+                plugin_name = processed_public_plugins[plugin_name]
+
+            if activity_type.name == "LATEST":
+                timestamp = datetime_to_utc_timestamp_in_millis(activity['column_2'])
+                commit_count = activity['column_1']
```
manasaV3, line 89: From [here](https://github.com/chanzuckerberg/napari-hub/pull/939/files#diff-428b7b3e68dae07e5b5d9a5bc45651fd070646d4f2ebb25b13db4a4ae6bbcb84R37-R38), we can see that the value of the `column_1` is always going to be `1`.  
How do we benefit from storing that in the database for the `LATEST` record?

data-workflows/activity/github_activity_model.py
```diff
@@ -0,0 +1,110 @@
...
+            elif activity_type.name == "MONTH":
+                timestamp = date_to_utc_timestamp_in_millis(activity['column_1'])
```
manasaV3, line 91: Are we certain the object being returned here is not a datetime object as well?

data-workflows/activity/github_activity_model.py
```diff
@@ -0,0 +1,110 @@
...
+def transform_and_write_to_dynamo(data: dict[str, List], activity_type: GitHubActivityType) -> None:
+    LOGGER.info(f'Starting item creation for github-activity type={activity_type.name}')
+
+    batch = GitHubActivity.batch_write()
+
+    start = time.perf_counter()
+    count = 0
+    public_plugins = get_public_plugins()
+    excluded_plugins = get_excluded_plugins()
+    processed_public_plugins = {k.lower(): k for k, v in public_plugins.items()}
```
manasaV3, line 77: Can you please explain why we are doing this? Also, how are we accounting for the hidden plugins?

manasaV3 (changes requested):

data-workflows/activity/github_activity_model.py
```diff
@@ -0,0 +1,110 @@
...
+    for repo_name, github_activities in data.items():
+        for activity in github_activities:
+            plugin_name = repo_name.split('/')[1]
```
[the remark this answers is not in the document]
manasaV3: These are some of the plugins where the name doesn't match the repo name: `'misic-napari', 'napari-denoiseg', 'napari-lattice', 'napari-time-series-plotter', 'iacs-ipac-reader', 'napari-plot', 'napari-stl-exporter', 'napari-cellseg3d', 'napari-pyclesperanto-assistant', 'napari-nucleaizer', 'napari-PICASSO', 'napari-arboretum', 'morphospaces', 'napari-svetlana', 'napari_video', 'elastix-napari', 'recOrder-napari', 'koopa-viz', 'napari-PHILOW', 'napari-mrcfile-handler'`

data-workflows/activity/snowflake_adapter.py
richaagarwal: Also, if the above suggestion does not resolve it, or there's a scenario where we need to add additional fields, we should consider having specific functions for install vs github activity, as I think the variable naming here is important and we don't need to force this function to be extensible.

data-workflows/activity/github_activity_model.py
```diff
@@ -0,0 +1,110 @@
...
+        host = os.getenv('LOCAL_DYNAMO_HOST')
+        region = os.getenv('AWS_REGION')
+        table_name = f'{os.getenv("STACK_NAME")}-github-activity'
+
+    plugin_name = UnicodeAttribute(hash_key=True)
+    type_identifier = UnicodeAttribute(range_key=True)
+    granularity = UnicodeAttribute(attr_name='type')
+    timestamp = NumberAttribute(null=True)
+    commit_count = NumberAttribute()
+    repo = UnicodeAttribute(null=True)
```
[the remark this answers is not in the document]
klai95: Repo values should never be null, since `imaging.github.commits` have non-null values for repo

data-workflows/activity/github_activity_model.py
```diff
@@ -0,0 +1,110 @@
...
+from enum import Enum, auto
+from typing import List, Union
+import os
+
+from pynamodb.models import Model
+from pynamodb.attributes import UnicodeAttribute, NumberAttribute
+
+from utils.utils import get_current_timestamp, date_to_utc_timestamp_in_millis, datetime_to_utc_timestamp_in_millis
+
+from backend.api.model import get_public_plugins, get_excluded_plugins
```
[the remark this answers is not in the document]
manasaV3: To resolve this, we can create a plugin module, which temporarily returns results by replicating the functionality of reading data from s3.

data-workflows/activity/github_activity_model.py
manasaV3: We should be using logic referenced here: https://github.com/chanzuckerberg/napari-hub/blob/main/backend/api/model.py#L483

klai95: Note to self for changes: create a module that contains logic just like `_get_repo_to_plugin_dict()` in `data-workflows` for this lambda

data-workflows/activity/snowflake_adapter.py
```diff
@@ -53,6 +54,50 @@ def get_plugins_install_count_since_timestamp(plugins_by_earliest_ts: dict[str,
     return _mapped_query_results(query, 'PYPI', {}, _cursor_to_plugin_activity_mapper)
 
 
+def get_plugins_commit_count_since_timestamp(plugins_by_earliest_ts: dict[str, datetime],
+                                             github_activity_type: GitHubActivityType) -> dict[str, List]:
+    if github_activity_type is GitHubActivityType.MONTH:
+        plugins_by_formatted_timestamp = {plugin: ts.replace(day=1) for plugin, ts in plugins_by_earliest_ts.items()}
+        subquery = ' OR '.join(
+            [f"LOWER(repo) = '{name}' AND to_timestamp(commit_author_date) >= "f"{TIMESTAMP_FORMAT.format(ts)}"
+             for name, ts in plugins_by_formatted_timestamp.items()])
```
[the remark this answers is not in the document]
klai95: Just want to clarify, the suggested change should be the below right?
```
subquery = ' OR '.join( [f"LOWER(repo) = '{name}' AND to_timestamp(commit_author_date) >= "f"{TIMESTAMP_FORMAT.format(ts.replace(day=1))}" for name, ts in plugins_by_earliest_ts.items()])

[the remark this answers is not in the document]
Comment (manasaV3) [reply data-workflows/activity/snowflake_adapter.py:63]: Good catch.. yes..

**Comment by klai95** [comment]
Verified that the changes are working as expected

Screenshot 2023-04-06 at 8 21 20 PM

File: data-workflows/plugin/s3.py [hunk 12]
```diff
@@ -0,0 +1,50 @@
...
+    """
+    index = get_cache('cache/index.json')
+    if index:
+        return index
+    else:
+        return {}
+
+
+def _get_repo_to_plugin_dict():
+    index_json = get_index()
```
manasaV3, line 45: Instead of having a get_index method, what are your thoughts on directly calling the get_cache, like below:

```
index_json = _get_cache('cache/index.json')
```

The get_cache handles only one use cases compared to the varied usecases it handles in the source where we duplicated it from. So, it can directly return an empty dictionary. 
```
def _get_cache(key: str) -> Dict:
    try:
        bucket = os.getenv('BUCKET')
        bucket_path = os.getenv('BUCKET_PATH', '')
        s3_client = boto3.client("s3")
        return json.loads(s3_client.get_object(Bucket=bucket, Key=os.path.join(bucket_path, key))['Body'].read())
    except ClientError:
        ....
        return {}
```

data-workflows/plugin/s3.py
```diff
@@ -0,0 +1,50 @@
...
+    repo_to_plugin_dict = {}
+    for plugin_obj in index_json:
+        if 'code_repository' in plugin_obj and plugin_obj['code_repository']:
+            repo_to_plugin_dict[plugin_obj['code_repository'].replace('https://github.com/', '')] = plugin_obj['name']
```
manasaV3, line 49: Can we simplify this to:
```
code_repository = plugin_obj.get('code_repository')
if code_repository:
    repo_to_plugin_dict[code_repository.replace('https://github.com/', '')] = plugin_obj['name']
```

data-workflows/activity/snowflake_adapter.py
```diff
@@ -50,7 +51,40 @@ def get_plugins_install_count_since_timestamp(plugins_by_earliest_ts: dict[str,
...
             """
     LOGGER.info(f'Fetching data for granularity={install_activity_type.name}')
-    return _mapped_query_results(query, 'PYPI', {}, _cursor_to_plugin_activity_mapper)
+    return _mapped_query_results(query, 'PYPI', {}, _cursor_to_plugin_install_activity_mapper)
+
+
+def get_plugins_with_commits_in_window(start_millis: int, end_millis: int) -> dict[str, datetime]:
+    query = f"""
+            SELECT 
+                repo AS name, DATE_TRUNC('DAY', TO_DATE(min(commit_author_date))) AS earliest_timestamp
```
manasaV3, line 60: Do we need the `TO_DATE(..)` on the `min(commit_author_date)`?

data-workflows/activity/snowflake_adapter.py
```diff
@@ -50,7 +51,40 @@ def get_plugins_install_count_since_timestamp(plugins_by_earliest_ts: dict[str,
...
+            SELECT 
+                repo AS name, DATE_TRUNC('DAY', TO_DATE(min(commit_author_date))) AS earliest_timestamp 
+            FROM 
+                imaging.github.commits  
+            WHERE 
+                repo_type = 'plugin'
+                AND TO_TIMESTAMP(ingestion_time) > {_format_timestamp(timestamp_millis=start_millis)}
+                AND TO_TIMESTAMP(ingestion_time) <= {_format_timestamp(timestamp_millis=end_millis)}
+            GROUP BY repo
+            ORDER BY repo
```
manasaV3, line 68: This needs to be updated in accordance with the projection.

data-workflows/activity/snowflake_adapter.py
```diff
@@ -17,7 +18,7 @@
 def get_plugins_with_installs_in_window(start_millis: int, end_millis: int) -> dict[str, datetime]:
     query = f"""
             SELECT 
-                LOWER(file_project) AS plugin, DATE_TRUNC('DAY', MIN(timestamp)) AS earliest_timestamp
+                LOWER(file_project) AS name, DATE_TRUNC('DAY', MIN(timestamp)) AS earliest_timestamp
```
manasaV3, line 21: I believe this was missed in the existing code as well, we should update the group_by, and order_by to use the `name` not `file_project`.

data-workflows/activity/snowflake_adapter.py
```diff
@@ -50,7 +51,40 @@ def get_plugins_install_count_since_timestamp(plugins_by_earliest_ts: dict[str,
...
+                                             github_activity_type: GitHubActivityType) -> dict[str, List]:
+    query = github_activity_type.get_query(plugins_by_earliest_ts)
+    LOGGER.info(f'Fetching data for granularity={github_activity_type.name}')
+
+    if github_activity_type is GitHubActivityType.LATEST:
+        accumulator_updater = _cursor_to_plugin_github_activity_latest_mapper
+    elif github_activity_type is GitHubActivityType.MONTH:
+        accumulator_updater = _cursor_to_plugin_github_activity_month_mapper
+    else:
+        accumulator_updater = _cursor_to_plugin_github_activity_total_mapper
```
manasaV3, line 85: This is nice. Definitely, something that we should consider incorporating for Install Activity as well in the future. 

What are your thoughts on having the accumulator_updater be a part of the GitHubActivityType enum, so we could fetch it as `github_activity_type.get_accumulator_updater()`?

data-workflows/activity/snowflake_adapter.py
```diff
@@ -109,6 +143,48 @@ def _cursor_to_plugin_activity_mapper(accumulator: dict[str, List], cursor) -> d
...
+    Updates the accumulator with data from the cursor for GitHubActivityType.MONTH.
+    Object with timestamp and count is created from the cursor record and added to the accumulator keyed on repo name.
+    The cursor contains the fields repo, timestamp, and count
+    :param dict[str, List] accumulator: Accumulator that will be updated with new data
+    :param SnowflakeCursor cursor:
+    :returns: Accumulator after data from cursor has been added
+   """
+    for repo, timestamp, count in cursor:
+        accumulator.setdefault(repo, []).append({'timestamp': timestamp, 'count': count})
+    return accumulator
```
manasaV3, line 171: I believe this and [_cursor_to_plugin_install_activity_mapper](https://github.com/chanzuckerberg/napari-hub/pull/939/files#diff-7380baa5c51177f0c5a8b84b254b7b5116c998a300a601840a67f38e780bc48cR132) are doing the exact same this.

manasaV3 (changes requested):

data-workflows/activity/snowflake_adapter.py
klai95: Yes. `Function DATE_TRUNC does not support VARCHAR(16777216) argument type` is displayed when `TO_DATE(...)` is not used.

data-workflows/activity/snowflake_adapter.py
klai95: I think that is a good idea!

data-workflows/activity/snowflake_adapter.py
klai95: They are very similar, but they are different in that `_cursor_to_plugin_install_activity_mapper` contains the field`plugin` while `_cursor_to_plugin_github_activity_month_mapper` contains the field `repo`

data-workflows/activity/snowflake_adapter.py
klai95: updated!

klai95:
Verified the code changes after review feedback are working as expected and the screenshot below matches the schema requirements of the ticket: https://github.com/chanzuckerberg/napari-hub/issues/922
Screenshot 2023-04-07 at 6 32 32 PM

data-workflows/tests/test_handler.py
```diff
@@ -30,7 +30,7 @@ def test_handle_valid_activity_event(self, monkeypatch):
         monkeypatch.setattr(utils.utils, 'get_current_timestamp', lambda: CURRENT_TIMESTAMP)
 
         from handler import handle
-        handle({'Records': [{'body': '{"type":"activity"}'}, {'body': '{"type":"bar"}'}]}, None)
+        handle({'Records': [{'body': '{"type":"install_activity"}'}, {'body': '{"type":"bar"}'}]}, None)
```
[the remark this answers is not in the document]
klai95: Because as of right now `test_handle_valid_activity_event` should only be testing `install_acitvity`. The way the code was structured before in `handler.py`, `_update_activity()` calls both `update_install_activity` and `update_github_activity`, which leads to the error provided in the screenshot below. 

Screenshot 2023-04-07 at 6 41 18 PM
GitHub Action: https://github.com/chanzuckerberg/napari-hub/actions/runs/4642105826/jobs/8215690608?pr=939

In contrast, this is the screenshot after the code changes have been implemented.
Screenshot 2023-04-07 at 6 48 27 PM
Note that the error in `TestHandle.test_handle_valid_activity_event` no longer appears.
GitHub Action: https://github.com/chanzuckerberg/napari-hub/actions/runs/4642173319/jobs/8215810116?pr=939

@manasaV3

klai95:
I have verified the changes after code review feedback are working as expected and the screenshot below matches the schema requirements of the ticket: https://github.com/chanzuckerberg/napari-hub/issues/922. Note that the current implementation adds commit data for hidden plugins as well on top of public plugins. As can be seen in the screenshot below, the table includes data for the `napari-threedee` plugin. 

Screenshot 2023-04-10 at 11 54 09 AM

klai95:
The current `failing check` error does not have anything to do with the code changes introduced in this PR, as verified with @manasaV3. She will have a fix out for this issue shortly.

manasaV3:
Wanted to make a call out, since the data has to be backfilled for github activity since the beginning of time, the `last_updated_timestamp` in the parameter store should be reset to the beginning of time.

data-workflows/activity/github_activity_model.py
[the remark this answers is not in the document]
klai95: The updated implementation takes care of this case

data-workflows/activity/github_activity_model.py
klai95: Yes

data-workflows/plugin/helpers.py
```diff
@@ -0,0 +1,157 @@
...
+                    filename_extensions = writer.get('filename_extensions', [])
+                    layer_types = writer.get('layer_types', [])
+                    for ext in filename_extensions:
+                        writer_file_extensions.add(ext)
+                    for layer_type in layer_types:
+                        if match := re.match(VALID_LAYER_REGEX, layer_type):
+                            writer_save_layers.add(match.groups()[0])
+                manifest_attributes['writer_file_extensions'] = list(writer_file_extensions)
+                manifest_attributes['writer_save_layers'] = list(writer_save_layers)
+        if 'themes' in manifest_contributions and manifest_contributions['themes']:
```
kne42, line 93:
```suggestion
        if manifest_contributions.get('themes'):
```

and similar down the chain

data-workflows/plugin/helpers.py
```diff
@@ -0,0 +1,157 @@
...
+            manifest_attributes['plugin_types'].append('theme')
```
kne42, line 94: you reuse `manifest_attributes['plugin_types']` a lot, i would put it in a variable

kne42 (approved):
LGTM. a few style requests. also i find it weird that in some places you use `Dict`/`List` from the `typing` library as annotations in certain places but use the built-in types elsewhere, sometimes mismatching them (e.g. i saw a `dict[str, List]`)

klai95:
> LGTM. a few style requests. also i find it weird that in some places you use `Dict`/`List` from the `typing` library as annotations in certain places but use the built-in types elsewhere, sometimes mismatching them (e.g. i saw a `dict[str, List]`)

Methods in `helps.py` are basically copied over from `model.py` under the `backend` folder because the `data-workflow` lambda will not have access to the methods in `model.py` at runtime. That being said, I do agree that it needs refactoring, but just calling out that we would need to make modifications in `model.py` as well, since that is where the code is coming from.

richaagarwal (changes requested):
@klai95 I saw a comment regarding the need to duplicate methods from `model.py` into `helpers.py`.  We should not be doing this.  Please make any changes/refactors necessary to avoid significant duplication of methods.

klai95:
> @klai95 I saw a comment regarding the need to duplicate methods from `model.py` into `helpers.py`. We should not be doing this. Please make any changes/refactors necessary to avoid significant duplication of methods.

As I have mentioned here https://github.com/chanzuckerberg/napari-hub/pull/939#issuecomment-1503877855, the reason is that the `data-workflow` lambda does not have access to methods such as `get_plugin`, etc provided in backend/model.py at runtime. This is something I have also discussed with @manasaV3 about.

data-workflows/activity/github_activity_model.py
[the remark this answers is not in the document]
klai95: @richaagarwal

klai95:
> Since the data has to be backfilled for github activity since the beginning of time, the `last_updated_timestamp` in the parameter store should be reset to the beginning of time for each environment to get the complete data.

Note to self: go to [/prod/napari-hub/data-workflows/config](https://us-west-2.console.aws.amazon.com/systems-manager/parameters/prod/napari-hub/data-workflows/config/description?region=us-west-2&tab=Table) and reset `last_activity_fetched_timestamp` to the beginning of time; do this for all environments

.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -243,6 +243,8 @@ module data_workflows_lambda {
     "SNOWFLAKE_PASSWORD" = local.snowflake_password
     "SNOWFLAKE_USER"     = local.snowflake_user
     "STACK_NAME"         = local.custom_stack_name
+    "BUCKET" = local.data_bucket_name
+    "BUCKET_PATH" = var.env == "dev" ? local.custom_stack_name : ""
```
manasaV3, line 247: Nit: Spacing

.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -410,6 +412,14 @@ data aws_iam_policy_document backend_policy {
 }
 
 data aws_iam_policy_document data_workflows_policy {
+  statement {
+    actions = [
+      "s3:PutObject",
```
manasaV3, line 417: We should not need the `s3:PutObject` permission. We only need the `s3:GetObject` for data-workflows.

manasaV3:
> > Since the data has to be backfilled for github activity since the beginning of time, the `last_updated_timestamp` in the parameter store should be reset to the beginning of time for each environment to get the complete data.
> 
> Note to self: go to [/prod/napari-hub/data-workflows/config](https://us-west-2.console.aws.amazon.com/systems-manager/parameters/prod/napari-hub/data-workflows/config/description?region=us-west-2&tab=Table) and reset `last_activity_fetched_timestamp` to the beginning of time (0); do this for all environments

You only need to set it to {'last_activity_fetched_timestamp': 0}

data-workflows/activity/github_activity_model.py
```diff
@@ -0,0 +1,151 @@
...
+            else:
+                timestamp = None
+                commit_count = activity['count']
+
+            item = GitHubActivity(plugin_name,
+                                  activity_type.format_to_type_identifier(identifier, repo),
+                                  granularity=activity_type.name,
+                                  timestamp=timestamp,
+                                  commit_count=commit_count,
+                                  repo=repo)
```
manasaV3, line 143: I believe this be simplified to:
```
timestamp = activity.get('timestamp')
commit_count = activity.get('count')
item = GitHubActivity(plugin_name,
                                  activity_type.format_to_type_identifier(repo, timestamp),
                                  granularity=activity_type.name,
                                  timestamp=activity_type.format_to_timestamp(timestamp),
                                  commit_count=commit_count,
                                  repo=repo)
```
We can leverage the `format_to_timestamp()` method, which was defined to be used just here.

data-workflows/activity/github_activity_model.py
```diff
@@ -0,0 +1,151 @@
...
+    batch = GitHubActivity.batch_write()
+
+    start = time.perf_counter()
+    count = 0
+    repo_to_plugin_dict = _get_repo_to_plugin_dict()
+    for repo, github_activities in data.items():
+        if repo in repo_to_plugin_dict:
+            plugin_name = repo_to_plugin_dict[repo]
+        else:
+            continue
```
manasaV3, line 125: Nit: While this code is functionally correct, it feels weird to assign a value only in the if block and use it in other places. What are your thoughts on formatting it to:
```
plugin_name = repo_to_plugin_dict.get(repo)
if not plugin_name:
    continue 
```

data-workflows/activity/github_activity_model.py
```diff
@@ -0,0 +1,151 @@
...
+            return f"""repo IN ({','.join([f"'{plugin}'" for plugin in plugins_by_earliest_ts.keys()])})"""
+
+    def get_query_sorting(self) -> str:
+        if self is GitHubActivityType.MONTH:
+            return 'repo, month'
+        else:
+            return 'repo'
+
+    def get_accumulator_updater(self) -> Callable:
+        import activity.snowflake_adapter as sf
```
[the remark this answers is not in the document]
manasaV3: It feels like it was done to avoid a circular dependency. This check could be a part of snowflake_adapter [here](https://github.com/chanzuckerberg/napari-hub/pull/939/files#diff-7380baa5c51177f0c5a8b84b254b7b5116c998a300a601840a67f38e780bc48cR34).

data-workflows/activity/snowflake_adapter.py
```diff
@@ -46,11 +47,40 @@ def get_plugins_install_count_since_timestamp(plugins_by_earliest_ts: dict[str,
                 download_type = 'pip'
                 AND project_type = 'plugin'
                 AND ({_generate_subquery_by_type(plugins_by_earliest_ts, install_activity_type)})
-            GROUP BY 1, 2
-            ORDER BY 1, 2
+            GROUP BY plugin, timestamp
+            ORDER BY plugin, timestamp
```
manasaV3, line 51: Nice clean up magic. ✨

data-workflows/activity/snowflake_adapter.py
```diff
@@ -46,11 +47,40 @@ def get_plugins_install_count_since_timestamp(plugins_by_earliest_ts: dict[str,
...
+
+def get_plugins_commit_count_since_timestamp(plugins_by_earliest_ts: dict[str, datetime],
+                                             github_activity_type: GitHubActivityType) -> dict[str, List]:
+    LOGGER.info(f'Fetching data for granularity={github_activity_type.name}')
+    return _mapped_query_results(
+        query=github_activity_type.get_query(plugins_by_earliest_ts),
+        schema="GITHUB",
+        accumulator={},
+        accumulator_updater=github_activity_type.get_accumulator_updater(),
+    )
```
manasaV3, line 83: This is very clean and easy to understand. 🔥

data-workflows/activity/snowflake_adapter.py
manasaV3: Could the name of the projection be changed to a common term if that would allow us to reuse the method?

data-workflows/handler.py
```diff
@@ -6,17 +6,27 @@
 import utils.utils
 
 
+def _fetch_and_update_timestamps():
+    parameter_store_adapter = ParameterStoreAdapter()
+    last_updated_timestamp = parameter_store_adapter.get_last_updated_timestamp()
+    current_timestamp = utils.utils.get_current_timestamp()
+    parameter_store_adapter.set_last_updated_timestamp(current_timestamp)
+    return last_updated_timestamp, current_timestamp
```
manasaV3, line 14: BLOCKER: Wouldn't this update the parameter store with current_timestamp even if the workflow fails?

data-workflows/handler.py
```diff
@@ -27,5 +37,7 @@ def handle(event, context):
             continue
         event_type = json.loads(record.get('body')).get('type')
 
-        if event_type == 'activity':
-            _update_activity()
+        if event_type == 'install_activity':
+            _update_install_activity()
+        elif event_type == 'github_activity':
+            _update_github_activity()
```
manasaV3, line 43: BLOCKER: This will break the workflow.

data-workflows/handler.py
```diff
@@ -6,17 +6,27 @@
...
-    current_timestamp = utils.utils.get_current_timestamp()
+def _update_install_activity() -> None:
+    last_updated_timestamp, current_timestamp = _fetch_and_update_timestamps()
     activity.processor.update_install_activity(last_updated_timestamp, current_timestamp)
-    parameter_store_adapter.set_last_updated_timestamp(current_timestamp)
+
+
+def _update_github_activity() -> None:
+    last_updated_timestamp, current_timestamp = _fetch_and_update_timestamps()
+    activity.processor.update_github_activity(last_updated_timestamp, current_timestamp)
```
manasaV3, line 29: BLOCKER: We are updating the same timestamp in two different workflows.

manasaV3 (changes requested):

data-workflows/activity/snowflake_adapter.py
klai95: Yes. Similar to how we have renamed `plugin` & `repo` to `name` in the SQL queries, we can make the same change here. As a result, we can rename  `_cursor_to_plugin_install_activity_mapper` to `_cursor_to_plugin_activity_mapper`, and get rid of `_cursor_to_plugin_github_activity_month_mapper`

data-workflows/activity/github_activity_model.py
```diff
@@ -0,0 +1,127 @@
...
+TIMESTAMP_FORMAT = "TO_TIMESTAMP('{0:%Y-%m-%d %H:%M:%S}')"
+
+
+class GitHubActivityType(Enum):
+    def __new__(cls, timestamp_formatter, type_identifier_formatter):
+        github_activity_type = object.__new__(cls)
+        github_activity_type._value = auto()
+        github_activity_type.timestamp_formatter = timestamp_formatter
+        github_activity_type.type_identifier_formatter = type_identifier_formatter
+        return github_activity_type
```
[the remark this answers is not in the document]
klai95: In this case, I believe we are using `__new__` to customize the creation process of `GitHubActivityType` instances by setting additional attributes (`timestamp_formatter` and `type_identifier_formatter`) on the object before it is returned as the new instance, and it  allows the `GitHubActivityType` instances to have these additional attributes immediately upon creation, without needing to set them in the `__init__` method.

data-workflows/activity/snowflake_adapter.py
```diff
@@ -46,13 +46,47 @@ def get_plugins_install_count_since_timestamp(plugins_by_earliest_ts: dict[str,
...
+            GROUP BY name
+            ORDER BY name
+            """
+    LOGGER.info(f'Querying for plugins added between start_timestamp={start_millis} end_timestamp={end_millis}')
+    return _mapped_query_results(query, 'GITHUB', {}, _cursor_to_timestamp_by_name_mapper)
+
+
+def get_plugins_commit_count_since_timestamp(plugins_by_earliest_ts: dict[str, datetime],
+                                             github_activity_type: GitHubActivityType) -> dict[str, List]:
+    if github_activity_type.name == 'LATEST':
```
kne42, line 75: these are enums so i would be comparing them directly instead of by their names

data-workflows/activity/snowflake_adapter.py
```diff
@@ -82,30 +116,58 @@ def _format_timestamp(timestamp_millis):
...
 
 def _cursor_to_plugin_activity_mapper(accumulator: dict[str, List], cursor) -> dict[str, List]:
     """
     Updates the accumulator with data from the cursor. Object with timestamp and count attributes are created from the
-    cursor record and added to the accumulator keyed on plugin name.
-    The cursor contains the fields plugin, timestamp, and count
+    cursor record and added to the accumulator keyed on name.
+    The cursor contains the fields name, timestamp, and count
+    :param dict[str, List] accumulator: Accumulator that will be updated with new data
+    :param SnowflakeCursor cursor:
```
kne42, line 138: what's this?

klai95: it is the accumulator updater method used by `get_plugins_install_count_since_timestamp` for `install activity` and `get_plugins_commit_count_since_timestamp` for `github-activity-LATEST`. The reason both are being used by this method is that they have the exact code in their methods.

Previously, I had separated out `_cursor_to_plugin_activity_mapper` into the two corresponding methods for `install activity` and for `github-activity-LATEST`, but I have since refactored the code to avoid duplication.

If you think there is a name for the method that captures it better, please do let me know! Thanks!

data-workflows/activity/snowflake_adapter.py
klai95: Feedback has been addressed!

data-workflows/activity/github_activity_model.py
```diff
@@ -0,0 +1,127 @@
...
+    commit_count = NumberAttribute(null=True)
+    repo = UnicodeAttribute()
+    last_updated_timestamp = NumberAttribute(default_for_new=get_current_timestamp)
+
+    def __eq__(self, other):
+        if isinstance(other, GitHubActivity):
+            return ((self.plugin_name, self.type_identifier, self.granularity, self.timestamp, self.commit_count,
+                     self.repo) ==
+                    (other.plugin_name, other.type_identifier, other.granularity, other.timestamp,
+                     other.commit_count, other.repo))
```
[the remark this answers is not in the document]
klai95: Feedback has been addressed!

data-workflows/activity/github_activity_model.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def format_to_type_identifier(self, repo_name: str, identifier_timestamp: str) -> str:
+        return self.type_identifier_formatter.format(repo_name, identifier_timestamp)
+
+    def get_query_projection(self) -> str:
+        if self is GitHubActivityType.LATEST:
+            return 'repo AS name, to_timestamp(max(commit_author_date)) as latest_commit'
+        elif self is GitHubActivityType.MONTH:
+            return 'repo AS name, date_trunc("month", to_date(commit_author_date)) as month, count(*) as commit_count'
+        return 'repo AS name, count(*) as commit_count'
```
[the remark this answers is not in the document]
klai95: Feedback has been addressed!

data-workflows/activity/github_activity_model.py
```diff
@@ -0,0 +1,127 @@
...
+                    f"{TIMESTAMP_FORMAT.format(ts.replace(day=1))}"
+                    for name, ts in plugins_by_earliest_ts.items()
+                ]
+            )
+        return f"""repo IN ({','.join([f"'{plugin}'" for plugin in plugins_by_earliest_ts.keys()])})"""
+
+    def get_query_sorting(self) -> str:
+        if self is GitHubActivityType.MONTH:
+            return 'name, month'
+        return 'name'
```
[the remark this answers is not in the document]
klai95: Feedback has been addressed! As a result, the methods `query_projection` and `query_sorting` are no longer needed

kne42 (dismissed):
Lots cleaner thanks!

data-workflows/activity/github_activity_model.py
```diff
@@ -92,6 +92,13 @@ def __eq__(self, other):
 
 
 def transform_and_write_to_dynamo(data: dict[str, List], activity_type: GitHubActivityType) -> None:
+    """
+    Transforms plugin commit data generated by get_plugins_commit_count_since_timestamp to the expected format
+    and then writes the formatted data to the corresponding github-activity dynamo table in each environment
+    :param dict[str, list] data: plugin commit data of type dictionary in which the key is plugin name
+    of type str and the value is Github activities of type list
+    :param GitHubActivityType activity_type:
+    """
```
kne42, line 101:
```suggestion
    """Transforms plugin commit data generated by get_plugins_commit_count_since_timestamp to the expected format
    and then writes the formatted data to the corresponding github-activity dynamo table in each environment

    :param dict[str, list] data: plugin commit data of type dictionary in which the key is plugin name
    of type str and the value is Github activities of type list
    :param GitHubActivityType activity_type:
    """
```

data-workflows/activity/snowflake_adapter.py
```diff
@@ -72,6 +72,15 @@ def get_plugins_with_commits_in_window(start_millis: int, end_millis: int) -> di
...
                                              github_activity_type: GitHubActivityType) -> dict[str, List]:
+    """
+    This method gets the commit data since a specific starting point for each plugin.
+    If GitHubActivityType.LATEST, fetch the latest commit timestamp, so construct query without commit count constraint
+    If GitHubActivityType.MONTH, fetch the sum of commits from the beginning of the month of the timestamp specified
+    for each of the plugin.
+    If GitHubActivityType.TOTAL, fetch the sum of commits over all time, so construct query without timestamp constraint
+    :param dict[str, datetime] plugins_by_earliest_ts: plugin name by earliest timestamp of commit record added
+    :param GitHubActivityType github_activity_type:
+    """
```
kne42, line 83:
```suggestion
    """This method gets the commit data since a specific starting point for each plugin.

    If GitHubActivityType.LATEST, fetch the latest commit timestamp, so construct query without commit count constraint
    If GitHubActivityType.MONTH, fetch the sum of commits from the beginning of the month of the timestamp specified
    for each of the plugin.
    If GitHubActivityType.TOTAL, fetch the sum of commits over all time, so construct query without timestamp constraint

    :param dict[str, datetime] plugins_by_earliest_ts: plugin name by earliest timestamp of commit record added
    :param GitHubActivityType github_activity_type:
    """
```

kne42 (approved):
some nits, otherwise lgtm

Issue #946, opened 2023-03-16
Add `www` to sitemap URLs


codemonkey800:
In Google Search Console, our sitemap URLs do not map to the actual URL people see on the napari hub. This is because the sitemap does not include `www` in the URL. I'm not 100% sure what effect this has on SEO, but it would be good practice to use the same URL users see in the sitemap.

PR #948, opened 2023-03-16
Add `www` to frontend URL

codemonkey800:
## Description

https://github.com/chanzuckerberg/napari-hub/issues/946

This adds `www` to the frontend URL variable in terraform so that the `sitemap.xml` file will use the correct URL. This is required only for production because we have a [www redirect defined in Cloudfront](https://github.com/chanzuckerberg/sci-imaging-infra/blob/25575a3a5b75ae24d141fad77d7f67e1ff2b9396/terraform/envs/prod/happy-napari-hub/cloudfront.tf#L56).

## Demos

Tested this locally by running `FRONTEND_URL=https://www.napari-hub.org yarn prod` and verified it's working in all [the places where it's being used in the frontend](https://github.com/search?q=repo%3Achanzuckerberg%2Fnapari-hub%20path%3Afrontend%2F%20FRONTEND_URL&type=code):

image

image

image

image

### DIFF
--- a/.happy/terraform/envs/prod/main.tf
+++ b/.happy/terraform/envs/prod/main.tf
@@ -23,7 +23,7 @@ module stack {
   env                   = local.env
   delete_protected      = false
   require_okta          = false
-  frontend_url          = "https://napari-hub.org"
+  frontend_url          = "https://www.napari-hub.org"
   backend_url           = "https://api.napari-hub.org"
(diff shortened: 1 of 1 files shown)


klai95 (approved):

manasaV3 (approved):
:shipit:

Issue #818, opened 2023-01-06
Integration tests for backend API


manasaV3:
To make the backend more robust, introduce integration tests for the API. The integration test should validate the response structure and the data returned by the API wherever possible.

[Cucumber](https://cucumber.io/) can be used to create readable tests.

Issue #947, opened 2023-03-16
Respond with 404 for plugins that return empty data


codemonkey800:
There's an issue on the frontend and backend right now where calling the API with a plugin is not indexed will return empty data. We should update both the frontend to display a 404 page, and the backend to respond with a 404 status code if the plugin object is empty

## Backend

<img width="490" alt="image" src="https://user-images.githubusercontent.com/2176050/225749063-52c49431-4bac-4d42-8ccc-9c92312a38a1.png">

## Frontend

<img width="1440" alt="image" src="https://user-images.githubusercontent.com/2176050/225749117-4f737689-d8ac-467d-bc7a-c8c6f13b99a3.png">

richaagarwal:
I've taken care of the backend portion in https://github.com/chanzuckerberg/napari-hub/pull/1226

@codemonkey800 Updating the current 404 page seems less critical, it's basic but I think it does the job? Let me know if you still think frontend work is needed for this.

PR #957, opened 2023-03-21
Behavioural test github workflow update

manasaV3:
## Description
Closes https://github.com/chanzuckerberg/napari-hub/issues/818

- This fixes the broken GitHub workflow to execute the integration tests and executes it against the dev environment
- It also updates the tests for metrics to use `usage` and `maintenance`


backend/bdd_tests/scenarios/metric.feature
```diff
@@ -3,27 +3,43 @@ Feature: metrics
...
     Given we call metrics api for napari-assistant
     Then response status is 200
-    And it should have 12 entries for timeline
-    And it should have at least one non-zero installs in timeline
-    And it should have non-zero values for stats
+    And it should only have properties usage, maintenance
+    And it should have 12 entries for usage.timeline
+    And it should have at least one non-zero installs in usage.timeline
+    And it should have non-zero values for usage.stats
+#    And it should have 10 entries for maintenance.timeline
```
richaagarwal, line 10: I know there is a related issue open for this https://github.com/chanzuckerberg/napari-hub/issues/958, but overall I am not sure we should be hardcoding these numbers. If we define a limit, it seems sufficient to look for <= to that limit.

backend/bdd_tests/scenarios/metric.feature
```diff
@@ -3,27 +3,43 @@ Feature: metrics
...
+    And it should have at least one non-zero commits in maintenance.timeline
+    And it should have non-zero values for maintenance.stats
 
   Scenario: metrics api for invalid plugin
```
richaagarwal, line 14: These tests make me think we should re-think what we return for invalid plugins.  Perhaps it could be tackled as part of https://github.com/chanzuckerberg/napari-hub/issues/947.

backend/bdd_tests/scenarios/metric.feature
```diff
@@ -3,27 +3,43 @@ Feature: metrics
...
     Given we call metrics api for napari-assistant with limit 5
     Then response status is 200
-    And it should have 5 entries for timeline
-    And it should have at least one non-zero installs in timeline
-    And it should have non-zero values for stats
+    And it should only have properties usage, maintenance
+    And it should have 5 entries for usage.timeline
+    And it should have at least one non-zero installs in usage.timeline
+    And it should have non-zero values for usage.stats
+#    And it should have 5 entries for maintenance.timeline
```
richaagarwal, line 32: Similar feedback as this comment - https://github.com/chanzuckerberg/napari-hub/pull/957/files#r1144121074

If the limit is 5, we should look for <= 5

backend/bdd_tests/scenarios/metric.feature
```diff
@@ -3,27 +3,43 @@ Feature: metrics
...
     Given we call metrics api for foo with limit 15
     Then response status is 200
-    And it should have 15 entries for timeline
-    And it should have all zero installs in timeline
-    And it should have zero values for stats and timelines
+    And it should only have properties usage, maintenance
+    And it should have 15 entries for usage.timeline
+    And it should have all zero installs in usage.timeline
+    And it should have zero values for usage.stats
+#    And it should have 15 entries for maintenance.timeline
```
richaagarwal, line 43: Similar feedback as this comment https://github.com/chanzuckerberg/napari-hub/pull/957/files#r1144122290

backend/bdd_tests/scenarios/metric.feature
klai95: I mentioned it https://github.com/chanzuckerberg/napari-hub/issues/958#issuecomment-1478796291, but I think it is relevant to bring up how we should approach filling in default values for missing data moving forward. Currently, it is done on the backend for usage metrics, and on the frontend for maintenance metrics. As we get clarity on that, I think it would address some of the confusion / expectation for these behavioral tests as well.

backend/bdd_tests/test_metric.py
```diff
@@ -16,40 +16,80 @@ def call_metrics_with_plugin_name(plugin_name, context):
...
+    assert any(item['commits'] > 0 for item in context['maintenance']['timeline'])
 
-@then(parsers.parse('it should have at least one non-zero installs in timeline'))
-def verify_timeline_has_any_non_zero_installs(context):
-    assert any(item['installs'] > 0 for item in context['activity']['timeline'])
 
+@then(parsers.parse('it should have non-zero values for usage.stats'))
+def verify_it_has_non_zero_usage_stats(context):
+    stats = context['usage']['stats']
+    assert stats['installs_in_last_30_days'] >= 0
```
klai95, line 52: Just curious, in which case would `stats['installs_in_last_30_days'] >= 0` be false?

manasaV3: Yes, but as we are checking for a specific plugin, we cannot confirm that the plugin has had installs in the last 30 days.
 
I think this brings up the bigger problem that might have to revisit on what we are testing.

backend/bdd_tests/scenarios/metric.feature
manasaV3: @richaagarwal, as @klai95 mentioned, the check against the limit is in place only because the current expectation is to pad the response for missing entries. We can also update this check if that expectation is to be modified. 

Note: Also, the hardcoded value should be 12. I was testing against 10, and it got committed accidentally.

klai95:
> ## Description
> Closes #818
> 
> * This fixes the broken GitHub workflow to execute the integration tests and executes it against the dev environment
> * It also updates the tests for metrics to use `usage` and `maintenance`

Correct me if I am wrong, but I thought we have decided to close issues manually after releases, instead of closing the issue as the PR gets merged in.

richaagarwal:
> > ## Description
> > Closes #818
> > 
> > * This fixes the broken GitHub workflow to execute the integration tests and executes it against the dev environment
> > * It also updates the tests for metrics to use `usage` and `maintenance`
> 
> Correct me if I am wrong, but I thought we have decided to close issues manually after releases, instead of closing the issue as the PR gets merged in.

@klai95 You can find the instructions on linking or closing issues with PRs in the pull request template: https://raw.githubusercontent.com/chanzuckerberg/napari-hub/main/.github/pull_request_template.md.  In this case, tests are updated immediately so no deployment is required to close the issue.

klai95 (approved):
Nicely done! No more blocking feedback. The only thing I'd like to say is that we should document or keep in mind these behavioral tests will continuously need to be updated based on the implementation of the API endpoints

Issue #866, opened 2023-01-30
Seed category data into dynamo


manasaV3:
To seed the table with data, read the s3 file for category and batch write to the table.

The key from json should be converted to lowercase and mapped to the `name` field. This will help in the category endpoint being case insensitive. Example: `"Ablation" -> "ablation"`

The `version_hash` field should be set to `EDAM-BIOIMAGING:alpha06:<hash>`, where `<hash>` is an MD5 hash calculated from the strings in the category object. This will help us support cases where a single category would have multiple records.

For example, the category `Scanning electron cryomicroscopy` has the following entries whose labels are the same dimension `Image modality` and value `Electron microscopy`, the only thing that differs is their hierarchy:

```json
[
  {
    "dimension": "Image modality",
    "hierarchy": [
      "Electron microscopy",
      "Cryo electron microscopy",
      "Scanning electron cryomicroscopy"
    ],
    "label": "Electron microscopy"
  },
  {
    "dimension": "Image modality",
    "hierarchy": [
      "Electron microscopy",
      "Scanning electron microscopy",
      "Scanning electron cryomicroscopy"
    ],
    "label": "Electron microscopy"
  }
]
```

The `version_hash` for each item will be `EDAM-BIOIMAGING:alpha06:8a970cd15928e947a86359f0559ef8fe` and `EDAM-BIOIMAGING:alpha06:61a42786c40c9d4bce561ff73ec00e08`

The `version` field should be set to `EDAM-BIOIMAGING:alpha06`. 

The `formatted_name` field can be used for storing the key as is. 

The `last_updated_timestamp` field can be generated from `round(time.time()*1000)`.

All the other fields can be mapped as is. 

We should document the script created here to be used in case we need to populate the data in the future. 
It could be saved inside the ETL workflow folder in the `/scripts` folder.

manasaV3:
In the interest of consistency, we should set this up as an ETL workflow that can be triggered by an SQS message. This will allow us to run this in different environments without any dependencies on the local environment.

Recommended structure for the SQS message body:

```
{ 
  "type": "category",
   "version": "EDAM-BIOIMAGING:alpha06"
   "s3_file_path": "/category/edam_bioimaging"
}
```

PR #959, opened 2023-03-22
Script for seeding category data from S3

codemonkey800:
## Description

#866

Depends on https://github.com/chanzuckerberg/napari-hub/pull/986

This PR adds a script for seeding category data from the S3 bucket. This works by adding a general purpose `run_workflow.py` script that can be run manually to run a particular workflow from CLI, or by importing the `run_workflow()` function. This will allow us to run the same scripts in our workflows locally and from the data-workflows lambda.

To seed category data from S3, run the `run_workflow.py` script:

```sh
AWS_PROFILE=sci-imaging \
  BUCKET=<bucket> \
  STACK_NAME=<stack-name>
  python run_workflow.py seed-s3-categories \
  --edam-version EDAM-BIOIMAGING:alpha06 \
  --s3_path category/EDAM-BIOIMAGING/alpha06.json
```

For example, this is the command I used for testing:

```sh
AWS_PROFILE=sci-imaging \
  BUCKET=napari-hub-dev \
  STACK_NAME=dev-test-cat \
  python run_workflow.py seed-s3-categories \
  --edam-version EDAM-BIOIMAGING:alpha06 \
  --s3_path category/EDAM-BIOIMAGING/alpha06.json
```

The workflow can also be run from Python using:

```py
from run_workflow import run_workflow

run_workflow({
    'type': 'seed-s3-categories',
    'edam_version': 'EDAM-BIOIMAGING:alpha06',
    's3_path': 'category/EDAM-BIOIMAGING/alpha06.json',
})
```

## Demos

List of first 50 categories in Dynamo

image

### DIFF
--- a/data-workflows/handler.py
+++ b/data-workflows/handler.py
@@ -2,7 +2,7 @@ import json
 import logging
 
 import activity.processor
-import categories
+import categories.processor
 
 
 LOGGER = logging.getLogger()
@@ -27,5 +27,5 @@ def handle(event, context) -> None:
             version = event.get("version")
             categories_path = event.get("categories_path")
 
-            categories.run_seed_s3_categories_workflow(version, categories_path)
+            categories.processor.seed_s3_categories_workflow(version, categories_path)
             LOGGER.info(f"Update successful for type={event_type}")
--- a/data-workflows/run_workflow.py
+++ b/data-workflows/run_workflow.py
@@ -3,7 +3,7 @@ Module containing functionality for running data workflows as a standalone scrip
 """
 
 import argparse
-import categories
+import categories.processor
 import logging
 import activity.processor
 
@@ -32,7 +32,7 @@ def run_workflow(event: Dict):
         version = event.get("version")
         categories_path = event.get("categories_path")
 
-        categories.run_seed_s3_categories_workflow(version, categories_path)
+        categories.processor.seed_s3_categories_workflow(version, categories_path)
         LOGGER.info(f"Update successful for type={event_type}")
 
 
--- a/data-workflows/categories/category_model.py
+++ b/data-workflows/categories/category_model.py
@@ -1,0 +1,26 @@
+import os
+import logging
+
+from pynamodb.attributes import ListAttribute, NumberAttribute, UnicodeAttribute
+from pynamodb.models import Model
+from utils.utils import get_current_timestamp
+
+STACK_NAME = os.getenv("STACK_NAME", "local")
+
+LOGGER = logging.getLogger()
+
+
+class CategoryModel(Model):
+    class Meta:
+        host = os.getenv("LOCAL_DYNAMO_HOST")
+        region = os.getenv("AWS_REGION", "us-west-2")
+        table_name = f"{STACK_NAME}-category"
+
+    name = UnicodeAttribute(hash_key=True)
+    version_hash = UnicodeAttribute(range_key=True)
+    version = UnicodeAttribute()
+    formatted_name = UnicodeAttribute()
+    dimension = UnicodeAttribute()
+    hierarchy = ListAttribute()  # List[str]
+    label = UnicodeAttribute()
+    last_updated_timestamp = NumberAttribute(default_for_new=get_current_timestamp)
--- a/data-workflows/categories/processor.py
+++ b/data-workflows/categories/processor.py
@@ -3,33 +3,15 @@ import time
 import hashlib
 import logging
 
-from pynamodb.attributes import ListAttribute, NumberAttribute, UnicodeAttribute
-from pynamodb.models import Model
+from categories.category_model import CategoryModel
 from slugify import slugify
 from typing import Dict
 from utils.env import get_required_env
 from utils.s3 import S3Client
-from utils.utils import get_current_timestamp
 
 STACK_NAME = os.getenv("STACK_NAME", "local")
 
 LOGGER = logging.getLogger()
-
-
-class CategoryModel(Model):
-    class Meta:
-        host = os.getenv("LOCAL_DYNAMO_HOST")
-        region = os.getenv("AWS_REGION", "us-west-2")
-        table_name = f"{STACK_NAME}-category"
-
-    name = UnicodeAttribute(hash_key=True)
-    version_hash = UnicodeAttribute(range_key=True)
-    version = UnicodeAttribute()
-    formatted_name = UnicodeAttribute()
-    dimension = UnicodeAttribute()
-    hierarchy = ListAttribute()  # List[str]
-    label = UnicodeAttribute()
-    last_updated_timestamp = NumberAttribute(default_for_new=get_current_timestamp)
 
 
 def _hash_category(category: Dict[str, str]) -> str:
@@ -53,7 +35,7 @@ def _hash_category(category: Dict[str, str]) -> str:
     return category_hash.hexdigest()
 
 
-def run_seed_s3_categories_workflow(version: str, categories_path: str):
+def seed_s3_categories_workflow(version: str, categories_path: str):
     """
     Runs data workflow for populating the category dynamo table from an S3
     source on the same depoyment stack.


data-workflows/run_workflow.py
```diff
@@ -0,0 +1,39 @@
...
+
+def run_workflow(event: Dict):
+    type = event['type']
+
+    if type == 'seed-s3-categories':
+        run_seed_s3_categories_workflow(
+            event['table'],
+            event['bucket'],
+            event['edam_version'],
+        )
```
manasaV3, line 16: For consistency, does it make sense to trigger this from the handler?

data-workflows/categories.py
```diff
@@ -0,0 +1,74 @@
...
+
+    return CategoryModel
+
+
+def hash_category(category: Dict[str, str]) -> str:
+    label = category.get('label', '')
+    dimension = category.get('dimension')
+    hierarchy = category.get('hierarchy', [])
+
+    h = hashlib.new('md5')
```
manasaV3, line 35: Nit: Could we rename this variable

data-workflows/categories.py
```diff
@@ -0,0 +1,74 @@
...
+    h.update(label.encode('utf-8'))
+    h.update(dimension.encode('utf-8'))
+
+    for value in hierarchy:
+        h.update(value.encode('utf-8'))
+
+    return h.hexdigest()
+
+def run_seed_s3_categories_workflow(table: str, bucket: str, edam_version: str):
+    print(f'Seeding category data from S3 version={edam_version} bucket={bucket} table={table}')
```
manasaV3, line 45: Nit: We could use logger here.

data-workflows/categories.py
```diff
@@ -0,0 +1,74 @@
...
+    edam_name, version = edam_version.split(':')
+    s3_path = f'category/{edam_name}/{version}.json'
```
manasaV3, line 47: We could be getting this as a complete path from the input data. That would allow us also specify paths of different formats.

data-workflows/categories.py
```diff
@@ -0,0 +1,74 @@
...
+
+from pynamodb.attributes import ListAttribute, NumberAttribute, UnicodeAttribute
+from pynamodb.models import Model
+from slugify import slugify
+from typing import Dict
+from utils.utils import S3Client, get_current_timestamp
+
+
+def get_category_model(table: str):
+    class CategoryModel(Model):
```
manasaV3, line 13: I believe this is inside a method because the table is passed a parameter. 

Because the category table name is always going to follow the format `<stack_name>-category`, what are your thoughts on fetching the prefix from the env variables, and appending it to a hard-coded category table name?

data-workflows/categories.py
```diff
@@ -0,0 +1,74 @@
...
+        dimension = UnicodeAttribute()
+        hierarchy = ListAttribute() # List[str]
+        label = UnicodeAttribute()
+        last_updated_timestamp = NumberAttribute(default_for_new=get_current_timestamp)
+
+    return CategoryModel
+
+
+def hash_category(category: Dict[str, str]) -> str:
+    label = category.get('label', '')
```
manasaV3, line 31: Just curious, did we ever encounter a case where we didn't have a label or hierarchy?

data-workflows/categories.py
```diff
@@ -0,0 +1,74 @@
...
+        h.update(value.encode('utf-8'))
+
+    return h.hexdigest()
+
+def run_seed_s3_categories_workflow(table: str, bucket: str, edam_version: str):
+    print(f'Seeding category data from S3 version={edam_version} bucket={bucket} table={table}')
+    edam_name, version = edam_version.split(':')
+    s3_path = f'category/{edam_name}/{version}.json'
+
+    client = S3Client(bucket)
```
manasaV3, line 49: We should also pass prefix here. 

As the bucket we are fetching from will depend on the env, what are your thoughts on getting the bucket and prefix from env variables?

manasaV3 (commented):
Really cool stuff..  🎉

data-workflows/categories.py
manasaV3: Could we please delete the picture that exposes the snowflake username, password?

klai95 (dismissed):

data-workflows/categories.py
manasaV3: > thinking about it more, I think what I need to do is first check if STACK_NAME == 'prod or STACK_NAME == 'staging, otherwise pass STACK_NAME as the bucket prefix

As the lambda will only have permissions from the s3 bucket in its environment, I think it makes sense to pass the bucket name as an env variable too. 
In the backend, we have a similar use case. To handle that, we set the bucket prefix env in terraform only for dev. 
Code Ref: https://github.com/chanzuckerberg/napari-hub/blob/main/.happy/terraform/modules/ecs-stack/main.tf#L179

data-workflows/activity/update_activity.py
```diff
@@ -0,0 +1,13 @@
+import activity.processor
+
+from utils.utils import ParameterStoreAdapter, get_current_timestamp
+
+
+def update_activity() -> None:
```
manasaV3, line 6: I believe this file got added incorrectly during a rebase.

data-workflows/categories.py
```diff
@@ -0,0 +1,95 @@
...
+    category_hash.update(label.encode("utf-8"))
+    category_hash.update(dimension.encode("utf-8"))
+
+    for value in hierarchy:
+        category_hash.update(value.encode("utf-8"))
+
+    return category_hash.hexdigest()
+
+
+def run_seed_s3_categories_workflow(edam_version: str, s3_path: str):
```
manasaV3, line 55: Nit: Not sure, if the versioning would always be an Edam version.

data-workflows/utils/env.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def get_required_env(key: str) -> str:
+    """
+    Utility for getting the value of an environment variable with the additional
+    constraint that it must be defined, otherwise the application will be unable
+    to run without it.
+    """
+
+    value = os.environ.get(key)
```
manasaV3, line 11: Nit: In most places, we currently use the `os.getenv(..)`, a wrapper around the `os.environ`. It should also return `None` if the value doesn't exist.

The only reason I'd recommend using `os.getenv` over this is consistency. What are your thoughts on that?

data-workflows/utils/env.py
```diff
@@ -0,0 +1,16 @@
+import os
+
+
+def get_required_env(key: str) -> str:
+    """
```
manasaV3, line 5: Could we update the documentation to use sphinx styling? 
https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html

data-workflows/utils/env.py
```diff
@@ -0,0 +1,16 @@
+import os
+
+
+def get_required_env(key: str) -> str:
```
manasaV3, line 4: Really cool idea. 😀

data-workflows/utils/s3.py
```diff
@@ -0,0 +1,42 @@
...
+    def __init__(self, bucket: str, prefix=""):
+        self._bucket = bucket
+        self._prefix = prefix
+        self._client = boto3.client("s3")
+
+    def _get_complete_path(self, s3_path: str):
+        return path.join(self._prefix, s3_path)
+
+    def _get_from_s3(self, s3_path):
+        obj = self._client.get_object(
```
manasaV3, line 27: Can we add a timing log around this s3 call?

data-workflows/handler.py
```diff
@@ -14,18 +14,19 @@ def _setup_logging():
...
-    for record in event.get('Records', []):
-        if 'body' not in record:
+    for record in event.get("Records", []):
+        if "body" not in record:
             continue
-        event_type = json.loads(record.get('body')).get('type')
 
-        if event_type == 'activity':
-            _update_activity()
+        payload = json.loads(record.get("body"))
```
manasaV3, line 31: We still need the existing check on the `event_type`. We would call the run_workflow, if `event_type == 'category'`

data-workflows/categories.py
```diff
@@ -0,0 +1,95 @@
...
+    bucket = get_required_env("BUCKET")
+
+    LOGGER.info(
+        f"Seeding {edam_version} category data from S3 "
+        f"s3_path={s3_path} bucket={bucket} table={CategoryModel.Meta.table_name}"
+    )
+
+    client = S3Client(
+        bucket=bucket,
+        prefix="" if STACK_NAME in ("prod", "staging") else STACK_NAME,
```
manasaV3, line 70: What are your thoughts on doing it similar to: https://github.com/chanzuckerberg/napari-hub/blob/main/.happy/terraform/modules/ecs-stack/main.tf#L179 ?

data-workflows/categories.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    Runs data workflow for populating the category dynamo table from an S3
+    source on the same depoyment stack.
+    """
+
+    bucket = get_required_env("BUCKET")
+
+    LOGGER.info(
+        f"Seeding {edam_version} category data from S3 "
+        f"s3_path={s3_path} bucket={bucket} table={CategoryModel.Meta.table_name}"
```
manasaV3, line 65: Can we also add prefix to the logged data, please.

data-workflows/categories.py
```diff
@@ -0,0 +1,95 @@
...
+
+from pynamodb.attributes import ListAttribute, NumberAttribute, UnicodeAttribute
+from pynamodb.models import Model
+from slugify import slugify
+from typing import Dict
+from utils.env import get_required_env
+from utils.s3 import S3Client
+from utils.utils import get_current_timestamp
+
+STACK_NAME = get_required_env("STACK_NAME")
```
manasaV3, line 14: Just a heads up, if this gets assigned here, you might have to set up mocks for this wherever this module gets imported.

data-workflows/run_workflow.py
```diff
@@ -0,0 +1,59 @@
...
+"""
+
+import argparse
+
+from activity.update_activity import update_activity
+from categories import run_seed_s3_categories_workflow
+from typing import Dict
+
+
+def run_workflow(event: Dict):
```
klai95, line 16: Nicely done! Having this workflow is super helpful for the future!

klai95 (dismissed):

data-workflows/utils/time.py
```diff
@@ -0,0 +1,10 @@
+import time
+
+
+def get_perf_duration(start: float) -> float:
+    return (time.perf_counter() - start) * 1000
+
+
+def print_perf_duration(start: float, message: str):
+    duration = get_perf_duration(start)
+    print(f"{message} duration={duration}ms")
```
manasaV3, line 10: Can we make this a log statement?

data-workflows/categories.py
```diff
@@ -0,0 +1,96 @@
...
+
+LOGGER = logging.getLogger()
+
+
+class CategoryModel(Model):
+    class Meta:
+        region = os.environ.get("AWS_REGION", "us-west-2")
+        table_name = f"{STACK_NAME}-category"
```
manasaV3, line 22: Can we also add host for local dynamo?

data-workflows/categories.py
```diff
@@ -0,0 +1,96 @@
...
+    return category_hash.hexdigest()
+
+
+def run_seed_s3_categories_workflow(version: str, s3_path: str):
+    """
+    Runs data workflow for populating the category dynamo table from an S3
+    source on the same depoyment stack.
+    """
+
+    bucket = get_required_env("BUCKET")
```
manasaV3, line 61: We should add this to the env variables of the lambda.

data-workflows/categories.py
```diff
@@ -0,0 +1,96 @@
...
+    env = get_required_env("ENV")
```
manasaV3, line 62: I believe we use the `env` only for determining the s3 prefix. What are your thoughts on having the s3_prefix also be an environment variable so that the workflow is environment agnostic?

data-workflows/handler.py
```diff
@@ -1,24 +1,39 @@
...
+            version = event.get("version")
+            s3_path = event.get("s3_path")
+
+            if not version:
+                LOGGER.error(f"Missing 'version' for type={event_type}")
+                return
+
+            if not s3_path:
+                LOGGER.error(f"Missing 's3_path' for type={event_type}")
+                return
```
manasaV3, line 36: 🔥

data-workflows/categories.py
klai95: As Manasa has mentioned, we would add another line of code
```
host = os.getenv('LOCAL_DYNAMO_HOST')

**Comment by manasaV3** [comment]
I believe we were holding off on incorporating changes from https://github.com/chanzuckerberg/napari-hub/pull/986 until Kevin's ETL PR was merged, to avoid conflicts that might be harder to resolve on his end.

File: data-workflows/handler.py [hunk 24]
```diff
@@ -1,31 +1,45 @@
...
 
-    for record in event.get('Records', []):
-        if 'body' not in record:
-            continue
-        event_type = json.loads(record.get('body')).get('type')
+        if event_type == "seed-s3-categories":
+            bucket = event.get("bucket")
+            version = event.get("version")
+            s3_path = event.get("s3_path")
+            s3_prefix = event.get("s3_prefix")
```
manasaV3, line 30: This should come from the `os.getenv` for consistency.

.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -243,6 +243,7 @@ module data_workflows_lambda {
     "SNOWFLAKE_PASSWORD" = local.snowflake_password
     "SNOWFLAKE_USER"     = local.snowflake_user
     "STACK_NAME"         = local.custom_stack_name
+    "ENV"                = var.env
```
manasaV3, line 246: We don't use the ENV environment variable anywhere. I think this got added instead of [this](https://github.com/chanzuckerberg/napari-hub/blob/56e863b93414dff3e57aa8cf7d98b6c5a5ea917f/.happy/terraform/modules/ecs-stack/main.tf#L178-L179).

data-workflows/handler.py
```diff
@@ -1,31 +1,45 @@
...
+        # TODO: Create a dict for event_type by method to be called
+        if event_type == "activity":
+            activity.processor.update_activity()
+            LOGGER.info(f"Update successful for type={event_type}")
 
-    for record in event.get('Records', []):
-        if 'body' not in record:
-            continue
-        event_type = json.loads(record.get('body')).get('type')
+        if event_type == "seed-s3-categories":
```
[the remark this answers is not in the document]
klai95: I think `elif` is a good callout.

data-workflows/handler.py
```diff
@@ -1,31 +1,30 @@
...
-        event_type = json.loads(record.get('body')).get('type')
 
-        if event_type == 'activity':
-            _update_activity()
+        body = record.get("body")
+        LOGGER.info(f"Received message with body: {body}")
+        event = json.loads(body)
+        event_type = event.get("type", "").lower()
+
+        # TODO: Create a dict for event_type by method to be called
```
klai95, line 21: Is this comment still needed?

data-workflows/activity/processor.py
```diff
@@ -4,20 +4,30 @@
...
 import activity.install_activity_model as install_model
 import activity.snowflake_adapter as snowflake_adapter
+from utils.utils import ParameterStoreAdapter
+import utils.utils
 
 LOGGER = logging.getLogger()
 
 
-def _fetch_data_and_write_to_dynamo(data: dict[str, datetime], install_activity_type: InstallActivityType):
+def _fetch_data_and_write_to_dynamo(data: dict[str, datetime], install_activity_type: InstallActivityType) -> None:
```
klai95, line 13: I think adding docstring in this method to detail exactly what data is would be helpful. It can also help clarify why it is being passed in `get_plugins_install_count_since_timestamp`

data-workflows/categories.py
```diff
@@ -0,0 +1,101 @@
...
+
+def _hash_category(category: Dict[str, str]) -> str:
+    """
+    Hashes a category object using the MD5 hash algorithm. This works by
+    creating a hash from the string and string array fields in the category
+    object.
+    """
+
+    label = category.get("label", "")
+    dimension = category.get("dimension")
```
klai95, line 43: Is there a reason why `dimension` does not have a default value?

data-workflows/run_workflow.py
```diff
@@ -0,0 +1,48 @@
...
+    )
+    subparsers = parser.add_subparsers(required=True, dest="type")
+
+    seed_s3_categories_parser = subparsers.add_parser(
+        "seed-s3-categories", help="categories help"
+    )
+    seed_s3_categories_parser.add_argument("--bucket", required=True)
+    seed_s3_categories_parser.add_argument("--version", required=True)
+    seed_s3_categories_parser.add_argument("--s3-path", required=True)
+    seed_s3_categories_parser.add_argument("--s3-prefix", default="")
```
manasaV3, line 35: Looking at this [snippet](https://github.com/chanzuckerberg/napari-hub/pull/959/files#diff-48bc4f460db59d31cd383f878f7ec7cb346afc564703a5025e6b180fe6070f93R56), we get the bucket and the path from os, variable and not from the event.  So, we would only need to fetch `version` and `s3-path` here.

data-workflows/run_workflow.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def run_workflow(event: Dict):
+    """
+    Function for running a particular data workflow based on the event type. The
+    event dictionary should contain all of the necessary data to run the
+    specified data workflow.
+    """
+
+    handler.handle({"Records": [{"body": json.dumps(event)}]}, {})
```
manasaV3, line 19: I recall we had landed on this as an approach earlier. But, with how modular the current categories file  now is we could simply call it as follows:
```
categories.run_seed_s3_categories_workflow(event.get("version"), event.get("s3_path"))
```

This would be similar to how we would call it in the handler. 

I feel that would reduce the complexity of the above statement and make it more intuitive here.

data-workflows/utils/s3.py
```diff
@@ -0,0 +1,46 @@
...
+        print_perf_duration(start, f'_get_from_s3("{s3_path}")')
+
+        return obj["Body"].read().decode("utf-8")
+
+    def load_json_from_s3(self, s3_path: str) -> Dict:
+        """
+        Load JSON file from S3 path and convert to a Python dictionary.
+        """
+
+        try:
```
manasaV3, line 42: Nit: Can we add timing logs here as well?

data-workflows/utils/s3.py
```diff
@@ -0,0 +1,46 @@
...
+class S3Client:
+    """
+    Client for accessing S3 resources.
+    """
+
+    _bucket: str
+    _prefix: str
+    _client: Any
+
+    def __init__(self, bucket: str, prefix=""):
```
manasaV3, line 20: Nit: prefix also needs typing.

data-workflows/utils/s3.py
```diff
@@ -0,0 +1,46 @@
...
+        self._bucket = bucket
+        self._prefix = prefix
+        self._client = boto3.client("s3")
+
+    def _get_complete_path(self, s3_path: str):
+        return path.join(self._prefix, s3_path)
+
+    def _get_from_s3(self, s3_path):
```
manasaV3, line 28: Nit: s3_path needs type

data-workflows/categories.py
```diff
@@ -0,0 +1,101 @@
...
+
+from pynamodb.attributes import ListAttribute, NumberAttribute, UnicodeAttribute
+from pynamodb.models import Model
+from slugify import slugify
+from typing import Dict
+from utils.env import get_required_env
+from utils.s3 import S3Client
+from utils.utils import get_current_timestamp
+
+STACK_NAME = os.getenv("STACK_NAME")
```
manasaV3, line 14: Can we add default value for this to be `local` as we now have a default dynamo setup.

data-workflows/handler.py
```diff
@@ -1,32 +1,46 @@
...
 
-import activity.processor
 from utils.utils import ParameterStoreAdapter
-import utils.utils
 
 
 def _setup_logging():
     logger = logging.getLogger()
     logger.setLevel(logging.INFO)
+    return logger
```
manasaV3, line 13: We could make this a constant so it is consistent with the other implementation. Untested code snippet alert:
```
LOGGER = logging.getLogger()
LOGGER.setLevel(logging.INFO)
```
With that, we would no longer need this method, and the log statements below could leverage this `LOGGER`.

data-workflows/tests/test_handler.py
```diff
@@ -12,48 +12,76 @@
...
+            ({"Records": [{"body": '{"type":"foo"}'}, {"body": '{"type":"bar"}'}]}),
+            ({"Records": [{"body": '{"type":"foo"}'}]}),
+            ({"Records": [{"foo": "bar"}]}),
+            ({"Records": []}),
+            ({}),
+        ],
+    )
     def test_handle_invalid_event(self, event):
         from handler import handle
+
```
manasaV3, line 85: Nit: Extra line?

data-workflows/tests/test_handler.py
```diff
@@ -12,48 +12,76 @@
...
         handle(event, None)
         self._verify(0)
```
manasaV3, line 87: If my memory serves me correctly, I remember there being additional unit tests for the new components that were added, those seem missing currently.

manasaV3 (changes requested):

data-workflows/handler.py
```diff
@@ -1,32 +1,46 @@
...
+        logger.info(f"Received message with body: {body}")
+        event = json.loads(body)
+        event_type = event.get("type", "").lower()
 
-        if event_type == 'activity':
+        if event_type == "activity":
             _update_activity()
+            logger.info(f"Update successful for type={event_type}")
+        elif event_type == "seed-s3-categories":
+            version = event.get("version")
```
klai95, line 42: since the parameter types for `version` and `s3_path` are `str`, what are your thoughts on having `version = event.get("version", "")` and `s3_path = event.get("s3_path", "")`?

data-workflows/run_workflow.py
```diff
@@ -0,0 +1,48 @@
...
+import argparse
+import handler
+import json
+
+from typing import Dict
+
+
+def run_workflow(event: Dict):
+    """
+    Function for running a particular data workflow based on the event type. The
```
klai95, line 14: It might be helpful to add some detail to the structure/values of the dict in the docstring. What are your thoughts on that?

data-workflows/handler.py
richaagarwal: @klai95 Since we require values for these parameters (see this line - https://github.com/chanzuckerberg/napari-hub/pull/959/files#diff-48bc4f460db59d31cd383f878f7ec7cb346afc564703a5025e6b180fe6070f93R62), we would not want to set empty defaults.

data-workflows/categories.py
richaagarwal: Neither of these need a default value. Setting a default value implies that we would sometimes expect missing data and, further, would not want to error out based on that. We control the data source in this case.

data-workflows/handler.py
```diff
@@ -1,32 +1,28 @@
...
+        LOGGER.info(f"Received message with body: {body}")
+        event = json.loads(body)
+        event_type = event.get("type", "").lower()
+
+        if event_type == "activity":
+            update_activity()
+            LOGGER.info(f"Update successful for type={event_type}")
+        elif event_type == "seed-s3-categories":
+            version = event.get("version")
+            categories_path = event.get("categories_path ")
```
manasaV3, line 25: There is an extra space after `categories_path`

manasaV3 (commented):
I believe this is missing unit tests. 😅

data-workflows/categories.py
```diff
@@ -0,0 +1,109 @@
...
+    source on the same depoyment stack.
+
+    :param version: The categories version
+    :type version: str
+    :param categories_path: The categories path in S3
+    :type categories_path: str
+    :raises ValueError: If params are not defined
+    """
+
+    if not all(version, categories_path):
```
manasaV3, line 68: `all` takes an iterable as parameter. So, we should make this `all([version, categories_path])`

manasaV3 (dismissed):
Functionality LGTM. But, might require unit test coverage.

data-workflows/activity/update_activity.py
```diff
@@ -0,0 +1,15 @@
+import activity.processor
+import utils.utils
+
+from utils.utils import ParameterStoreAdapter
+
+
+def update_activity() -> None:
```
manasaV3, line 7: Sorry, just noticed this. As we are moving this already, would it make sense for this to live in `activity.processor` file? That way we could expose only this method to the handler from that file.

data-workflows/categories.py
```diff
@@ -0,0 +1,109 @@
+import os
```
manasaV3, line 1: What do you think about moving this file into a category folder? I think it could help create consistency so all the workflows have their package.

manasaV3 (approved):
LGTM

Issue #958, opened 2023-03-21
Filter maintenance.timeline to return data from expected date range based on limit


klai95:
`Problem`: If we take https://api.napari-hub.org/metrics/zarpaint as an example, we can see that a portion of data returned by the API is outside of the range of 12 months, as `limit=12` by default. https://github.com/chanzuckerberg/napari-hub/pull/960 is the PR that addresses the bug by filtering out maintenance data that falls outside of the expected data range.

klai95:
> I'm not sure that `maintenance` data needs to be consistent with `usage` data. Could you clarify what the bug is?
> 
> Looking at the commented out tests, it looks like the test is expecting 12 `maintenance` entries for e.g. https://www.napari-hub.org/plugins/napari-assistant, but I think that is an issue with the test and not the functionality. There are 10 entries for it as expected since there was no data the past two months, and it renders correctly. I believe at some point we discussed that @codemonkey800 would fill in any missing months on his end, but might be worth clarifying that point.
> 
> cc @manasaV3

If we take https://api.napari-hub.org/metrics/zarpaint as an example, we can see that the first element of `maintenance.timeline` is `{
        "commits": 1,
        "timestamp": 1625097600000
      }`
      
The problem is that the timestamp is equivalent to `Thursday, July 1, 2021 12:00:00 AM` in GMT, and therefore falls out of the range of data we should return for the API endpoint. Therefore, the bug is that the API shouldn't return data that is out of the expected date range based on `limit`, and https://github.com/chanzuckerberg/napari-hub/pull/960 is the PR that addresses the bug.

Using the same case of https://api.napari-hub.org/metrics/zarpaint, the correct implementation would only return 6 valid items, assuming `limit=12` as our default value. I do agree with @richaagarwal that in this case it would be fine to just return those 6 items as is without padding the additional 6 items with default values, since @codemonkey800 fills them in already. 

However, I should bring up that @manasaV3's implementation for usage metrics, as seen in https://api.napari-hub.org/metrics/hesperos, fills in the default values, so that is the discrepancy between usage metrics (backend fills in default values )and maintenance metrics (frontend fills in default values), which is why in [this file](https://github.com/chanzuckerberg/napari-hub/blob/5c24b371f9f06fbfb199bed7e6535af8b76ea2e8/backend/bdd_tests/scenarios/metric.feature) the tests failed.

klai95:
[...]

PR #960, opened 2023-03-22
Filter maintenance.timeline to return data from expected date range based on limit

klai95:
Issue: https://github.com/chanzuckerberg/napari-hub/issues/958


backend/api/model.py
```diff
@@ -443,6 +442,19 @@ def _process_for_timeline(plugin_df, limit):
     return result
 
 
+def _process_maintenance_timeline(commit_activity, limit):
+    maintenance_timeline = []
+    end_date = date.today().replace(day=1) + relativedelta(months=-1)
+    start_date = end_date + relativedelta(months=-limit + 1)
+    end_date_str = str(end_date)
```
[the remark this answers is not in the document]
richaagarwal: I see.  But you are converting the timestamp to a `datetime` in line 452, so you could easily convert  it to a `date` instead, e.g. `date.fromtimestamp(item['timestamp'])`, and then compare all 3 dates directly.

I think the code would be much cleaner this way, without having to format the datetimes to strings and doing less conversions overall. What do you think?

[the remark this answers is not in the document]
klai95: I agree with this assessment! By doing so, the only conversion we need to do is on line 452, and then we can compare all three `date` objects on line 453 without having to purpose two lines for converting `date` objects to strings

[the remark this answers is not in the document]
manasaV3: It looks like the `item['timestamp']` is in milliseconds from line 452. 
Converting the start and end date to milliseconds as well would be more optimal here for two reasons. 
1. You no longer have to do the conversion on the `item['timestamp']`. You can do a direct comparison against the value.
2. The comparison implementation for a date time object is more expensive than a comparison for int.

backend/api/model.py
```diff
@@ -443,6 +442,19 @@ def _process_for_timeline(plugin_df, limit):
...
+    start_date_str = str(start_date)
+    for item in commit_activity:
+        item_date_str = datetime.fromtimestamp(item['timestamp'] / 1000).strftime("%Y-%m-%d")
+        if start_date_str <= item_date_str <= end_date_str:
+            maintenance_timeline.append(item)
+    return maintenance_timeline[-limit:]
```
manasaV3, line 455: We don't need the slicing here on the limit anymore as we only append the entries that are within the limit.

manasaV3 (changes requested):

klai95:
`(get_metrics_for_plugin('zarpaint', '3', False))`
`Response`
{'usage': {'timeline': [{'timestamp': 1672531200000, 'installs': 2}, {'timestamp': 1675209600000, 'installs': 8}, {'timestamp': 1677628800000, 'installs': 6}], 'stats': {'total_installs': 220, 'installs_in_last_30_days': 4}}, 'maintenance': {'timeline': [{'timestamp': 1672531200000, 'commits': 0}, {'timestamp': 1675209600000, 'commits': 0}, {'timestamp': 1677628800000, 'commits': 0}], 'stats': {'latest_commit_timestamp': 1668955796000, 'total_commits': 35}}}

`get_metrics_for_plugin('napari-bigfish', '3', False)`
`Response`
{'usage': {'timeline': [{'timestamp': 1672531200000, 'installs': 0}, {'timestamp': 1675209600000, 'installs': 0}, {'timestamp': 1677628800000, 'installs': 0}], 'stats': {'total_installs': 4, 'installs_in_last_30_days': 4}}, 'maintenance': {'timeline': [{'timestamp': 1672531200000, 'commits': 0}, {'timestamp': 1675209600000, 'commits': 0}, {'timestamp': 1677628800000, 'commits': 0}], 'stats': {'latest_commit_timestamp': None, 'total_commits': 0}}}

`get_metrics_for_plugin('zarpaint', '0', False)`
`Response`
{'usage': {'timeline': [], 'stats': {'total_installs': 220, 'installs_in_last_30_days': 4}}, 'maintenance': {'timeline': [], 'stats': {'latest_commit_timestamp': 1668955796000, 'total_commits': 35}}}

`get_metrics_for_plugin('zarpaint', 'foo', False)`
`Response`
{'usage': {'timeline': [], 'stats': {'total_installs': 220, 'installs_in_last_30_days': 4}}, 'maintenance': {'timeline': [], 'stats': {'latest_commit_timestamp': 1668955796000, 'total_commits': 35}}}

`get_metrics_for_plugin('zarpaint', '-5', False)`
`Response`
{'usage': {'timeline': [], 'stats': {'total_installs': 220, 'installs_in_last_30_days': 4}}, 'maintenance': {'timeline': [], 'stats': {'latest_commit_timestamp': 1668955796000, 'total_commits': 35}}}

backend/api/model.py
```diff
@@ -430,7 +429,7 @@ def _update_activity_timeline_data():
     write_data(csv_string, "activity_dashboard_data/plugin_installs.csv")
 
 
-def _process_for_timeline(plugin_df, limit):
+def _process_usage_timeline(plugin_df, limit):
```
richaagarwal, line 432: Great idea renaming this one!

backend/api/model.py
```diff
@@ -448,6 +447,18 @@ def _process_for_timeline(plugin_df, limit):
     return result
 
 
+def _process_maintenance_timeline(commit_activity, limit):
+    now = datetime.now()
+    end_date = datetime(now.year, now.month - 1, 1) if now.month > 1 else datetime(now.year - 1, 12, 1)
```
richaagarwal, line 452: I find the way `end_date` is calculated in line 434 a lot more readable. Any particular reason you didn't want to do it that way?

klai95: The original intention was to define `start_date` and `end_date` as type `datetime.datetime`, so that we could directly compare `start_date <= item_date <= end_date` without any conversions, since the implementation on line 434 and line 435 are type `datetime.date`.

That being said, after some thinking, I do agree the way that `end_date` is written here is a bit more complex than it needs to be, so I have decided to follow the logic in `_process_usage_timeline` to generate `start_date` and `end_date`.

I have also the `__process_for_dates` method that returns `start_date, end_date, dates`, which can be used by both `_process_usage_timeline` and `_process_maintenance_timeline` to avoid code duplication.

I have locally tested the current implementation and it works as expected, in the same way outlined in https://github.com/chanzuckerberg/napari-hub/pull/960#issuecomment-1506326020

backend/api/model.py
```diff
@@ -448,6 +452,15 @@ def _process_for_timeline(plugin_df, limit):
     return result
 
 
+def _process_maintenance_timeline(commit_activity, limit):
+    start_date, end_date, dates = _process_for_dates(limit)
+    maintenance_dict = {item_datetime: item for item in commit_activity if
```
richaagarwal, line 457: There's a lot going on here and it's hard to follow in a single line.  How about something similar to:
```
for commit in commit_activity:
    commit_date = ...
    if start_date <= commit_date <= end_date:
        add commit to list
```

richaagarwal (approved):

kne42 (commented):
I like the function name updates since it makes them more self-explanatory!

Issue #859, opened 2023-01-30
Update category API to use dynamoDB


manasaV3:
If the incoming request is in the test, fetch the data for the `categories/*` from dynamo, else fetch the results from s3.

##### `/categories`

- Scan the for records from categories table with filter `version=:version_name` and pagination to ensure that all the results have been fetched. 

```
aws dynamodb scan \
--table-name dev-test-categories \
--filter 'version=:version_name' \
--expression-attribute-values '{
":version_name": {"S": "EDAM-BIOIMAGING:alpha06"}
}'
```

In the above client call, replace table_name with the correct table name by stage and stack name.  
This query can be implemented in the backend using scan from boto3 linked [here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Client.scan).

[Link to pynamo resource](https://pynamodb.readthedocs.io/en/stable/batch.html?#scan-filters)  for the same.

-When constructing the response, use the `formatted_name` as the key in the JSON.
- Handle categories with multiple records by appending the second category data to existing entry. 

##### `/categories/<category>/versions/<version>`

- Convert the requested category_name to lowercase and the version_name defaults to `EDAM-BIOIMAGING:alpha06` if no version is provided in the request.
- Query for records from categories with key condition expression 'name=:category_name AND  begins_with(version_hash, version_name)’.

```
aws dynamodb query \
--table-name dev-test-categories \
--key-conditions 'name=:category_name AND begins_with(version_hash, :version_name)' \
--expression-attribute-values '{
":category_name": {"S": "foo"},
":version_name": {"S": "EDAM-BIOIMAGING:alpha06:"},
}'
```

In the above query, replace table_name with the correct table name by stage and stack name. Also, update the expression attribute value to have relevant values for all the fields. This query can be implemented in the backend using a query from boto3 linked [here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Client.query).

[Link to pynamo resource](https://pynamodb.readthedocs.io/en/stable/batch.html?#query-filters) for the same.

- Transform the response from the query into a list of results in the expected format as done in the /categories endpoint.

**Some of the data that could be useful to have logged:**
Time taken for dynamo scan
Time taken for dynamo query
Total number of results fetched

manasaV3:
Blocked by https://github.com/chanzuckerberg/napari-hub/issues/858

PR #962, opened 2023-03-24
Update categories API to use DynamoDB

codemonkey800:
## Description

https://github.com/chanzuckerberg/napari-hub/issues/859

Updates the following category endpoints to use Dynamo when the `use_dynamo_category` query parameter is `true`:

- `/categories`
- `/categories/<category>`
- `/categories/<category>/versions/<version>`

## `/categories`

image

## `/categories/<category>`

image

### DIFF
--- a/backend/utils/time.py
+++ b/backend/utils/time.py
@@ -1,4 +1,7 @@
 import time
+import logging
+
+LOGGER = logging.getLogger()
 
 
 def get_current_timestamp() -> int:
@@ -11,4 +14,4 @@ def get_perf_duration(start: float) -> float:
 
 def print_perf_duration(start: float, message: str):
     duration = get_perf_duration(start)
-    print(f"{message} duration={duration}ms")
+    LOGGER.info(f"{message} duration={duration}ms")
--- a/backend/api/app.py
+++ b/backend/api/app.py
@@ -112,7 +112,7 @@ def get_exclusion_list() -> Response:
 
 @app.route('/categories', defaults={'version': os.getenv('category_version', 'EDAM-BIOIMAGING:alpha06')})
 def get_categories(version: str) -> Response:
-    if request.args.get('use_dynamo_category', '').lower() == 'true':
+    if _is_query_param_true('use_dynamo_category'):
         return jsonify(CategoryModel.get_all_categories(version))
 
     return jsonify(get_categories_mapping(version))
@@ -121,7 +121,7 @@ def get_categories(version: str) -> Response:
 @app.route('/categories/<category>', defaults={'version': os.getenv('category_version', 'EDAM-BIOIMAGING:alpha06')})
 @app.route('/categories/<category>/versions/<version>')
 def get_category(category: str, version: str) -> Response:
-    if request.args.get('use_dynamo_category', '').lower() == 'true':
+    if _is_query_param_true('use_dynamo_category'):
         return jsonify(CategoryModel.get_category(category, version))
 
     return jsonify(get_category_mapping(category, get_categories_mapping(version)))


klai95 (dismissed):

backend/requirements.txt
```diff
@@ -17,3 +17,5 @@ build==0.8.0
 pyOpenSSL==22.0.0
 pandas==1.4.4
 snowflake-connector-python==2.8.0
+pynamodb==5.4.1
+python-slugify==8.0.1
```
manasaV3, line 21: 🔥

backend/api/models/category.py
```diff
@@ -0,0 +1,71 @@
...
+
+from slugify import slugify
+from collections import defaultdict
+from pynamodb.attributes import ListAttribute, NumberAttribute, UnicodeAttribute
+from pynamodb.models import Model
+from utils.time import get_current_timestamp, print_perf_duration
+from typing import Any, Dict
+
+
+def _get_category_from_model(category):
```
manasaV3, line 12: We could have this be a static method inside the class.

backend/api/models/category.py
```diff
@@ -0,0 +1,71 @@
...
+        start = time.perf_counter()
+        categories = cls.scan(
+            cls.version == version,
+            attributes_to_get=[
+                "formatted_name",
+                "version",
+                "dimension",
+                "hierarchy",
+                "label",
+            ],
```
manasaV3, line 59: As this will only exclude the last_updated_timestamp, curious about the impact on latency if we remove the projection list.

manasaV3 (commented):
Awesome stuff.. Could we add unit test coverage as well to these?

backend/api/models/category.py
```diff
@@ -0,0 +1,71 @@
...
+    version_hash = UnicodeAttribute(range_key=True)
+    version = UnicodeAttribute()
+    formatted_name = UnicodeAttribute()
+    dimension = UnicodeAttribute()
+    hierarchy = ListAttribute()  # List[str]
+    label = UnicodeAttribute()
+    last_updated_timestamp = NumberAttribute(default_for_new=get_current_timestamp)
+
+    @classmethod
+    def get_category(cls, name: str, version: str):
```
klai95, line 35: I think adding docstrings to methods would be helpful!

backend/api/models/category.py
```diff
@@ -0,0 +1,71 @@
...
+            slugify(name), cls.version_hash.startswith(f"{version}:")
+        ):
+            category.append(_get_category_from_model(item))
+
+        print_perf_duration(start, f"CategoryModel.get_category({name})")
+
+        return category
+
+    @classmethod
+    def get_all_categories(cls, version: str):
```
klai95, line 49: Adding docstrings to methods would be helpful!

backend/api/models/category.py
```diff
@@ -0,0 +1,71 @@
...
+                "dimension",
+                "hierarchy",
+                "label",
+            ],
+        )
+
+        mapped_categories = defaultdict(list)
+
+        for category in categories:
+            mapped_categories[category.formatted_name].append(
```
klai95, line 65: Would it make sense to have `mapped_categories[category.formatted_name].append(_get_category_from_model(category))` on one line?

richaagarwal: @klai95 PEP8 recommends lines be [shorter than 80 characters](https://peps.python.org/pep-0008/#maximum-line-length) and depending on which linters you're running it may automatically adjust for for you like this.  I know we don't enforce this within the hub but I do find the code more readable this way. Curious why you were suggesting having it on one line?

klai95: @richaagarwal That makes sense! I didn't actually realize the line is more than 80 characters just from viewing the PR on GitHub, but yeah in that case I do agree that code readability + maintaining each line to be shorter than 80 characters take priority!

klai95 (approved):

backend/utils/time.py
```diff
@@ -0,0 +1,14 @@
...
+    return round(time.time() * 1000)
+
+
+def get_perf_duration(start: float) -> float:
+    return (time.perf_counter() - start) * 1000
+
+
+def print_perf_duration(start: float, message: str):
+    duration = get_perf_duration(start)
+    print(f"{message} duration={duration}ms")
```
manasaV3, line 14: I like this method a lot. 🔥 

Nit: Can we please use a logger here if it has been initialized for this stack, if the logger still needs to be initialized, please ignore this.

manasaV3:
The test coverage on this is great!

klai95 (approved):
LGTM!

Issue #862, opened 2023-01-30
Update plugin workflow for fetching manifest data to write to dynamoDB


manasaV3:
We will write the distribution data fetched from manifest to dynamo DB in addition to s3. This will create the record of type `distribution` as referenced in [here](https://docs.google.com/document/d/1WOYxiFU7YVwugYeZmHUR5ynIz9QwRegAp9TCI5uH3CU/edit#heading=h.cvnc73au77qr).

- Update the plugin lambda to have write access permissions to the `plugins` dynamo DB table. 
- Use the [data](https://github.com/chanzuckerberg/napari-hub/blob/main/plugins/get_plugin_manifest.py#L36) to create the data field value.
-   Use the [plugin](https://github.com/chanzuckerberg/napari-hub/blob/main/plugins/get_plugin_manifest.py#L19) for creating the value for the name field
-  Use the [version](https://github.com/chanzuckerberg/napari-hub/blob/main/plugins/get_plugin_manifest.py#L20) to create the value of the version-type field with the formatting `f'{version}:distribution`.
- Use `round(time.time()*1000)` to create value for `last_updated_timestamp` field. 

Some of the data that could be useful to have logged:
- Time taken to generate the manifest data
- Time taken for dynamo writes
- Errors encountered when processing the manifest 
- Errors encountered when writing the data to s3/dynamo

PR #980, opened 2023-04-01
Cleaning plugin manifest fetch workflow

manasaV3:
## Description
Relates to https://github.com/chanzuckerberg/napari-hub/issues/862

- Breaks out the s3 components into a modular S3Adapter
- Sets up logging, updates prints to logging and adds timing logs
- Updates tests for s3 using moto


klai95 (approved):
👍

PR #982, opened 2023-04-03
Update to write Plugin manifest to dynamo and s3

manasaV3:
Relates to https://github.com/chanzuckerberg/napari-hub/issues/862

## Description
- In addition to s3, writes new manifest data to dynamo
- If data already exists in s3 ensures the data is also present in dynamo, if not, it copies over the data from s3 to dynamo

### DIFF
--- a/plugins/_tests/test_discovery.py
+++ b/plugins/_tests/test_discovery.py
@@ -1,15 +1,13 @@
 import json
 import time
 from unittest.mock import Mock
-
-import boto3
 import pytest
 from botocore.exceptions import ClientError
 from moto import mock_s3, mock_dynamodb
-import moto.dynamodb.urls
 
-TEST_BUCKET = 'test-bucket'
-TEST_BUCKET_PATH = 'test-path'
+from conftest import TEST_BUCKET_PATH, create_plugin_item, setup_dynamo, \
+    verify_plugin_item, setup_s3, put_s3_object
+
 TEST_PLUGIN = 'test-plugin'
 TEST_VERSION = '0.0.1'
 TEST_CACHE_PATH = f'cache/{TEST_PLUGIN}/{TEST_VERSION}-manifest.json'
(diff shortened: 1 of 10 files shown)


klai95 (dismissed):

plugins/models/plugin.py
```diff
@@ -0,0 +1,46 @@
...
+
+from pynamodb.models import Model
+from pynamodb.attributes import UnicodeAttribute, NumberAttribute, MapAttribute
+
+from utils.s3_adapter import S3Adapter
+
+LOGGER = logging.getLogger()
+
+
+def _get_version_type(version):
```
klai95, line 14: Correct me if I am wrong, but `_get_version_type` as a method name makes me think that this method will return different types of versions, but it seems like it's just returning a formatted string that combines the version and `:DISTRIBUTION`. What are your thoughts on renaming the method?

plugins/models/plugin.py
```diff
@@ -0,0 +1,46 @@
...
+    def write_manifest_data(plugin: str, version: str, data_str: str):
+        start = time.perf_counter()
+        data = json.loads(data_str)
+        version_type = _get_version_type(version)
+        Plugin(name=plugin, version_type=version_type, version=version, data=data).save()
+        duration = (time.perf_counter() - start) * 1000
+        logging.info(f'Put {plugin} {version_type} record time taken={duration}ms')
+
+    @staticmethod
+    def verify_exists_in_dynamo(plugin, version, path):
```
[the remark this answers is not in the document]
richaagarwal: @klai95 The method name combined with the parameters in this case should make that clear - we wouldn't want to hardcode the parameters into the method name itself.  Did you have something else in mind here?

plugins/models/plugin.py
richaagarwal: Agreed, something like `format_version` would be better.  @manasaV3 What is `type` supposed to be referencing here?

plugins/models/plugin.py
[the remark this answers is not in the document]
klai95: @richaagarwal That makes sense, and I agree! This is really minor, and probably doesn't require a change, but I thought having something like "verify_[noun]_exists_in_dynamo" made more sense at first glance. Since we definitely don't want to hardcode the parameters into the method name itself, I think it's fine though.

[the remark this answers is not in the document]
manasaV3: As this entire workflow is only for fetching plugin manifest. So, I felt that context might have been sufficient to know what we are verifying exists in Dynamo. 

But I am happy to update the name if you still feel the context does not provide a complete picture. :)

plugins/models/plugin.py
manasaV3: @richaagarwal  `version_type` references [this](https://github.com/chanzuckerberg/napari-hub/pull/982/files#diff-1b9d6a5fb4a57eb5bf238f8705fc2257952670d77145d30b35193ee985829b24R25) attribute.

@klai95 you are right in your description of what the method does. As defined in the schema, the type for the manifest record is `DISTRIBUTION`. As this specific workflow only deals with `DISTRIBUTION` we don't have an enum defined and are hardcoding the values.

plugins/_tests/test_discovery.py
```diff
@@ -1,99 +1,126 @@
...
-
-import boto3
 import pytest
 from botocore.exceptions import ClientError
-from moto import mock_s3
-import get_plugin_manifest
+from moto import mock_s3, mock_dynamodb
+
+from conftest import TEST_BUCKET_PATH, create_plugin_item, setup_dynamo, \
+    verify_plugin_item, setup_s3, put_s3_object
```
kne42, line 9:
```suggestion
from conftest import (
    TEST_BUCKET_PATH,
    create_plugin_item,
    put_s3_object,
    setup_dynamo,
    setup_s3,
    verify_plugin_item,
)
```

kne42: style request: use parentheses to break up multi-line statements instead of a backslash

kne42 (commented):
i lack proper context to give this a full review on how the testing/setup is done but overall LGTM. only other thing i might do is assign `'last_updated_timestamp'` to a constant in case you ever foresee that parameter's name getting changed but it's really not necessary

richaagarwal (approved):

PR #983, opened 2023-04-04
Adding logging for incoming requests

manasaV3:
## Description
Adding logging of incoming requests to help get metrics and support debugging.

### DIFF
--- a/backend/api/app.py
+++ b/backend/api/app.py
@@ -179,7 +179,7 @@ def authenticate_request():
 
 @app.after_request
 def add_header(response):
-    print('%s %s %s', request.method, request.full_path, response.status)
+    print(f'{request.method} {request.full_path} {response.status}')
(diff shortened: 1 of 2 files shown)


backend/api/app.py
```diff
@@ -179,6 +179,7 @@ def authenticate_request():
 
 @app.after_request
 def add_header(response):
+    print('%s %s %s', request.method, request.full_path, response.status)
```
richaagarwal, line 182: Are we continuing to use `print` statements or would it make sense to utilize the logger?

manasaV3: I added print for now to keep it consistent with the other statements here. The plan is to upgrade everything to logging in the future.

richaagarwal (approved):

richaagarwal (approved):

PR #989, opened 2023-04-10
No commit data shown on napari hub for hidden plugins 

klai95:
`Problem`: 
Currently, there are plugins we are not showing commit data even though we have Snowflake data for them. The reason for this behavior is as follows:
- Since we only have access to repo values in `latest_commit` and `commit_activity` SQL queries, we needed to created repo to plugin mappings so that `latest_commit.json` and `commit_activity.json` can store data with respect to plugins instead of repos. 
- However, such mappings are only available for public plugins, or in other words, plugins that exist in `index.json`, and this is the reason why hidden plugins (such as `napari-threedee`), which do not exist in `index.json`, did not have commit data.

`References`: Please check out https://www.napari-hub.org/plugins/napari-threedee and click on the `Activity` tab and view the Maintenance section, and see the commits at https://github.com/napari-threedee/napari-threedee.

### DIFF
--- a/backend/api/model.py
+++ b/backend/api/model.py
@@ -501,7 +501,8 @@ def _update_latest_commits(repo_to_plugin_dict):
             imaging.github.commits
         WHERE 
             repo_type = 'plugin'
-        GROUP BY repo
+        GROUP BY repo 
+        ORDER BY repo
     """
     hidden_plugins = get_hidden_plugins()
     cursor_list = _execute_query(query, "GITHUB")
@@ -530,6 +531,7 @@ def _update_commit_activity(repo_to_plugin_dict):
         WHERE 
             repo_type = 'plugin'
         GROUP BY repo, month
+        ORDER BY repo, month
     """
     hidden_plugins = get_hidden_plugins()
     cursor_list = _execute_query(query, "GITHUB")
--- a/backend/api/model.py
+++ b/backend/api/model.py
@@ -480,13 +480,26 @@ def _update_recent_activity_data(number_of_time_periods=30, time_granularity='DA
     write_data(json.dumps(data), "activity_dashboard_data/recent_installs.json")
 
 
+def _update_repo_to_plugin_dict(repo_to_plugin_dict: dict, plugin_obj: dict):
+    code_repository = plugin_obj.get('code_repository')
+    if code_repository:
+        repo_to_plugin_dict[code_repository.replace('https://github.com/', '')] = plugin_obj['name']
+    return repo_to_plugin_dict
+
+
 def _get_repo_to_plugin_dict():
     index_json = get_index()
+    hidden_plugins = get_hidden_plugins()
+    excluded_plugins = get_excluded_plugins()
     repo_to_plugin_dict = {}
     for public_plugin_obj in index_json:
-        code_repository = public_plugin_obj.get('code_repository')
-        if code_repository:
-            repo_to_plugin_dict[code_repository.replace('https://github.com/', '')] = public_plugin_obj['name']
+        repo_to_plugin_dict = _update_repo_to_plugin_dict(repo_to_plugin_dict, public_plugin_obj)
+    for excluded_plugin_name, excluded_plugin_visibility in excluded_plugins.items():
+        if excluded_plugin_visibility == "hidden":
+            excluded_plugin_obj = get_plugin(excluded_plugin_name, hidden_plugins[excluded_plugin_name])
+        else:
+            excluded_plugin_obj = get_plugin(excluded_plugin_name, None)
+        repo_to_plugin_dict = _update_repo_to_plugin_dict(repo_to_plugin_dict, excluded_plugin_obj)
     return repo_to_plugin_dict
 
 
@@ -504,18 +517,14 @@ def _update_latest_commits(repo_to_plugin_dict):
         GROUP BY repo 
         ORDER BY repo
     """
-    hidden_plugins = get_hidden_plugins()
     cursor_list = _execute_query(query, "GITHUB")
     data = {}
     for cursor in cursor_list:
         for row in cursor:
             repo = row[0]
-            repo_name = repo.split('/')[1]
             if repo in repo_to_plugin_dict:
                 plugin = repo_to_plugin_dict[repo]
-            elif repo_name in hidden_plugins:
-                plugin = repo_name
-            data[plugin] = int(pd.to_datetime(row[1]).strftime("%s")) * 1000
+                data[plugin] = int(pd.to_datetime(row[1]).strftime("%s")) * 1000
     write_data(json.dumps(data), "activity_dashboard_data/latest_commits.json")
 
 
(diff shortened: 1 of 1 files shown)


klai95:
`Snowflake latest_commit data` for `napari-threedee`
Screenshot 2023-04-09 at 10 03 44 PM
nt.com/70177777/230817147-3a3c822a-a1bc-40d0-8823-6d24dca159ae.png">

`Snowflake commit_activity data` for `napari-threedee`
Screenshot 2023-04-09 at 10 00 12 PM
`commit_activity.json` for `napari-threedee`
Screenshot 2023-04-09 at 10 01 07 PM

klai95:
Here is the current list of hidden plugins. [The current implementation](https://github.com/chanzuckerberg/napari-hub/pull/989/files#diff-0ab08ad5c775cdda3530d5690e3f32dc4b670e91f86849f4d17260fda27196bcR516) adds commit data for all hidden plugins except for `PoseR-napari`. The reason is that `PoseR-napari` is the only plugin in which the repo name is not an exact match to its corresponding plugin name, as shown below.
Screenshot 2023-04-09 at 10 16 12 PM
Screenshot 2023-04-09 at 10 21 09 PM

If we do want to include commit data for `PoseR-napari`, I see two potential approaches:
- we could do substring match, and since `PoserR` is a substring of `PoseR-napari`, we would be able to store its data. This approach is not very rigorous, as we may run into future cases where there are multiple substring matches, and therefore the approach may introduce false positives
- we could maintain a hard-coded dict that maps repos to plugins. If we were to go with this approach, `PoseR-napari` would be the only item in the dictionary currently, as all other plugins (public or hidden) would have commit data. This is also not an ideal solution.

What are your thoughts on this matter? @richaagarwal

backend/api/model.py
```diff
@@ -500,17 +501,21 @@ def _update_latest_commits(repo_to_plugin_dict):
...
     data = {}
     for cursor in cursor_list:
         for row in cursor:
             repo = row[0]
+            repo_name = repo.split('/')[1]
             if repo in repo_to_plugin_dict:
                 plugin = repo_to_plugin_dict[repo]
-                data[plugin] = int(pd.to_datetime(row[1]).strftime("%s")) * 1000
+            elif repo_name in hidden_plugins:
+                plugin = repo_name
```
manasaV3, line 517: I believe we went over why this would not work in [here](https://github.com/chanzuckerberg/napari-hub/pull/939#discussion_r1159419991)

manasaV3 (changes requested):

backend/api/model.py
klai95: That make sense to me. I have some confusion as to why we would use `get_excluded_plugins()` instead of `get_hidden_plugins()`. Take `napping` the plugin which has `disabled` visibility for example, it does not show up at all on https://www.napari-hub.org/plugins/napping. The same thing applies for https://www.napari-hub.org/plugins/napari-aphid, which is also a disabled plugin.

manasaV3: It is more so to improve coverage for future extensibility.  
For example, if a plugin went from disabled to visible during the day, we would already have data for it if it exists in the snowflake. 
But, having the data from the workflow running the subsequent day would not be too bad. So, while I would recommend using excluded plugins, we should also be fine with using just the hidden plugins.

klai95: Ok, I see what you're saying. I think using excluded plugins makes sense then. Thanks for the clarification

klai95: Just to verify, this means that, for the non-hidden plugins in `excluded_plugins.json`, we would call `get_plugin` but we don't pass in the version since we don't have access to that value right?

klai95:
Based on [Manasa's recommended approach](https://github.com/chanzuckerberg/napari-hub/pull/989#discussion_r1161943122), the PR now captures commit data for `PoseR-napari`, as well as any other plugins whose plugin names do not match their repo names. Furthermore, we are leveraging `get_excluded_plugins()` instead of `get_hidden_plugins()` to further increase the coverage. 


`latest_commit` for `PoseR-napari`
Screenshot 2023-04-10 at 4 11 40 PM
png">

`commit activity` for `PoseR-napari`
<img width="409" alt="Screenshot 2023-04-10 at 4 11 20 PM" src="https://user-images.githubusercontent.com/70177777/230999706-a67f95a9-007f-4d67-92e1-6efe453158a7.

backend/api/model.py
```diff
@@ -480,12 +480,26 @@ def _update_recent_activity_data(number_of_time_periods=30, time_granularity='DA
     write_data(json.dumps(data), "activity_dashboard_data/recent_installs.json")
 
 
+def _update_repo_to_plugin_dict(repo_to_plugin_dict: dict, plugin_obj: dict):
+    code_repository = plugin_obj.get('code_repository')
+    if code_repository:
+        repo_to_plugin_dict[code_repository.replace('https://github.com/', '')] = plugin_obj['name']
+    return repo_to_plugin_dict
```
manasaV3, line 487: Nit: We don't need to return it here as the caller method already has the dict available.

manasaV3 (approved):
:shipit:

backend/api/model.py
```diff
@@ -480,12 +480,26 @@ def _update_recent_activity_data(number_of_time_periods=30, time_granularity='DA
...
-    for plugin_obj in index_json:
-        if 'code_repository' in plugin_obj and plugin_obj['code_repository']:
-            repo_to_plugin_dict[plugin_obj['code_repository'].replace('https://github.com/', '')] = plugin_obj['name']
+    for public_plugin_obj in index_json:
+        repo_to_plugin_dict = _update_repo_to_plugin_dict(repo_to_plugin_dict, public_plugin_obj)
+    for excluded_plugin_name, excluded_plugin_visibility in excluded_plugins.items():
+        if excluded_plugin_visibility == "hidden":
+            excluded_plugin_obj = get_plugin(excluded_plugin_name, hidden_plugins[excluded_plugin_name])
+        else:
+            excluded_plugin_obj = get_plugin(excluded_plugin_name, None)
```
richaagarwal, line 501: I believe this line will fail to retrieve plugin metadata: if no version is passed, and it is an excluded plugin that is not hidden, it won't be returned from `get_valid_plugins()` that `get_plugin` calls in order to retrieve the latest version.  

Were you able to test this before merging @klai95 ?

richaagarwal: Thinking through this more, I don't think we'd have metadata for an excluded/non-hidden plugin in the first place, so we may not be able to include these in the activity fetch.

klai95: @richaagarwal Yeah, I did test this to ensure that the workflow produces the correct data and no error takes place. For the non-hidden plugins in `excluded_plugins.json`, excluded_plugin_obj is `{}`. `plugin_obj.get('code_repository')` in `_update_repo_to_plugin_dict` is `None`, and therefore those plugins are not added to repo_to_plugin_dict due to the `if code_repository` check in `_update_repo_to_plugin_dict`. As a result, `repo_to_plugin_dict` contains data for public and hidden plugins only. We can certainly only call `get_hidden_plugins()`, but the actual data in `repo_to_plugin_dict`  would be exactly the same and no errors are produced in either scenario.

klai95: If it makes sense, I could create a follow-up PR. As it stands, we wouldn't be able to increase coverage (https://github.com/chanzuckerberg/napari-hub/pull/989#discussion_r1161943122), since there is no data for the non-hidden plugins that are excluded.

richaagarwal: Okay, just wanted to flag this so we're all on the same page. This block of code is a little complicated and potentially misleading since we're not, in practice, getting additional metadata from the excluded list. However, since we're going to clean this all up soon with the migration effort, I think it's fine to leave as is for now.

klai95: Sounds good, I will keep this in mind. I will also make the necessary changes in the current PR I am working on: https://github.com/chanzuckerberg/napari-hub/pull/939

klai95: Just going to put the snippet of code here so we can use it when we refactor:
```
hidden_plugins = _get_cache('cache/hidden-plugins.json')
for hidden_plugin_name, hidden_plugin_version in hidden_plugins.items():
  hidden_plugin_obj = get_plugin(hidden_plugin_name, hidden_plugin_version)
  repo_to_plugin_dict = _update_repo_to_plugin_dict(repo_to_plugin_dict, hidden_plugin_obj)

Issue #993, opened 2023-04-11
Add total_installs to /plugins/index endpoint


manasaV3:
Blocks https://github.com/chanzuckerberg/napari-hub/issues/706

- Get all the plugin names in the plugins_metadata [here](https://github.com/chanzuckerberg/napari-hub/blob/main/backend/api/model.py#L262) 
- Get total installs for all plugins from the list
- Add the total_installs to relevant to plugins_metadata.values()

PR #994, opened 2023-04-11
Add total_installs to plugins/index cache

manasaV3:
## Description
Relates to https://github.com/chanzuckerberg/napari-hub/issues/993

TODO: 
- Needs unit tests

### DIFF
--- a/backend/api/model.py
+++ b/backend/api/model.py
@@ -225,13 +225,12 @@ def build_plugin_metadata(plugin: str, version: str) -> Tuple[str, dict]:
     return plugin, metadata
 
 
-def generate_index(plugins_metadata):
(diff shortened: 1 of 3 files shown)


klai95 (dismissed):

backend/api/model.py
```diff
@@ -226,6 +226,15 @@ def build_plugin_metadata(plugin: str, version: str) -> Tuple[str, dict]:
     return plugin, metadata
 
 
+def generate_index(plugins_metadata: Dict[str, Any]):
```
klai95, line 229: Can we add docstring to this method? I think it would be helpful in providing context in the future.

klai95 (approved):

Issue #923, opened 2023-03-06
Update metric API maintenance to use dynamoDB


manasaV3:
If the incoming request is in the test, fetch the data for `maintenance ` from dynamo else, fetch the results from s3.

### maintenance.timeline

- Query for records from `github-activity` with key condition expression `name=:plugin_name AND type_identifier BETWEEN :start_month AND :end_date` and projection of installs and timestamp field.
```
aws dynamodb query \
--table-name dev-test-github-activity \
--key-condition-expression 'name = :plugin_name AND type_identifier BETWEEN :start_date AND :end_month' \
--expression-attribute-values '{
":plugin_name": {"S": "foo"},
":start_month": {"S": "MONTH:202202"},
":end_month": {"S": "MONTH:202301"},
}'
--projection 'number_of_commits, timestamp'
```
In the above query, replace table_name with the correct table name by stage and stack name. Also, update the expression attribute value to have relevant values for all the fields. The `start_month` and `end_month` should be computed dynamically based on the current date.

- Then, transform the results from the dynamo to the expected format while adding records with a default value for installs for the missing months.

### maintenance.stats.latest_commit_timestamp

- Query for records from `github-activity` with key condition expression `name=:plugin_name AND type_identifier BEGINS_WITH :type` and projection of timestamp field.
```
aws dynamodb query \
--table-name dev-test-github-activity \
--key-condition-expression 'name = :plugin_name AND type_timestamp BEGINS_WITH :type' \
--expression-attribute-values '{
":plugin_name": {"S": "foo"},
":type": {"S": "LATEST:"}
}'
--projection 'timestamp'
```
In the above query, replace table_name with the correct table name by stage and stack name. Also, update the expression attribute value to have relevant values for the plugin_name field. 

- Then, transform the results from the dynamo to the expected format with a fallback to `None` if there are no records.

### maintenance.stats.total_commits

- Query for records from `github-activity` with key condition expression `name=:plugin_name AND type_identifier BEGINS_WITH :type` and projection of timestamp field.
```
aws dynamodb query \
--table-name dev-test-github-activity \
--key-condition-expression 'name = :plugin_name AND type_timestamp BEGINS_WITH :type' \
--expression-attribute-values '{
":plugin_name": {"S": "foo"},
":type": {"S": "TOTAL:"}
}'
--projection 'number_of_commits'
```
[...]

PR #996, opened 2023-04-13
Update metric API maintenance to use dynamoDB

klai95:
Issue: https://github.com/chanzuckerberg/napari-hub/issues/923
Reference: https://github.com/chanzuckerberg/napari-hub/pull/935/files

Changes introduced in this PR: on a high level, the work enables feature flagging for metric API maintenance. What this means is that users can pass the boolean variable `use_dynamo_metric_maintenance` to fetch maintenance data from dynamo if true else fetch from s3.

Additionally, tests are added to capture the new code changes, and some refactoring is also done to reduce tech debt


klai95:
Verified that data in DynamoDB matches with data returned by `get_total_commits`, `get_latest_commit`, and `get_maintenance_timeline` in class `GitHubActivity`

Screenshot 2023-04-13 at 7 10 15 PM

Screenshot 2023-04-13 at 7 11 06 PM

klai95:
Note: the tests for this PR are not yet complete, but the functionality part of the code is ready for review

backend/api/model.py
```diff
@@ -617,30 +624,53 @@ def _get_usage_data(plugin: str, limit: int, use_dynamo: bool) -> Dict[str, Any]
     return {'timeline': usage_timeline, 'stats': usage_stats, }
 
 
-def get_metrics_for_plugin(plugin: str, limit: str, use_dynamo_for_usage: bool) -> Dict[str, Any]:
+def _get_maintenance_data(plugin: str, repo: str, limit: int, use_dynamo: bool) -> Dict[str, Any]:
+    """
+    Fetches plugin maintenance_data from s3 or dynamo based on the in_test variable
```
manasaV3, line 629: Nit: use_dynamo variable

backend/api/model.py
```diff
@@ -617,30 +624,53 @@ def _get_usage_data(plugin: str, limit: int, use_dynamo: bool) -> Dict[str, Any]
...
     :return dict[str, Any]: A map with entries for usage and maintenance
 
     :params str plugin: Name of the plugin in lowercase for which usage data needs to be fetched.
     :params str limit_str: Number of records to be fetched for timeline. Defaults to 0 for invalid number.
     :params bool use_dynamo_for_usage: Fetch data from dynamo if True else fetch from s3. (default= False)
+    :params bool use_dynamo_for_maintenance: Fetch data from dynamo if True else fetch from s3. (default= False)
     """
     plugin = plugin.lower()
-    commit_activity = get_commit_activity(plugin)
+    repo = _get_repo_from_plugin(plugin)
```
manasaV3, line 666: Is the value stored in repo used when `use_dynamo_for_maintenance=False`?

backend/api/model.py
```diff
@@ -617,30 +624,58 @@ def _get_usage_data(plugin: str, limit: int, use_dynamo: bool) -> Dict[str, Any]
...
+    """
+    if use_dynamo_for_maintenance:
+        maintenance_timeline = []
+        maintenance_total_commits = 0
+        maintenance_latest_commit_timestamp = 0
+        if limit:
+            repo = _get_repo_from_plugin(plugin)
+            maintenance_timeline = github_activity.get_maintenance_timeline(plugin, repo, limit)
+            maintenance_total_commits = github_activity.get_total_commits(plugin, repo)
+            maintenance_latest_commit_timestamp = github_activity.get_latest_commit(plugin, repo)
```
manasaV3, line 643: We would get the values for these even if the limit is invalid.

backend/api/model.py
```diff
@@ -617,30 +623,52 @@ def _get_usage_data(plugin: str, limit: int, use_dynamo: bool) -> Dict[str, Any]
...
+def get_metrics_for_plugin(plugin: str, limit: str, use_dynamo_for_usage: bool,
+                           use_dynamo_for_maintenance: bool) -> Dict[str, Any]:
     """
     Fetches plugin metrics from s3 or dynamo based on the in_test variable
     :return dict[str, Any]: A map with entries for usage and maintenance
 
     :params str plugin: Name of the plugin in lowercase for which usage data needs to be fetched.
     :params str limit_str: Number of records to be fetched for timeline. Defaults to 0 for invalid number.
     :params bool use_dynamo_for_usage: Fetch data from dynamo if True else fetch from s3. (default= False)
+    :params bool use_dynamo_for_maintenance: Fetch data from dynamo if True else fetch from s3. (default= False)
```
manasaV3, line 662: Nit: We can do away with the default value, as we are not specifying default here. Also, as we have two different booleans now, we should add more clarity like `Fetch install data from dynamo`

manasaV3 (commented):
- Have we deployed this to dev env and tested the response against the current response with [script](https://gist.github.com/manasaV3/1114fa53f0916e8bc31d6afa66fe9104) after modifying to test for `use_dynamo_metric_maintenance`? 
- Can we add unit test for github activity?

manasaV3 (commented):
- Have we deployed this to dev env and tested the response against the current response with [script](https://gist.github.com/manasaV3/1114fa53f0916e8bc31d6afa66fe9104) after modifying to test for `use_dynamo_metric_maintenance`? 
- Can we add unit test for github activity?

backend/api/model.py
klai95: Can you clarify what you mean by doing away with the default value here?

backend/api/s3.py
```diff
@@ -113,7 +113,7 @@ def get_recent_activity_data() -> Dict:
 
 
 def get_latest_commit(plugin: str) -> Any:
-    return _load_json_from_s3("activity_dashboard_data/latest_commits.json").get(plugin)
+    return _load_json_from_s3("activity_dashboard_data/latest_commits.json").get(plugin, 0)
```
richaagarwal, line 116: What is this change for?

backend/bdd_tests/scenarios/metric.feature
```diff
@@ -7,6 +7,7 @@ Feature: metrics
     And it should have 12 entries for usage.timeline
     And it should have at least one non-zero installs in usage.timeline
     And it should have non-zero values for usage.stats
+    And it should have 12 entries for maintenance.timeline
```
richaagarwal, line 10: All of these additional assertions need to have corresponding tests added to `test_metric.py`.

backend/api/s3.py
klai95: I just synced with Jeremy on what the frontend expects of this value in the case where the data is not available, and the frontend expects either Null or 0. However, I think keeping it `return _load_json_from_s3("activity_dashboard_data/latest_commits.json").get(plugin)` makes more sense. I have also updated the tests and `github_activity.get_latest_commit` so that the return values for when the data is not available are consistent across s3 and dynamo implementations.

backend/bdd_tests/scenarios/metric.feature
klai95: I have verified that all the additional assertions have the corresponding tests. 

For example, `And it should have 12 entries for maintenance.timeline` has the corresponding test
```
@then(parsers.parse('it should have {limit:d} entries for maintenance.timeline'))
def verify_maintenance_timeline_limit(context, limit):
    _validate_timeline(context['maintenance']['timeline'], limit)

**Review by manasaV3 — COMMENTED** [review]
Not a blocker, but more of a thought on optimization. 

Do we know how many plugins might have multiple repos? I believe the number might not be large. But, we are fetching the plugin data and getting the repo from it for all the cases. We could reduce that to only cases where there are multiple repo entries. 

For example, take the `github_activity.get_total_commits(plugin, repo)` it returns only the total_commit value right now after doing a get_item call on the table. Instead, we could query the table for the `hash_key=plugin_name and the range_key=begins_with("LATEST:")`.  

That could look like the following untested code: 
```
_GitHubActivityModel.get(
    plugin_name,
    _GitHubActivityModel.type_identifier.begins_with('LATEST:'))
)
```
From the results, we could construct a map such that: `{repo : commit_count}`.  

If total_commit, latest_timestamp or timeline have more than one repo entry key, we could do the additional fetch needed for getting the current repo for a plugin. Or else, we can verify that all of them have the same repo key and use that for building the `maintenance` object.

File: backend/api/models/github_activity.py [hunk 7]
```diff
@@ -0,0 +1,87 @@
...
+def get_maintenance_timeline(plugin: str, repo: str, month_delta: int) -> List[Dict[str, int]]:
+    """
+    Fetches plugin commit count at a month level granularity from dynamo in the previous month_delta months.
+    :returns List[Dict[str, int]]: Entries for the month_delta months
+
+    :param str plugin: Name of the plugin in lowercase for which maintenance timeline data needs to be fetched.
+    :param str repo: Name of the GitHub repo.
+    :param int month_delta: Number of months in maintenance timeline.
+    """
+    month_type_format = 'MONTH:{0:%Y%m}:{1}'
```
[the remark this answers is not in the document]
klai95: I think using the variable names here is a good call out. I am curious about your recommendation to make it a module constant. Given that `get_maintenance_timeline` is the only place that this variable is being used, would love to know your rationale for making it a module constant.

backend/api/models/github_activity.py
```diff
@@ -0,0 +1,87 @@
...
+    start_date = datetime.datetime.now().replace(day=1) - relativedelta(months=1)
+    end_date = start_date - relativedelta(months=month_delta - 1)
+    condition = _GitHubActivityModel.type_identifier.between(month_type_format.format(end_date, repo),
+                                                             month_type_format.format(start_date, repo))
```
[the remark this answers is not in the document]
klai95: The method signature is `between(self, lower: Any, upper: Any)`

Give that, what are your thoughts on naming them 'lower or `lower_bound` for `end_date` and `upper` or _upper_bound` for `start_date`?

backend/api/app.py
```diff
@@ -142,11 +142,14 @@ def get_plugin_metrics(plugin: str) -> Response:
...
     :query_params limit: Number of months to be fetched for timeline. (default=12).
     :query_params use_dynamo_metric_usage: Fetch usage data from dynamo if True else fetch from s3. (default=False)
+    :query_params use_dynamo_metric_maintenance: Fetch maintenance data from dynamo if True else fetch from s3.
+                  (default=False)
     """
     return jsonify(get_metrics_for_plugin(
         plugin=plugin,
         limit=request.args.get('limit', '12'),
         use_dynamo_for_usage=_is_query_param_true('use_dynamo_metric_usage'),
+        use_dynamo_for_maintenance=_is_query_param_true('use_dynamo_metric_maintenance'),
```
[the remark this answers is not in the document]
klai95: I believe the spelling is consistent as is here

backend/api/models/_tests/test_github_activity.py
```diff
@@ -0,0 +1,108 @@
...
+        ])
+    def test_get_latest_commit(self, github_activity_table, data, expected):
+        start = datetime.date.today()
+        for count, period in data:
+            timestamp = pd.Timestamp(start - relativedelta(days=period))
+            self._put_item(github_activity_table, 'LATEST', timestamp, count)
+
+        actual = github_activity.get_latest_commit(PLUGIN_NAME, REPO_NAME)
+
+        assert actual == expected
```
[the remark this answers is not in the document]
klai95: I have updated the test; the expected value is directly defined in `pytest.mark.parametrize`

backend/api/models/github_activity.py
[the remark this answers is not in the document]
manasaV3: Yeah, that makes it clearer.

backend/api/models/github_activity.py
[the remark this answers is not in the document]
manasaV3: I see the reasoning behind wanting to keep it closer to where it gets used, but with how it is currently set up, it doesn't immediately convey that it is a constant. Having the variable name in all caps usually helps with that.

backend/api/models/github_activity.py
```diff
@@ -0,0 +1,86 @@
...
+    Fetches plugin commit count at a month level granularity from dynamo in the previous month_delta months.
+    :returns List[Dict[str, int]]: Entries for the month_delta months
+
+    :param str plugin: Name of the plugin in lowercase for which maintenance timeline data needs to be fetched.
+    :param str repo: Name of the GitHub repo.
+    :param int month_delta: Number of months in maintenance timeline.
+    """
+    upper = datetime.datetime.now().replace(day=1) - relativedelta(months=1)
+    lower = upper - relativedelta(months=month_delta - 1)
+    condition = _GitHubActivityModel.type_identifier.between('MONTH:{lower:%Y%m}:{repo}'.format(lower=lower, repo=repo),
```
manasaV3, line 76: Nit: I think my earlier feedback about constant might have been misunderstood. It was good that we were using a single constant string to format both upper and lower, but the suggestion was to make it more evident that it was a constant. 

For example, if we still wanted to keep it within the function, it could look like:
```
MONTH_TYPE_IDENTIFIER = f'MONTH:{{timestamp:%Y%m}}:{repo}'

upper = datetime.datetime.now().replace(day=1) - relativedelta(months=1)

lower_str = MONTH_TYPE_IDENTIFIER.format(
   timestamp=upper - relativedelta(months=month_delta - 1)
)
condition = _GitHubActivityModel.type_identifier.between(
lower_str, MONTH_TYPE_IDENTIFIER.format(timestamp=upper)
)
```

Or if we wanted to make it a module constant, it could look like this:

```
MONTH_TYPE_IDENTIFIER = 'MONTH:{timestamp:%Y%m}:{repo}'
```

backend/api/models/github_activity.py
```diff
@@ -0,0 +1,86 @@
...
+    plugin_name = UnicodeAttribute(hash_key=True)
+    type_identifier = UnicodeAttribute(range_key=True)
+    granularity = UnicodeAttribute(attr_name='type')
+    timestamp = NumberAttribute(null=True)
+    commit_count = NumberAttribute(null=True)
+    repo = UnicodeAttribute()
+    last_updated_timestamp = NumberAttribute()
+
+
+def get_total_commits(plugin: str, repo: str) -> int:
```
manasaV3, line 29: The doc string and logging in this module are great. 🚀

backend/api/models/github_activity.py
```diff
@@ -0,0 +1,86 @@
...
+    :param str repo: Name of the GitHub repo.
+    """
+    start = time.perf_counter()
+    try:
+        return _GitHubActivityModel.get(plugin, f'TOTAL:{repo}').commit_count
+    except _GitHubActivityModel.DoesNotExist:
+        logging.warning(f'No TOTAL:{repo} record found for plugin={plugin}')
+        return 0
+    finally:
+        logging.info(f'get_total_commits for plugin={plugin} time_taken={(time.perf_counter() - start) * 1000}ms')
```
manasaV3, line 44: Can we include the repo in all the log statements in this module, where we are already including `plugin`?

backend/api/models/_tests/test_github_activity.py
```diff
@@ -0,0 +1,98 @@
...
+
+    @pytest.mark.parametrize(
+        'data, expected', [
+            ([], 0),
+            ([('bar', 'repo/bar', 100)], 0),
+            ([(PLUGIN_NAME, 'repo/foo2', 100)], 0),
+            ([('bar', REPO_NAME, 100)], 0),
+            ([(PLUGIN_NAME, REPO_NAME, 100)], 100),
+        ])
```
manasaV3, line 91: 🔥

backend/api/_tests/test_fixtures.py
```diff
@@ -10,7 +10,7 @@ def _to_timestamp(i):
 
 
 def _generate_timeline(start_range, value_key, to_value, timestamp_key='timestamp'):
-    return [{timestamp_key: _to_timestamp(i), value_key: to_value(i)} for i in range(start_range, 0)]
+    return [{timestamp_key: _to_timestamp(i), value_key: to_value(-i)} for i in range(start_range, 0)]
```
manasaV3, line 13: I think negating the `i` value here could lead to unexpected behaviour. What do you think about having an absolute value instead?

backend/api/models/github_activity.py
klai95: Is there a reason why we can't do the following for the `lower` part so we are keeping the types of `lower` and `upper` consistent as well as how they're being applied in the condition? 

```
lower = upper - relativedelta(months=month_delta - 1)
MONTH_TYPE_IDENTIFIER.format(timestamp=lower)

File: backend/api/_tests/test_fixtures.py (hunk 15 above)
Comment (klai95) [reply backend/api/_tests/test_fixtures.py]: Resolved!

File: backend/api/models/github_activity.py (hunk 13 above)
Comment (klai95) [reply backend/api/models/github_activity.py]: Resolved

**Comment by klai95** [comment]
Screenshot 2023-05-17 at 3 27 14 PM

After turning on the feature flag for the maintenance metric in split.io and deploying the current code on `dev-shared`, I am able to verify that the endpoints are working as expected by hitting the endpoint: https://api.dev.napari-hub.org/dev-shared/metrics/btrack?use_dynamo_metric_maintenance=True
https://api.dev.napari-hub.org/dev-shared/metrics/btrack?use_dynamo_metric_maintenance=True

**Comment by richaagarwal** [comment]
> After turning on the feature flag for the maintenance metric in split.io and deploying the current code on dev-shared, I am able to verify that the endpoints are working as expected by hitting the endpoints for a couple plugins. For example,
https://api.dev.napari-hub.org/dev-shared/metrics/btrack?use_dynamo_metric_maintenance=True

@klai95 Please check the frontend dev URL as well and confirm the plugin page loads maintenance data as expected

File: backend/api/models/github_activity.py (hunk 11 above)
Comment (manasaV3) [reply backend/api/models/github_activity.py]: I only did that because we don't reference the lower variable anywhere else in the code.

File: backend/api/models/github_activity.py [hunk 16]
```diff
@@ -0,0 +1,87 @@
...
+    :param str repo: Name of the GitHub repo.
+    """
+    start = time.perf_counter()
+    try:
+        return _GitHubActivityModel.get(plugin, f'LATEST:{repo}').timestamp
+    except _GitHubActivityModel.DoesNotExist:
+        logging.warning(f'No LATEST{repo}: record found for plugin={plugin}')
+        return None
+    finally:
+        logging.info(f'get_latest_commit for plugin={plugin} time_taken={(time.perf_counter() - start) * 1000}ms')
```
manasaV3, line 62: Can we include the `repo` in all the log statements in this module, where we are already including plugin?

backend/api/models/github_activity.py
```diff
@@ -0,0 +1,87 @@
...
+    MONTH_TYPE_IDENTIFIER = f'MONTH:{{timestamp:%Y%m}}:{repo}'
+    upper = datetime.datetime.now().replace(day=1) - relativedelta(months=1)
+    lower = upper - relativedelta(months=month_delta - 1)
+    condition = _GitHubActivityModel.type_identifier.between(MONTH_TYPE_IDENTIFIER.format(timestamp=lower, repo=repo),
+                                                             MONTH_TYPE_IDENTIFIER.format(timestamp=upper, repo=repo))
+
+    start = time.perf_counter()
+    results = {row.timestamp: row.commit_count for row in _GitHubActivityModel.query(plugin, condition)}
+    duration = (time.perf_counter() - start) * 1000
+    logging.info(f'Query for plugin={plugin} month_delta={month_delta} time_taken={duration}ms')
```
manasaV3, line 83: Can we include the `repo` in all the log statements in this module, where we are already including plugin?

backend/api/model.py
```diff
@@ -617,30 +625,51 @@ def _get_usage_data(plugin: str, limit: int, use_dynamo: bool) -> Dict[str, Any]
...
+        repo = _get_repo_from_plugin(plugin)
+        plugin = plugin.lower()
+        maintenance_timeline = github_activity.get_maintenance_timeline(plugin, repo, limit) if limit else []
+
+        maintenance_stats = {
+            'total_commits': github_activity.get_total_commits(plugin, repo),
+            'latest_commit_timestamp': github_activity.get_latest_commit(plugin, repo),
+        }
+    else:
+        data = get_commit_activity(plugin.lower())
```
manasaV3, line 646: We should also do `plugin = plugin.lower()` here. As it gets used [here](https://github.com/chanzuckerberg/napari-hub/pull/996/files#diff-0ab08ad5c775cdda3530d5690e3f32dc4b670e91f86849f4d17260fda27196bcR649)

backend/api/models/_tests/test_github_activity.py
```diff
@@ -0,0 +1,98 @@
...
+class TestGitHubActivity:
+
+    @pytest.fixture()
+    def github_activity_table(self, aws_credentials):
+        with mock_dynamodb():
+            yield create_dynamo_table(github_activity._GitHubActivityModel, 'github-activity')
+
+    @classmethod
+    def _to_type_identifier(cls, granularity, repo, timestamp):
+        if granularity == 'LATEST':
```
kne42, line 37: i thought we had an enum to capture this stuff? ignore if im wrong

backend/api/models/_tests/test_github_activity.py
```diff
@@ -0,0 +1,98 @@
...
+            return f'LATEST:{repo}'
+        elif granularity == 'MONTH':
+            return f'MONTH:{timestamp.strftime("%Y%m")}:{repo}'
```
kne42, line 40: seems related to below, may be able to capture as a constant somewhere

backend/api/models/github_activity.py
```diff
@@ -0,0 +1,87 @@
...
+def get_maintenance_timeline(plugin: str, repo: str, month_delta: int) -> List[Dict[str, int]]:
+    """
+    Fetches plugin commit count at a month level granularity from dynamo in the previous month_delta months.
+    :returns List[Dict[str, int]]: Entries for the month_delta months
+
+    :param str plugin: Name of the plugin in lowercase for which maintenance timeline data needs to be fetched.
+    :param str repo: Name of the GitHub repo.
+    :param int month_delta: Number of months in maintenance timeline.
+    """
+    MONTH_TYPE_IDENTIFIER = f'MONTH:{{timestamp:%Y%m}}:{repo}'
```
kne42, line 74: seems related to above; may be able to capture as a constant somewhere

kne42 (commented):
small nits

backend/api/models/_tests/test_github_activity.py
```diff
@@ -0,0 +1,98 @@
...
+        return f'TOTAL:{repo}'
+
+    def _put_item(self, table, granularity, timestamp, commit_count, plugin=PLUGIN_NAME, repo=REPO_NAME):
+        item = {
+            'plugin_name': plugin,
+            'type_identifier': self._to_type_identifier(granularity, repo, timestamp),
+            'commit_count': commit_count,
+            'granularity': granularity,
```
manasaV3, line 48: Nit: This should be `'type': granularity`, as the actual attribute's name is `type`.

manasaV3 (dismissed):

manasaV3 (approved):
:shipit:

PR #1014, opened 2023-05-01
Fixing use of timestamp keyword in snowflake query

manasaV3:
## Description

Changes introduced in this PR:

- Fixes failure in the activity data workflow because the keyword `timestamp` is used as an alias in the projections. 
- Removes unused import
- Updates `as` -> `AS` in query for consistency

### DIFF
--- a/data-workflows/activity/github_activity_model.py
+++ b/data-workflows/activity/github_activity_model.py
@@ -9,7 +9,7 @@ from pynamodb.models import Model
 from pynamodb.attributes import UnicodeAttribute, NumberAttribute
 
 from utils.utils import get_current_timestamp, date_to_utc_timestamp_in_millis, datetime_to_utc_timestamp_in_millis
-from plugin.helpers import _get_cache, _get_repo_to_plugin_dict
+from plugin.helpers import _get_repo_to_plugin_dict
 
 
 LOGGER = logging.getLogger()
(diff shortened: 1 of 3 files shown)


data-workflows/activity/github_activity_model.py
```diff
@@ -27,11 +27,11 @@ def __new__(cls, timestamp_formatter, type_identifier_formatter, query_projectio
         return github_activity_type
 
     LATEST = (datetime_to_utc_timestamp_in_millis, 'LATEST:{0}',
-              'repo AS name, to_timestamp(max(commit_author_date)) as latest_commit', 'name')
+              'repo AS name, to_timestamp(max(commit_author_date)) AS latest_commit', 'name')
```
klai95, line 30: nit: we should change `max` to `MAX` for consistency

data-workflows/activity/github_activity_model.py
```diff
@@ -27,11 +27,11 @@ def __new__(cls, timestamp_formatter, type_identifier_formatter, query_projectio
...
     MONTH = (date_to_utc_timestamp_in_millis, 'MONTH:{1:%Y%m}:{0}',
-             'repo AS name, date_trunc("month", to_date(commit_author_date)) as month, count(*) as commit_count',
+             'repo AS name, date_trunc("month", to_date(commit_author_date)) AS month, count(*) AS commit_count',
```
klai95, line 32: nit: `TO_DATE`

data-workflows/activity/github_activity_model.py
```diff
@@ -27,11 +27,11 @@ def __new__(cls, timestamp_formatter, type_identifier_formatter, query_projectio
...
              'name, month')
-    TOTAL = (lambda timestamp: None, 'TOTAL:{0}', 'repo AS name, count(*) as commit_count', 'name')
+    TOTAL = (lambda timestamp: None, 'TOTAL:{0}', 'repo AS name, count(*) AS commit_count', 'name')
```
klai95, line 34: nit: `COUNT`

data-workflows/activity/github_activity_model.py
manasaV3: I agree, but we need to do a once-over to identify all of the keywords that are in lowercase and update it.

klai95 (approved):

PR #1028, opened 2023-05-04
Updating InstallActivity unit tests to use moto

manasaV3:
## Description

Changes introduced in this PR:
- Restructuring the InstallActivity class to reduce exposure of the dynamo tables outside the install_activity file.
- Updating unit tests for install-activity table access to use moto instead of pytest mock
- Update the categories model to use the correct env variable for prefix to the table name.


backend/api/models/_tests/test_install_activity.py
```diff
@@ -0,0 +1,107 @@
...
+
+def to_timestamp(start, i):
+    return to_millis(pd.Timestamp(start + relativedelta(months=i)))
+
+
+def to_installs(i):
+    return 2 if i % 2 == 0 else 0
+
+
+class TestInstallActivity:
```
klai95, line 27: I think it'd be good to have docstring for these tests, in order to provided context on (1) what the class is trying to accomplish and (2) why the class is structured in this particular way and whether there were alternatives to the existing implementation

backend/api/models/_tests/test_install_activity.py
```diff
@@ -0,0 +1,107 @@
...
+        self._put_item(install_activity_table, 'TOTAL', None, expected)
+
+        actual = install_activity.get_total_installs(plugin=PLUGIN_NAME)
+
+        assert actual == expected
+
+    @pytest.mark.parametrize(
+        'results, expected', [
+            ([(100, 30)], 0),
+            ([(10, 5), (24, 12), (19, 10), (100, 30)], 53),
```
klai95, line 68: Can we declare what's being passed in on line 67 and line 68 as variables at the top?

backend/api/models/_tests/test_install_activity.py
```diff
@@ -0,0 +1,107 @@
...
+        actual = install_activity.get_recent_installs(plugin=PLUGIN_NAME, day_delta=15)
+
+        assert actual == expected
+
+    @pytest.mark.parametrize('results, month_delta, expected', [
+        ([], 0, []),
+        ([], 1, generate_installs_timeline(-1, to_value=lambda i: 0)),
+        ([(to_installs(i), i) for i in range(0, 7, 2)], 4, generate_installs_timeline(-4, to_value=to_installs)),
+    ])
+    def test_get_timeline(self, install_activity_table, results, month_delta, expected):
```
klai95, line 85: Could we rename the method to `test_get_usage_timeline`?

backend/api/models/helper.py
```diff
@@ -0,0 +1,15 @@
+import functools
+import os
+from typing import Union
+
+from pynamodb.models import Model
+
+
+def set_ddb_metadata(table_name: str, dynamo_model_cls: Model = None) -> Union[Model, functools.partial]:
```
klai95, line 8: Docstring for this method would be helpful

backend/api/models/_tests/test_install_activity.py
manasaV3: The method it is testing is called `get_timeline`, so I think it makes sense to leave it as it currently is. 

The  `get_timeline` method, will be referenced as `install_activity.get_timeline(...)`, that is why I left it as get_timeline instead of adding "usage" to the method name.

backend/api/models/_tests/test_install_activity.py
manasaV3: > (1) what the class is trying to accomplish? 

I thought the class name, the file name, and test name provided clarity on that. 

> (2) why the class is structured in this particular way and whether there were alternatives to the existing implementation

Is there something specific about the structure that you are looking explanation for? 😅

backend/api/models/_tests/test_install_activity.py
manasaV3: I believe it is already declared in line 66. 

But, if you are looking for the `results`, looking [here](https://github.com/chanzuckerberg/napari-hub/pull/1028/files#diff-233b695542ad3b7eb61dc3c2228617f9812c10159e4897638f4c0ba47d997230R72) would provide more clarity. 

I think renaming `results` -> `data` might also help make that more intuitive.

backend/api/models/helper.py
manasaV3: Also, for more context, I added this so all pynamo models are initialized consistently.

klai95 (approved):

PR #1030, opened 2023-05-05
Fix TOC position

codemonkey800:
## Description

Fixes the position of the TOC after the install button was removed in #1016 

### Before

image

### After

image


manasaV3 (approved):
:shipit:

Issue #995, opened 2023-04-13
Workflow step filter menu categories duplicated


Janeece:
#### Description
In workflow step filter on the plugin list, when i scroll through the filter menu, I noticed that many of the subsections are repeating as separate sections instead of consolidating. 

#### Steps/Code to Reproduce
Open workflow step menu

#### Expected Results
Only one section of each time with multiple options

#### Actual Results
<img width="268" alt="Screenshot 2023-03-30 at 1 04 50 PM" src="https://user-images.githubusercontent.com/63013360/231883342-b981ef40-de90-4162-9f55-31bc261eddd8.png">

Janeece:
cc @codemonkey800

PR #1034, opened 2023-05-08
Fix workflow step header duplication

codemonkey800:
## Description

https://github.com/chanzuckerberg/napari-hub/issues/995

- Fixes workflow step group duplication by sorting options by group
- Issue was that options weren't sorted by group, [which is a requirement for the feature to work](https://mui.com/material-ui/react-autocomplete/#grouped)

### Before

image

### After

image


manasaV3 (approved):
:shipit:

Issue #1009, opened 2023-05-01
Add `/plugins` page route


codemonkey800:
A new route `/plugins` needs to be added to the frontend to serve the new plugins page in the home page redesign. This will depend on the `homePageRedesign` feature flag and should have the following behavior when the feature flag is on:

- `/plugins` visible in `/sitemap.xml`
- Plugins section links to `/plugins` on `/sitemap` page
- `/plugins` renders a basic page for now until implementation

PR #1039, opened 2023-05-09
Plugin page route

codemonkey800:
## Description

https://github.com/chanzuckerberg/napari-hub/issues/1009

- Extracts not found page to separate component
- Adds `homePageRedesign` feature flag
- Adds `/plugins` page route that shows 404 when flag is off and plugin page when flag is on

### Not found page when flag is off

image

### Plugins page when flag is on

image


klai95 (approved):
LGTM!

PR #1047, opened 2023-05-12
Fix sitemap for plugin page

codemonkey800:
## Description

This fixes the sitemap.xml and sitemap page so that they display the `/plugins` route correctly after merging #1039. This includes:

1. Showing `/plugins` only when the feature flag is on
2. Change `Plugins` link to `/plugins` on the sitemap page when the feature flag is on


manasaV3 (approved):
:shipit:

klai95 (approved):

PR #1049, opened 2023-05-12
Seed categories script unit tests

codemonkey800:
## Description

- Adds unit tests for #959
- Refactors unit tests from #962 to use moto


data-workflows/categories/category_model.py
```diff
@@ -24,3 +24,13 @@ class Meta:
     hierarchy = ListAttribute()  # List[str]
     label = UnicodeAttribute()
     last_updated_timestamp = NumberAttribute(default_for_new=get_current_timestamp)
+
+    def __eq__(self, other):
+        return isinstance(other, CategoryModel) and (
+            self.name == other.name
+            and self.version == other.version
```
manasaV3, line 31: Is there a reason why we have not included `version_hash`?

backend/api/models/_tests/test_category.py
```diff
@@ -1,23 +1,72 @@
...
+
+    def _get_version_hash(self, hash: str) -> str:
+        return f"{TEST_CATEGORY_VERSION}:{hash}"
 
+    def _save_models(self):
         from api.models.category import CategoryModel
 
-        monkeypatch.setattr(CategoryModel, "query", mock_category)
-        actual = CategoryModel.get_category("name", "version")
+        CategoryModel(
```
manasaV3, line 34: As the data gets saved in dynamo by a different workflow, we should ensure that the field name mappings for the dynamo calls are consistent. Adding data with explicit field name mapping here manually using `putItem` might help test for that. 

For example, if someone accidentally renames a field in the dynamo field like `UnicodeAttribute(attr_name='foo')`, these tests would not catch that as the pynamo model would take care of that mapping internally.

backend/api/models/_tests/test_category.py
```diff
@@ -1,23 +1,72 @@
...
+    def install_activity_table(self, aws_credentials, setup_env_variables):
+        from api.models.category import CategoryModel
+
+        with mock_dynamodb():
+            yield create_dynamo_table(CategoryModel, "category")
+
+    def _get_version_hash(self, hash: str) -> str:
+        return f"{TEST_CATEGORY_VERSION}:{hash}"
 
+    def _save_models(self):
```
manasaV3, line 31: Nit: Could we rename this method to `seed_data`?

backend/api/models/category.py
```diff
@@ -77,3 +77,13 @@ def get_all_categories(cls, version: str):
         print_perf_duration(start, "CategoryModel.get_all_categories()")
 
         return mapped_categories
+
+    def __eq__(self, other):
+        return isinstance(other, CategoryModel) and (
+            self.name == other.name
+            and self.version == other.version
```
manasaV3, line 84: What is the reason for not including `version_hash` in the equals check? 🤓

backend/api/models/_tests/test_category.py
```diff
@@ -1,23 +1,72 @@
...
-                    dimension="dimension2",
-                    hierarchy=["hierarchy1", "hierarchy2"],
-                ),
-            ]
-        )
+    @pytest.fixture
+    def setup_env_variables(self, monkeypatch):
+        monkeypatch.setenv("BUCKET", TEST_BUCKET)
+        monkeypatch.setenv("BUCKET_PATH", TEST_BUCKET_PATH)
+        monkeypatch.setenv("STACK_NAME", TEST_STACK_NAME)
```
manasaV3, line 19: I believe, setting the stack name here has no impact, as the `CategoryModel` class gets generated much before this test gets executed. The env variable for STACK_NAME will return `None` at that point making the dynamo table name `None-category`.

backend/api/models/category.py
```diff
@@ -10,7 +10,7 @@
 
 class CategoryModel(Model):
     class Meta:
-        host = os.getenv('LOCAL_DYNAMO_HOST')
+        host = os.getenv("LOCAL_DYNAMO_HOST")
```
manasaV3, line 13: Not a blocker: In the interest of consistency, can we follow [this](https://github.com/chanzuckerberg/napari-hub/blob/964059fa5f5978a3f58ff839fb1da31ff662f267/backend/api/models/install_activity.py#L29-L33) pattern here for the class design.

Also, I recently learned that in accordance to Python best practices, if a class only has class methods and static methods, it is better for it to be a module with individually importable methods. 

Also, setting up the pynamo model as a protected class reduces the access to dynamo directly from other places.

data-workflows/tests/test_handler.py
```diff
@@ -5,47 +5,83 @@
...
+        )
 
-    def _verify(self, activity_call_count: int = 0):
+    def _verify_update_activity(self, activity_call_count: int = 0):
         assert self._update_activity.call_count == activity_call_count
 
-    @pytest.mark.parametrize('event_type', ["Activity", "AcTiviTy", "ACTIVITY"])
+    def _verify_s3_seed(self, s3_seed_call_count: int = 0):
+        assert self._seed_s3_categories_workflow.call_count == s3_seed_call_count
+
```
manasaV3, line 32: Could be a single method? 
```
 def _verify(self, activity_call_count: int = 0, category_call_count: int = 0):
        assert self._update_activity.call_count == activity_call_count
        assert self._seed_s3_categories_workflow.call_count == category_call_count
```

data-workflows/tests/test_handler.py
```diff
@@ -5,47 +5,83 @@
...
+    @pytest.mark.parametrize("event_type", ["Activity", "AcTiviTy", "ACTIVITY"])
     def test_handle_event_type_in_different_case(self, event_type: str):
         from handler import handle
-        handle({'Records': [{'body': '{"type":"activity"}'}]}, None)
 
-        self._verify(activity_call_count=1)
+        handle({"Records": [{"body": '{"type":"activity"}'}]}, None)
```
manasaV3, line 37: Existing bug, the value for event_type should be passed here.

data-workflows/tests/test_handler.py
```diff
@@ -5,47 +5,83 @@
...
 
-    def _verify(self, activity_call_count: int = 0):
+    def _verify_update_activity(self, activity_call_count: int = 0):
         assert self._update_activity.call_count == activity_call_count
 
-    @pytest.mark.parametrize('event_type', ["Activity", "AcTiviTy", "ACTIVITY"])
+    def _verify_s3_seed(self, s3_seed_call_count: int = 0):
+        assert self._seed_s3_categories_workflow.call_count == s3_seed_call_count
+
+    @pytest.mark.parametrize("event_type", ["Activity", "AcTiviTy", "ACTIVITY"])
```
manasaV3, line 33: We should add tests cases for category event here. We could update the parameters to `event_type, activity_call_count, category_call_count`.

data-workflows/tests/test_handler.py
```diff
@@ -5,47 +5,83 @@
...
     def test_handle_activity_event_type(self):
         from handler import handle
-        handle({'Records': [
-            {'body': '{"type":"activity"}'},
-            {'body': '{"type":"bar"}'}
-        ]}, None)
-        self._verify(activity_call_count=1)
+
+        handle(
+            {"Records": [{"body": '{"type":"activity"}'}, {"body": '{"type":"bar"}'}]},
```
manasaV3, line 45: Could this message be updated to: 
```
{"Records": [
    {"body": '{"type":"activity"}'}, 
    {"body": '{"type":"seed-s3-categories"}'},
    {"body": '{"type":"bar"}'}
]}
```

So, it tests the case of the following method as well?

backend/api/models/_tests/test_category.py
```diff
@@ -1,23 +1,72 @@
...
+            name="name2",
+            version=TEST_CATEGORY_VERSION,
+            version_hash=self._get_version_hash("hash3"),
+            formatted_name="Name2",
+            dimension="dimension3",
+            label="label3",
+            hierarchy=["hierarchy3"],
+        ).save()
+
+    def test_get_category_has_result(
```
manasaV3, line 62: We should also add tests for the cases where the table is empty, or there is no match for a specific request.

backend/api/models/_tests/test_category.py
```diff
@@ -33,40 +82,19 @@ def test_get_category_has_result(self, monkeypatch):
...
-                Mock(
-                    formatted_name="name2",
-                    version="version",
-                    label="label3",
-                    dimension="dimension3",
-                    hierarchy=["hierarchy3"],
-                ),
-            ]
-        )
+    def test_get_all_categories(
```
manasaV3, line 85: We should also add tests for the case where the table is empty.

data-workflows/categories/tests/test_processor.py
```diff
@@ -0,0 +1,102 @@
...
+
+        with mock_dynamodb():
+            yield create_dynamo_table(CategoryModel, "category")
+
+    def _set_up_s3(self, bucket_name=TEST_BUCKET):
+        self._s3 = boto3.resource("s3")
+        bucket = self._s3.Bucket(bucket_name)
+        bucket.create()
+
+    def test_write_category_data(
```
manasaV3, line 60: We should also add tests for the failure cases.

manasaV3 (changes requested):

backend/api/models/category.py
```diff
@@ -23,57 +23,68 @@ class Meta:
...
+    """
+
+    category = []
+    start = time.perf_counter()
+
+    for item in CategoryModel.query(
+        slugify(name), CategoryModel.version_hash.startswith(f"{version}:")
+    ):
+        category.append(_get_category_from_model(item))
+
```
manasaV3, line 58: Nit: We could do this as follows leveraging some pythonic magic.

```suggestion
    results = CategoryModel.query(
        slugify(name), CategoryModel.version_hash.startswith(f"{version}:")
    )
     category = [ _get_category_from_model(item) for item in results ]
```

backend/api/models/category.py
```diff
@@ -23,57 +23,68 @@ class Meta:
...
-        for category in categories:
-            mapped_categories[category.formatted_name].append(
-                cls._get_category_from_model(category)
-            )
+def _get_category_from_model(category):
+    return {
+        "label": category.label,
+        "dimension": category.dimension,
+        "hierarchy": category.hierarchy,
+    }
```
manasaV3, line 43: 🔥

manasaV3 (approved):
🚢

manasaV3 (approved):

backend/api/models/category.py
```diff
@@ -23,57 +23,68 @@ class Meta:
...
+        slugify(name), CategoryModel.version_hash.startswith(f"{version}:")
+    ):
+        category.append(_get_category_from_model(item))
+
+    print_perf_duration(start, f"CategoryModel.get_category({name})")
+
+    return category
+
+
+def get_all_categories(version: str):
```
manasaV3, line 64: The caller of this method (ref [here](https://github.com/chanzuckerberg/napari-hub/pull/1049/files#diff-36d307d286a0bbe1266bd3dffb13ef84014283e3e08f30c79283c8c45b5e85ffL117)) still seems to be referencing it as if it were a class method for `CategoryModel`.

backend/api/models/category.py
```diff
@@ -23,57 +23,68 @@ class Meta:
...
-            )
+def _get_category_from_model(category):
+    return {
+        "label": category.label,
+        "dimension": category.dimension,
+        "hierarchy": category.hierarchy,
+    }
+
+
+def get_category(name: str, version: str):
```
manasaV3, line 46: The caller of this method (ref [here](https://github.com/chanzuckerberg/napari-hub/pull/1049/files#diff-36d307d286a0bbe1266bd3dffb13ef84014283e3e08f30c79283c8c45b5e85ffR124)) still seems to be referencing it as if it were a class method for `CategoryModel`.

We should update that reference to `api.models.category.get_category(..)`.

PR #1052, opened 2023-05-13
Fix maintenance spelling

richaagarwal:
Updates a few references to `maintenance` that were misspelled as `maintainence`.  These variables are misspelled in Split.io as well, so those will need to be updated as part of this change.  

Once this is merged to `main`, we should update the dev and staging feature flags, and after our next release, we should update the prod feature flags.


klai95 (approved):

manasaV3 (approved):

richaagarwal:
I've gone ahead and made the updates in Split.io for dev and staging environments, and also added this variable to the prod environment (without removing the old spelling so that it continues to work there).  After our next release, I'll clean up the prod feature flag.

PR #1067, opened 2023-05-19
Creating timing log decorator

manasaV3:
## Description

As the timing log allows for custom messages, it handles cases where a parameter identified by position in the log is passed as a keyword param and vice versa.


manasaV3:
Closing this, as it would require some more work for this to be adapted into the code.

Issue #1008, opened 2023-05-01
Plugin cards section component


codemonkey800:
Implement a new component that renders a section of plugin cards with a link to see more plugins of the same type. The UI for the plugin cards section component should have the following:

- Responsive for all screen sizes
- Hover states
- 3 plugin cards for desktop
- 2 plugin cards for smaller screens
- Show the following information:
  - Section title
  - Section icon
  - See all link
  - Specified metadata for a section (example: total installs or release date)

## Designs
  
<img width="400" alt="image" src="https://github.com/chanzuckerberg/napari-hub/assets/2176050/12338886-d601-4db6-8529-8849af093d42">

<img width="400" alt="image" src="https://github.com/chanzuckerberg/napari-hub/assets/2176050/2189a8cc-74bd-42e6-a460-5a6224074825">

PR #1075, opened 2023-05-25
Home page cards

codemonkey800:
## Description

https://github.com/chanzuckerberg/napari-hub/issues/1008

- Implements plugin cards section
- Adds plugin cards section to home page
- Disables home page redesign for E2E tests temporarily until `/plugins` page is ready

## Demos

image

image

image

### DIFF
--- a/frontend/src/components/HomePage/PluginCard.tsx
+++ b/frontend/src/components/HomePage/PluginCard.tsx
@@ -99,7 +99,7 @@ export function PluginCard({ className, metadataToShow, plugin }: Props) {
         {metadataToShow?.map((key) => (
           <div key={key}>
             <Text element="span" variant="bodyS">
-              {t(I18N_KEY_MAP[key] as I18nKeys<'pluginData'>)}:
+              {t(I18N_KEY_MAP[key] as I18nKeys<'pluginData'>) as string}:
             </Text>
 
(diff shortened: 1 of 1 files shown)


frontend/src/components/HomePage/PluginCard.tsx
```diff
@@ -0,0 +1,113 @@
...
+              .join(', ')}
+          </Text>
+
+          {plugin.authors.length > MAX_AUTHORS_COUNT && (
+            <>
+              {' '}
+              <Text className="underline mr-2" element="span" variant="bodyXS">
+                +{plugin.authors.length - MAX_AUTHORS_COUNT} more
+              </Text>
+            </>
```
manasaV3, line 95: I believe we are doing the truncation of authors in the interest of page real estate. What are your thoughts on having this be done by character count, as it might be a closer estimate to pixels used?
@Janeece

Janeece: If I recall correctly, when I discussed with with Danielle we concluded that Authors wouldn't like their names partially cut off, so I spec'd it to cut off by whole names instead of max character count (even though thats more straightforward). 

IMO, this is not a deal breaker behavior— if behavior will be smoother (and less lift) with character count instead thats fine with me.

manasaV3 (approved):

Issue #1057, opened 2023-05-16
Replace S3 category calls with DynamoDB


codemonkey800:
## `get_categories_mapping`

- Used for getting mapping for all categories
- Replace with `CategoryModel.get_all_categories(version)`

https://github.com/chanzuckerberg/napari-hub/blob/15ea22a7e9789f42844a5ff2241e9d30d66071eb/backend/api/model.py#L211-L213

## `get_category_mapping`

- Used for getting  a specific category
- Replace with `CategoryModel.get_category(category, version)`

https://github.com/chanzuckerberg/napari-hub/blob/15ea22a7e9789f42844a5ff2241e9d30d66071eb/backend/api/model.py#L215-L217

https://github.com/chanzuckerberg/napari-hub/blob/15ea22a7e9789f42844a5ff2241e9d30d66071eb/backend/preview/preview.py#L60-L62

richaagarwal:
@codemonkey800 I think enabling feature flagging on the back-end would still be very valuable (see https://github.com/chanzuckerberg/napari-hub/issues/811#issuecomment-1502426348) but I'd like to ensure we wrap up migrating the category work sooner than later, so let's not make it a dependency for this work.

manasaV3:
As we will be moving away from the current plugin update workflow triggered by the `/update` endpoint and to use data-workflows. I think it might be more optimal to only update the preview's usage of the categories to dynamo table. 

Also, if the categories have been sufficiently validated while guarded by a feature flag, would this issue be a good opportunity to work on updating all the categories endpoints to serve data from dynamo by default?

PR #1080, opened 2023-05-26
Remove use_dynamo_category query param

codemonkey800:
## Description

https://github.com/chanzuckerberg/napari-hub/issues/1057

Remove `use_dynamo_category` query param from categories endpoint so that we always query dynamo for the categories API. This change will any new preview page builds after it gets deployed to production:

https://github.com/chanzuckerberg/napari-hub/blob/383352d13d17ef0fc2c3f5f4d2125daadd8b1bf7/backend/preview/preview.py#L57

### DIFF
--- a/backend/api/app.py
+++ b/backend/api/app.py
@@ -114,19 +114,13 @@ def get_exclusion_list() -> Response:
 
 @app.route('/categories', defaults={'version': os.getenv('category_version', 'EDAM-BIOIMAGING:alpha06')})
 def get_categories(version: str) -> Response:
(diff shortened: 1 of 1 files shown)


richaagarwal (approved):

manasaV3 (approved):

manasaV3:
We should also clean up the updates made to the bdd tests as a part of this. https://github.com/chanzuckerberg/napari-hub/blob/main/backend/bdd_tests/test_category.py#L22-L37

Issue #1023, opened 2023-05-03
Ensure that Plugin Preview Page checklist reflects napari community priorities for metadata locations


neuromusic:
The plugin preview page still has a checklist item which tells plugin developers to provide a napari-hub-specific description in `napari-hub/DESCRIPTION.md`

This should be modified such that ANY description (whether from PyPI or `napari-hub/DESCRIPTION.md`) satisfies the checkboxes

richaagarwal:
@neuromusic to provide a screenshot.   @manasaV3 @codemonkey800 @klai95 to discuss in next backlog refinement

neuromusic:
screenshot from docs: 
![image](https://user-images.githubusercontent.com/1245615/236052216-61f4c937-1c3f-4fad-8488-c4afafc5ebe1.png)

kandarpksk:
Here's a live example of the plugin Kevin and I made during home week!
https://preview.napari-hub.org/klai95/test-detect-spots/1

manasaV3:
Relevant source code to look into: https://github.com/chanzuckerberg/napari-hub/blob/main/backend/preview/preview.py#L148

dgmccart:
I can update our eng docs [here](https://github.com/chanzuckerberg/napari-hub/wiki/Eng-Tech:-Plugin-Metadata#github-metadata) once these changes have been made.

klai95:
> This might require a spike ticket to figure out the required code changes to use `README.md` cc: @klai95
> 
> #1032

Details of the spike is here: https://github.com/chanzuckerberg/napari-hub/issues/1032#issuecomment-1550487150

richaagarwal:
I've opened a PR here - https://github.com/chanzuckerberg/napari-hub/pull/1082 - but would need someone else to look into testing these changes as these are untested as of now.  A couple notes on this:
- The reason we appear to render the `README.md` as the description on the napari hub site currently is that pypi uses that file to populate its `description` field. However, we don't specifically examine the github `README.md` ourselves (this PR will change that).
- I am assuming the reason for the discrepancy between the napari hub plugin page, and the preview page (in terms of what is available for the `description` field) is due to what is available in the wheel's metadata at the time of the preview page action.  Haven't verified that though. 

I'll leave this in code review for now but if anyone would like to claim testing of these, please do so :)

richaagarwal:
Moving this back into the backlog as we've paused work on this for now

richaagarwal:
[...]

PR #1082, opened 2023-05-26
Add README.md as a source for the description for preview page

richaagarwal:
### Description and overview of changes

- Addresses https://github.com/chanzuckerberg/napari-hub/issues/1023
- I ran `black` on this file. To view the functional changes, see this commit: https://github.com/chanzuckerberg/napari-hub/commit/beac6602e28d83f9d43bcd14333c8182cf78422b

### Notes
- The reason we appear to render the README.md as the description on the napari hub site currently is that packages are commonly configured to instruct PyPI to use their README to populate the `long_description` field, which we then use to populate a plugin's description on the hub. However, we don't specifically examine the github README.md ourselves (this PR will change that).
- I am assuming the reason for the discrepancy between the napari hub plugin page, and the preview page (in terms of what is available for the description field) is due to what is available in the wheel's metadata at the time of the preview page action. Haven't verified that though.

### DIFF
--- a/backend/utils/github.py
+++ b/backend/utils/github.py
@@ -10,13 +10,13 @@ from cffconvert.citation import Citation
 from requests.auth import HTTPBasicAuth
 from requests.exceptions import HTTPError
 
-from utils.utils import get_attribute, render_description
+from utils.utils import get_attribute
 from utils.auth import HTTPBearerAuth
 
 # Environment variable set through ecs stack terraform module
-github_client_id = os.environ.get('GITHUB_CLIENT_ID', None)
-github_client_secret = os.environ.get('GITHUB_CLIENT_SECRET', None)
-github_token = os.environ.get('GITHUB_TOKEN', None)
+github_client_id = os.environ.get("GITHUB_CLIENT_ID", None)
+github_client_secret = os.environ.get("GITHUB_CLIENT_SECRET", None)
+github_token = os.environ.get("GITHUB_TOKEN", None)
 
 auth = None
 if github_token:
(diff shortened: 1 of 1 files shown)


DragaDoncila:
> I am assuming the reason for the discrepancy between the napari hub plugin page, and the preview page (in terms of what is available for the description field) is due to what is available in the wheel's metadata at the time of the preview page action. 

This is correct. Although the long description is present in the wheel's `METADATA` file, it's not parsed as part of the package metadata for some reason. PyPI have their own logic to grab this long description, and they store it in the `description` field.

This means in the current (on `main`) implementation for the hub, we always have the PyPI description stored in the metadata, and it's overwritten by the content in `DESCRIPTION.md` (if present). The approach in this PR would also overwrite the PyPI description if the `DESCRIPTION.md` is present, but it will **also** re-overwrite the description field with the `README.md` content if `DESCRIPTION.md` is **not** present. 

This is unlikely to be an issue, as I would imagine there are probably only a handful of packages on PyPI that point to a different file than the `README.md` for their description, but it **is** possible, and in that case, we would overwrite the desired description with `README.md` content. 

It's really annoying that the `long_description` is in the `METADATA` file in the wheel, but we have (as far as I can tell) no blessed way to access it before it hits PyPI. The closest I've found is `build`'s `project_wheel_metadata`, which has the description available, but under a private attribute 🤔 .

I think if we're explicit that we're rendering the `README.md` and **not** the PyPI description field (even though they're in 99.9999% of cases the same), then this implementation should be ok and I can look at adding tests. Happy to chat more when you're back @richaagarwal

richaagarwal:
Thanks for looking into this, @DragaDoncila! 

> but it will also re-overwrite the description field with the README.md content if DESCRIPTION.md is not present.

Ah good catch! I'd be a little hesitant to move forward with this change in that case, as we'd probably want to continue for the hub plugin page to display any PyPI-specific description in the absence of a napari-specific description, even if it's not a common scenario that it would deviate from what's in the README.

If there was an easy way to isolate this change to the preview page only, that might be worth considering.

@DragaDoncila I'll reach out to discuss further, but you can hold off on testing for now.

cc @neuromusic

PR #1098, opened 2023-06-06
Move data-workflows dynamo models to napari-hub-commons

manasaV3:
## Description

- Helps reduce code duplication, by moving the models used in data-workflow module to napari-hub-commons. 
- This will further help when the models are reused in backend


manasaV3:
> I understand moving the models to `nhcommons`, but I'm curious why the models were made private here such that now the workflows assemble lists of plain dicts before writing.

The reason to make the models private is to prevent any accidental usage of those models, which would result in changes to the dynamo data. This way, we only allow expected behaviour around handling the data. It is encapsulation to add more security to the data. 😸 

> I also wonder about the `to_model` which is basically doing `_Model(**data)` but will silently ignore extra fields. Also that it will use defaults (`None`) for any unspecified field. I trust your DB experience here but would it be better to add `null=True` and a default value to the model attributes?

The `to_model` method, should also do some basic data cleanup if necessary. 

In Pynamo, checks are made if a value is `None` to verify if the value is null is valid by checking if `null=True` has been set up for that attribute or else it throws an error. Also, fields that have been marked nullable have been done intentionally. In some cases, it might be to reduce space utilization, and in others, it might be for indexing purposes. 

Pynamo also throws an error if we add attributes not explicitly defined in the model. So, our method protects against clients inadvertently passing unknown attributes.


> Finally I know this is a super-nit but I think I'd prefer `from_dict` to `to_model` for that method name 😬.

I can incorporate this name suggestion. 😁

aganders3:
Thanks for the explanations! I think I understand.  It makes sense if the models are also the interface to reading/writing the database; they're not _just_ objects that represent entries. I've used Pydantic or Marshmallow for this in the past to keep a schema separate from the DB model.

I'll give the code another quick look but otherwise LGTM.

data-workflows/activity/github_activity_model.py
```diff
@@ -100,7 +72,7 @@ def transform_and_write_to_dynamo(data: dict[str, List], activity_type: GitHubAc
     """
     LOGGER.info(f'Starting item creation for github-activity type={activity_type.name}')
 
-    batch = GitHubActivity.batch_write()
+    batch = []
 
     start = time.perf_counter()
     count = 0
```
aganders3, line 78: I would remove this `count` variable and use `len(batch)` instead.

data-workflows/activity/install_activity_model.py
```diff
@@ -36,53 +33,30 @@ def get_query_timestamp_projection(self) -> str:
...
-        return False
-
-
 def transform_and_write_to_dynamo(data: dict[str, List],
                                   activity_type: InstallActivityType) -> None:
-    LOGGER.info(f'Starting item creation for install-activity type={activity_type.name}')
-    batch = InstallActivity.batch_write()
+    LOGGER.info(f"Starting item creation for install-activity type={activity_type.name}")
+    batch = []
     count = 0
```
aganders3, line 40: Same comment about `count` - I know this was part of the original code but I think it was because the writes were never collected in a list before.

aganders3 (dismissed):
Looks good!

data-workflows/activity/github_activity_model.py
manasaV3: Good catch! This should now be viable with the update. :)

manasaV3:
> Thanks for the explanations! I think I mostly understand, haha. It makes sense if the models are also the interface to writing the database; they're not just objects that represent entries. It still feels a little weird to have these loose types (plain dicts) as it's like a hidden API. I've used Pydantic or Marshmallow for this to keep a schema separate from the DB model, but that's obviously a much bigger refactoring that I'm not suggesting here.

I see what you mean, maybe as the objects get more complex, we could look into incorporating a serializing library here. 😁 

> I feel like I might want to understand this a bit more, but maybe better to talk about it elsewhere 😅. When you say client here do you mean external clients or e.g. our data workflows?

As of now, the planned clients for this should only be our data workflows and backend. 

>  In either case would we actually want that error rather than just dropping the attributes?
At this point in time, I think quietly handling it might be the best way to go. But, maybe we could add a warning log too.

PR #1099, opened 2023-06-07
fix categories API

codemonkey800:
## Description

fixes the categories API errors as a result a required code change that was not added in #1049

### Before

error in localhost:

image

### After

successful API call:

image


manasaV3 (approved):
:shipit:

PR #1101, opened 2023-06-09
Handling non github source code

manasaV3:
## Description

Relates to: https://github.com/chanzuckerberg/napari-hub/issues/1105

The PR introduces fixes for the following:
- Does not process for github metadata if the code repository does not match the prefix expected for github
- Gracefully handles json parsing error when fetching license data from github.

### DIFF
--- a/backend/api/model.py
+++ b/backend/api/model.py
@@ -214,7 +214,7 @@ def build_plugin_metadata(plugin: str, version: str) -> Tuple[str, dict]:
     if not metadata:
         return plugin, metadata
     github_repo_url = metadata.get('code_repository')
-    if github_repo_url:
+    if github_repo_url and github_repo_url.startswith("https://github.com/"):
         metadata = {**metadata, **get_github_metadata(github_repo_url)}
     if 'description' in metadata:
         metadata['description_text'] = render_description(metadata.get('description'))
--- a/backend/utils/github.py
+++ b/backend/utils/github.py
@@ -218,11 +218,11 @@ def get_citation_author(citation_str: str) -> Union[List[Dict[str, str]], None]:
         logging.error(e)
         return []
     authors = []
-    for entry in citation_yaml['authors']:
-        if 'given-names' in entry and 'family-names' in entry \
-                and entry['given-names'] and entry['family-names']:
-            author_name = entry['given-names'] + " " + entry['family-names']
-            authors.append({'name': author_name})
-        elif 'name' in entry and entry['name']:
-            authors.append({'name': entry['name']})
+    for author_entry in citation_yaml['authors']:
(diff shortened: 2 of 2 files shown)


backend/api/model.py
```diff
@@ -214,7 +214,7 @@ def build_plugin_metadata(plugin: str, version: str) -> Tuple[str, dict]:
     if not metadata:
         return plugin, metadata
     github_repo_url = metadata.get('code_repository')
-    if github_repo_url:
+    if github_repo_url and github_repo_url.startswith("https://github.com/"):
```
aganders3, line 217:
```suggestion
    if github_repo_url and github_repo_url.startswith(
        ("https://github.com/", "https://www.github.com/", "http://github.com/", "http://www.github.com/")
    ):
```
You might want to check for variants such as these - maybe even allow people to omit the scheme?

backend/utils/github.py
```diff
@@ -88,7 +89,7 @@ def get_license(url: str, branch: str = 'HEAD') -> [str, None]:
             return None
         else:
             return spdx_id
-    except HTTPError:
+    except (HTTPError, JSONDecodeError):
```
aganders3, line 92: You may want to log an error here to give insight if there are common errors.

backend/utils/github.py
```diff
@@ -215,9 +218,11 @@ def get_citation_author(citation_str: str) -> Union[Dict[str, str], None]:
...
-            authors.append({'name':author_entry['given-names'] + " " + author_entry['family-names']})
-        elif 'name' in author_entry and author_entry['name']:
-            authors.append({'name':author_entry['name']})
+    for entry in citation_yaml['authors']:
+        if 'given-names' in entry and 'family-names' in entry \
+                and entry['given-names'] and entry['family-names']:
+            author_name = entry['given-names'] + " " + entry['family-names']
+            authors.append({'name': author_name})
+        elif 'name' in entry and entry['name']:
+            authors.append({'name': entry['name']})
```
aganders3, line 227: Just checking - this is just formatting, right?

aganders3 (approved):
LGTM

backend/api/model.py
manasaV3: I am gonna capture this in a backlog issue, as this might require changes in a few other places.

backend/utils/github.py
manasaV3: Yes, but this should not have been a part of this PR. 😞

backend/utils/github.py
manasaV3: I think from the logging that exists in the newer workflow, we can see the common errors are from 404 on the file.

PR #1112, opened 2023-06-20
Returning empty list on no collection

manasaV3:
## Description

Fixes error `TypeError: The view function for 'collections' did not return a valid response. The function either returned None or ended without a return statement.`

### DIFF
--- a/backend/api/plugin_collections.py
+++ b/backend/api/plugin_collections.py
@@ -1,16 +1,20 @@
 import yaml
 from api.model import get_plugin
 from utils.github import get_file
+import logging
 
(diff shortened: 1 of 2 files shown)


aganders3 (approved):

manasaV3:
This might not be the right fix for this, as this silences the occurrences of errors on fetching data from github.

aganders3:
Ha - right after I approve making me look foolish! 😆

manasaV3:
@aganders3  sorry, I thought a different endpoint was having this issue, and this seemed like the right fix for that. On double-checking, I realized that this was not the correct solution. But your quick code review is always appreciated. 😄

richaagarwal (approved):

Issue #950, opened 2023-03-17
Update /plugins/index API to use dynamoDB


manasaV3:
### Description

The API should return only public plugins metadata as a list
e.g. visibility == 'public' and is_blocked == False

### Implementation

-  If the incoming request is in the test, fetch the data from dynamoDb or else fetch the results from s3.

- Scan the for records from the `plugin` table on the index on `is_latest` with filter `visibility=public and is_blocked does not exist` and pagination to ensure that all the results have been fetched. 

```
aws dynamodb scan \
--table-name dev-test-plugin \
--index-name dev-test-plugin_name-is_latest \
--filter 'visibility=:visibility and attribute_not_exists(is_blocked)' \
--projection-expression 'data, release_date' \
--expression-attribute-values '{
  ":visibility": {"S": "public"}
}'
```

In the above client call, replace table-name and index-name with the correct table name by stage and stack name.  
This can be implemented in the backend using a scan from boto3 linked [here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Client.scan).

[Link to pynamo resource](https://pynamodb.readthedocs.io/en/stable/batch.html?#scan-filters)  for the same.

- When constructing the response use the data field, and retain only the following fields from the data: 
```
'name', 'summary', 'description_text', 'description_content_type', 'authors', 'license', 'python_version', 'operating_system',
'release_date', 'version', 'first_released', 'development_status', 'category', 'display_name', 'plugin_types', 'reader_file_extensions',
'writer_file_extensions', 'writer_save_layers', 'npe2', 'error_message', 'code_repository'
```
- If there are two entries for a plugin, use the record with the latest `release_date`

### Data to be Logged

- Time taken for dynamo scan and transform
- Total number of results returned

richaagarwal:
We need to put a little more thought into this one before beginning implementation - we may want to secure this with an API key or consider alternatives to it once we introduce a search API.  I'll label this blocked for now

richaagarwal:
Following up on the above comment, we should secure this with an API key. If an API key is not provided, an error message should be returned.  TBD on the exact contents of the error message, but that shouldn't block work on this ticket for now.

manasaV3:
Also add functionality from: https://github.com/chanzuckerberg/napari-hub/issues/993

Issue #864, opened 2023-01-30
Update plugin data API endpoint to use dynamoDB


manasaV3:
### Description

The API should only return plugin data if it has public or hidden visibility and is not blocked.
e.g. visibility IN ('public', 'hidden') and is_blocked == False

#### If the version is not specified in the request:

- If the incoming request is in the test, fetch the data from dynamoDb or else fetch the results from s3.

- Convert the plugin_name to lowercase, and use it as the hash_key.

- Filter the index on `is_latest` field.

- Query the `plugin` table with additional filters `visibility='public' or 'hidden' and is_blocked does not exist`.

```
aws dynamodb query \
--table-name dev-test-plugin \
--index-name: dev-test-plugin_name-is_latest
--key-conditions 'name=:plugin_name' \
--filter-conditions 'visibility IN ('public', 'hidden) AND attribute_not_exists(is_blocked)' \
--projection-expression 'data, release_date' \
--expression-attribute-values '{
":plugin_name": {"S": "foo"}
}'
```

#### Processing the query results: 

In the above client call, replace table-name and index-name with the correct table name by stage and stack name. Also, update the expression attribute value to have relevant values for all the fields. These queries can be implemented in the backend using a query from boto3 linked [here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Client.query).

[Link to pynamo resource](https://pynamodb.readthedocs.io/en/stable/batch.html?#query-filters) for the same.

- If multiple records are returned, use the one with the latest release_date

- In the data attribute, remove/update any field as needed and return it as the response. 

- If the dynamo query returns no result, return `{}` as the response. 

### Data to be Logged

- Time taken for dynamo query
- Count of result for the query

manasaV3:
Check with @richaagarwal before implementation to verify "If the version is specified in the request" use case needs to be supported.

manasaV3:
**[Currently deprioritized]**
#### If the version is specified in the request:

- If the incoming request is in the test, fetch the data from dynamoDb or else fetch the results from s3.

- Convert the plugin_name to lowercase, and use it as the hash_key.

- Add the sort key to the key condition, by appending `:aggregate` to the `version`.

- Query the `plugin` table with additional filters `visibility='public' or 'hidden' and is_blocked does not exist`.

```
aws dynamodb query \
--table-name dev-test-plugin \
--key-conditions 'name=:plugin_name AND version_type = :version_type)' \
[...]

PR #1115, opened 2023-06-22
Adding Plugins API to use dynamo

manasaV3:
## Description
Relates to https://github.com/chanzuckerberg/napari-hub/issues/864, https://github.com/chanzuckerberg/napari-hub/issues/950

### Changes introduced:
- Makes tests for backend models environment agnostic
- Adding dynamo support for plugin endpoints

### DIFF
--- a/.happy/terraform/modules/ecs-stack/main.tf
+++ b/.happy/terraform/modules/ecs-stack/main.tf
@@ -514,6 +514,7 @@ data aws_iam_policy_document backend_policy {
       module.plugin_metadata_dynamodb_table.table_arn,
       module.plugin_dynamodb_table.table_arn,
       module.plugin_blocked_dynamodb_table.table_arn,
+      "${module.plugin_dynamodb_table.table_arn}/index/*",
     ]
   }
 
--- a/backend/api/_tests/test_shield.py
+++ b/backend/api/_tests/test_shield.py
@@ -1,26 +1,38 @@
 import unittest
 from unittest.mock import patch
 
-from api import model
+import api.shield as shield
 
 
-@patch.object(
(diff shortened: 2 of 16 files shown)


aganders3:
I have issues with `preview/_tests/test_preview_meta.py::test_clone_invalid_plugin` - it prompts for GitHub username on CLI.

Also should I be able to run the bdd tests? If so, can you help me because I can't figure it out 😆?

manasaV3:
> I have issues with `preview/_tests/test_preview_meta.py::test_clone_invalid_plugin` - it prompts for GitHub username on CLI.

That is the expected behaviour. It is not a bug. It is a feature. 😅  
This is triggered by a [specific test](https://github.com/chanzuckerberg/napari-hub/blob/main/backend/preview/_tests/test_preview_meta.py#L26). It is for an invalid repo that doesn't exist. So, git tries to handle the case where it might be a private repository and tries to get the username and password to access it. You should be able just to provide empty strings for the test to continue. 

> Also should I be able to run the bdd tests? If so, can you help me because I can't figure it out 😆?

You can run it locally, but I usually run them separately. I will look into enabling running that along with the other tests. 

I run them from the `bdd_tests` folder. They depend on the `PREFIX` environment variable to determine which environment they should run the tests against. 

The values for the prefix can be set to `staging` for the staging environment, `prod` for the prod environment, and any other value will try and run the test against a dev environment of that name.

If the `PREFIX` is not set, it runs the tests against the local application at `http://localhost:12345`.

aganders3:
Thanks! Unfortunately I don't think I'll have time to look at this until I get back from vacation on the 11th. Sorry I hope this doesn't hold you up.

Issue #1123, opened 2023-07-06
Update link for `Plugins` to redirect to new search page


codemonkey800:
Right now the link goes to the home page like in the old design, but for the redesign it should redirect to the new search page located at `/plugins`

<img width="143" alt="image" src="https://github.com/chanzuckerberg/napari-hub/assets/2176050/79540166-2641-4f68-8c00-8e8d45f6e73b">

PR #1124, opened 2023-07-06
Fix plugins redirect

codemonkey800:
## Description

#1123

Fixes `Plugins` link so that it redirects to `/plugins` page instead of the home page when the redesign flag is on


manasaV3 (approved):
:shipit:

Issue #1122, opened 2023-07-03
Remove home page redesign double footer 


codemonkey800:
There's two footers visible at the bottom of the plugin search page

<img width="1440" alt="image" src="https://github.com/chanzuckerberg/napari-hub/assets/2176050/f48dbfd1-43fd-427a-aa27-266dd90d669e">

PR #1125, opened 2023-07-06
fix double footer

codemonkey800:
## Description

#1122

Fixes issue with double footer showing on `/plugins` page. this was because the UI for the old layout was still visible with the flag on. this also fixes the alignment for the footer on the `/plugins` page


manasaV3 (approved):
🚢 it!

PR #1126, opened 2023-07-07
Plugin optimization and cleanup

manasaV3:
## Description

Changes introduced:
- Updates logging of old plugins to only plugins that have been validated as old plugin
- Handles no plugin manifest record gracefully
- Adds unit tests for dynamo models
- Adds alarm for data workflows plugin execution failure
- Adds support for blocked plugins to have exclusions of different types.

### DIFF
--- a/.happy/terraform/modules/ecs-stack/main.tf
+++ b/.happy/terraform/modules/ecs-stack/main.tf
@@ -389,6 +389,7 @@ resource aws_lambda_event_source_mapping data_workflow_plugin_metadata_event_sou
   function_name     = module.data_workflows_lambda.function_name
   batch_size        = 100
   starting_position = "LATEST"
+  maximum_batching_window_in_seconds  = 60
 }
 
 module api_gateway_proxy_stage {
--- a/backend/api/models/_tests/test_plugin.py
+++ b/backend/api/models/_tests/test_plugin.py
@@ -1,0 +1,205 @@
+import pytest
+from moto import mock_dynamodb
+
+from api.models._tests.conftest import create_dynamo_table
+from api.models import plugin
+
+
+class TestPlugin:
+
+    @pytest.fixture()
+    def plugin_table(self, aws_credentials, dynamo_env_variables):
+        with mock_dynamodb():
+            yield create_dynamo_table(plugin._Plugin, "plugin")
+
+    @pytest.fixture()
+    def plugin1_data2_2(self):
+        return {
+            "authors": [{"name": "Napari Hub"}],
+            "code_repository": "https://github.com/naparihub/plugin-1",
+            "description_content_type": "text",
+            "description_text": "napari plugin with a detailed description.",
+            "development_status": ["Development Status :: 2 - Pre-Alpha"],
+            "display_name": "Plugin 1",
+            "first_released": "2021-12-21T11:39:40.715897Z",
+            "license": "BSD-3-Clause",
+            "name": "plugin-1",
+            "npe2": True,
+            "operating_system": ["Operating System :: OS Independent"],
+            "plugin_types": ["reader"],
+            "python_version": ">=3.7",
+            "reader_file_extensions": ["*.pdb", "*.cif"],
+            "release_date": "2022-01-31T17:59:24.494345Z",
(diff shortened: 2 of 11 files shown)


backend/api/models/plugin_blocked.py
```diff
@@ -17,13 +17,15 @@ class Meta:
         pass
 
     name = UnicodeAttribute(hash_key=True)
+    exclusion = UnicodeAttribute(null=True)
```
richaagarwal, line 20: What does `exclusion` mean/ what would be an example value for this? And how is this field used? Is this separate from what's in the `is_excluded` field of a plugin?

Also, since we are deprecating `/plugins/excluded`, I don't think we'd be displaying this information anywhere, but fine to have for internal purposes if needed.

.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -389,6 +389,7 @@ resource aws_lambda_event_source_mapping data_workflow_plugin_metadata_event_sou
   function_name     = module.data_workflows_lambda.function_name
   batch_size        = 100
   starting_position = "LATEST"
+  maximum_batching_window_in_seconds  = 60
```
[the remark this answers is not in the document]
manasaV3: Trial and error. :) 
And, also 1 minute seems like a good enough upper threshold to wait before all data for aggregation is ready.

backend/api/models/plugin_blocked.py
manasaV3: The `exclusion` field is there just to ensure there is more extensibility for blocked plugins. In our blocked plugins, we have cases like invalid, which don't actually map to a specific visibility. If we ever wanted to add comments like that for internal usage, we could use this field. The value from this field will only be visible in our internal usage and not reflect in `is_excluded` in plugins.

richaagarwal: Got it, thanks for the clarification!  What are your thoughts on renaming this to something like `exclusion_reason` (or preferably even `blocked_reason`, especially if this is distinct from excluded plugins, as tracked by the `is_excluded` field)?

backend/api/models/_tests/test_plugin_metadata.py
```diff
@@ -0,0 +1,39 @@
...
+class TestPluginMetadata:
+
+    @pytest.fixture()
+    def plugin_metadata_table(self, aws_credentials, dynamo_env_variables):
+        with mock_dynamodb():
+            yield create_dynamo_table(
+                plugin_metadata._PluginMetadata, "plugin-metadata"
+            )
+
+    def _put_items(self, table):
```
aganders3, line 19: Again super-nit, this name seems inconsistent.

aganders3 (approved):
Looks good!

backend/api/models/_tests/test_plugin.py
```diff
@@ -0,0 +1,205 @@
...
+        self._put_items(plugin_table, data)
+
+        actual = plugin.get_plugin(name, None)
+
+        expected = {}
+        if name == "plugin-1":
+            expected = plugin1_data2_2
+        elif name == "plugin-2":
+            expected = plugin2_data1_0_0
+        assert actual == expected
```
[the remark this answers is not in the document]
manasaV3: That is awesome. Thanks for this tip. I might leverage that a little differently but this would clean that up nicely. :)

backend/api/models/_tests/test_plugin.py
```diff
@@ -0,0 +1,264 @@
...
+            "writer_file_extensions": [".zarr"],
+            "writer_save_layers": ["labels", "image"],
+        }
+
+    @pytest.fixture()
+    def plugin2_data0_5(self):
+        return {"foo": "bar", "release_date": "2025-04-03"}
+
+    @pytest.fixture()
+    def plugin2_data1_0_0(self):
```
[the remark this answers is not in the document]
manasaV3: I now see it is harder to see the pattern, but when there are multiple versions for a plugin, we use the version of the plugin in the name.  😞 

For example, this is the data for plugin2's version=1.0.0.

Issue #1035, opened 2023-05-08
Backend to return plugin sections data for home page


codemonkey800:
Create a backend endpoint to return various sections of the homepage consistently.

manasaV3:
Things that need clarity:

- How often do we refresh the sections?
- How do we handle duplicates across the sections? What is the priority for sections?

manasaV3:
API Contract for endpoint: https://docs.google.com/document/d/1Md39pS3v7yP8a7fNb6qfAmFY4arVUjIuJZTi3JVz9hw/edit#

manasaV3:
The endpoint should be API key protected. 
```
curl -X GET --header "X-API-Key: secret-api-key" https://api.napari-hub.org/plugin/home/sections/foo,bar,baz
```

PR #1132, opened 2023-07-10
Adding plugins api for homepage

manasaV3:
## Description
Relates to: https://github.com/chanzuckerberg/napari-hub/issues/1035

## Demo

- [Test endpoint using s3](https://api.dev.napari-hub.org/dev-shared/plugin/home/sections/newest,recently_updated,top_install,plugin_types?use_dynamo_plugin=False)
- [Test endpoint using dynamo](https://api.dev.napari-hub.org/dev-shared/plugin/home/sections/newest,recently_updated,top_install,plugin_types?use_dynamo_plugin=True)

### TODO
- [ ] Add unit tests
- [x] Add bdd tests
- [ ] Add check for API key on call

### DIFF
--- a/backend/api/home.py
+++ b/backend/api/home.py
@@ -1,0 +1,75 @@
+from typing import List, Dict, Set, Callable, Union
+
+from api.model import get_index
+from random import sample
+from datetime import datetime
+
+
+DEFAULT_FIELDS = {
+    "authors", "display_name", "first_released", "name", "release_date",
+    "summary", "total_installs",
+}
+PLUGIN_TYPES = ["reader", "sample_data", "widget", "writer"]
+
+
+def _filtered(data: Dict) -> Dict:
+    return {field: data.get(field) for field in DEFAULT_FIELDS}
+
+
+def _get_plugin_type() -> str:
+    index = int((datetime.now().minute / 5) % 4)
+    return PLUGIN_TYPES[index]
+
+
+def _get_plugins_by_type(index: List[Dict], limit: int) -> Dict:
+    plugin_type = _get_plugin_type()
+
+    plugins_of_type = list(
+        filter(lambda item: plugin_type in item.get("plugin_types", []), index)
+    )
+    rand_sample = sample(plugins_of_type, min(limit, len(index)))
+    random_sample = [_filtered(plugin) for plugin in rand_sample]
+    return {"plugin_types": {"type": plugin_type, "plugins": random_sample}}
+
+
+def _get_plugins_by_sort(
+        index: List[Dict], limit: int, key: str, default_val: Union[str, int]
+) -> Dict[str, List]:
+    index.sort(key=lambda item: item.get(key, default_val))
+    upper_limit = min(limit, len(index))
+    return {"plugins": [_filtered(index.pop()) for i in range(0, upper_limit)]}
+
+
+def _get_newest_plugins(index: List[Dict], limit: int) -> Dict:
+    plugins = _get_plugins_by_sort(index, limit, "first_released", "")
+    return {"newest": plugins}
+
+
+def _get_recently_updated_plugins(index: List[Dict], limit: int) -> Dict:
+    plugins = _get_plugins_by_sort(index, limit, "release_date", "")
+    return {"recently_updated": plugins}
+
+
+def _get_top_installed_plugins(index: List[Dict], limit: int) -> Dict:
+    plugins = _get_plugins_by_sort(index, limit, "total_installs", 0)
+    return {"top_install": plugins}
+
+
+def get_handler_by_section_name() -> Dict[str, Callable]:
+    return {
+        "plugin_types": _get_plugins_by_type,
+        "newest": _get_newest_plugins,
+        "recently_updated": _get_recently_updated_plugins,
+        "top_install": _get_top_installed_plugins,
+    }
+
+
+def get_plugin_sections(
+        sections: Set[str], use_dynamo: bool, limit: int = 3
+) -> Dict[str, Dict]:
+    response = {}
+    index = get_index(use_dynamo)
+    for name, handler in get_handler_by_section_name().items():
+        if name.lower() in sections:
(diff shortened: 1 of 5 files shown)


backend/api/home.py
```diff
@@ -0,0 +1,75 @@
...
+def _get_top_installed_plugins(index: List[Dict], limit: int) -> Dict:
+    plugins = _get_plugins_by_sort(index, limit, "total_installs", 0)
+    return {"top_install": plugins}
+
+
+def get_handler_by_section_name() -> Dict[str, Callable]:
+    return {
+        "plugin_types": _get_plugins_by_type,
+        "newest": _get_newest_plugins,
+        "recently_updated": _get_recently_updated_plugins,
```
richaagarwal, line 62: Does this return e.g. `"recently_updated": {"recently_updated": [list of plugins]}` ?  If so, is the nested dictionary here needed? It seems a little redundant.

backend/api/home.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def _get_recently_updated_plugins(index: List[Dict], limit: int) -> Dict:
+    plugins = _get_plugins_by_sort(index, limit, "release_date", "")
+    return {"recently_updated": plugins}
+
+
+def _get_top_installed_plugins(index: List[Dict], limit: int) -> Dict:
+    plugins = _get_plugins_by_sort(index, limit, "total_installs", 0)
+    return {"top_install": plugins}
```
richaagarwal, line 55: Very minor but `top_installs` or `top_installed` (plural) would read slightly clearer than the singular `install`

backend/api/home.py
```diff
@@ -0,0 +1,75 @@
...
+    random_sample = [_filtered(plugin) for plugin in rand_sample]
+    return {"plugin_types": {"type": plugin_type, "plugins": random_sample}}
+
+
+def _get_plugins_by_sort(
+        index: List[Dict], limit: int, key: str, default_val: Union[str, int]
+) -> Dict[str, List]:
+    index.sort(key=lambda item: item.get(key, default_val))
+    upper_limit = min(limit, len(index))
+    return {"plugins": [_filtered(index.pop()) for i in range(0, upper_limit)]}
```
richaagarwal, line 40: I think you could simplify lines 39 & 40 by instead slicing the list in line 38 to the value of the limit.  You could get the last _n_ items, or _first_ n items, depending on the sort order you need, although I would find it clearer if you explicitly pass in ascending or descending to the sort method and always get the first _n_ items in the list (e.g. for newest plugins and recently updated plugins, you'd want descending order)

backend/api/home.py
```diff
@@ -0,0 +1,75 @@
...
+        "recently_updated": _get_recently_updated_plugins,
+        "top_install": _get_top_installed_plugins,
+    }
+
+
+def get_plugin_sections(
+        sections: Set[str], use_dynamo: bool, limit: int = 3
+) -> Dict[str, Dict]:
+    response = {}
+    index = get_index(use_dynamo)
```
richaagarwal, line 71: Is there a way to avoid getting the full index for when the DynamoDB flag is on?  Wouldn't it be more efficient to query DynamoDB directly for e.g. the 3 most recently updated plugins, etc?

backend/api/home.py
manasaV3: No, this dictionary returns a mapping of the method to be called for a specific section. 

For example, if the `"recently_updated"` is a required section, the `_get_recently_updated_plugins` method will be called, which will return `{ "recently_updated": { "plugins": [...] } }`.

richaagarwal: Oh, got it! I see now that it's looking at `name` vs the `handler`. Thanks for clarifying. It would be nice to not have to define the section name more than once, but I think it's fine for now :)

backend/api/home.py
manasaV3: I intentionally chose to do it this way over slicing the list because `pop(..)` removes the item from the list. This is done so that we can prevent a plugin from recurring in multiple sections.  

If we were to slice the index, we would have to reassign the value of the sliced index to the index variable, and I found this to be a cleaner approach. Let me know your thoughts on this.

richaagarwal: Oh, that's clever! I wonder if there's a way to make it clearer that the index is being shared/modified across sections (particularly in case future changes are made to one section's function that otherwise unintentionally impacts another section).

backend/api/home.py
manasaV3: Just to ensure that we are on the same page, the final response structure looks like:

```
 {
  plugin_types: {
    type: ""
    plugins: []
  }
  newest: {
    plugins: []
  }
  recently_updated: {
     plugins: []
  }
  top_install: {
     plugins: []
  }
}
```

backend/api/home.py
manasaV3: Unfortunately, that would require a completely new index. 😞

backend/api/home.py
manasaV3: I agree. It could be clearer. I will think about how to make it more obvious in the code or add a comment to it. :)

backend/api/home.py
richaagarwal: Yes, I believe we're on the same page! I meant in case the variable name for the section changed for any reason, there's a couple places you'd have to update the references (e.g. for `top_install`, https://github.com/chanzuckerberg/napari-hub/pull/1132/files#diff-4f6c6f12cffea8b1d0b859d90ae31e7bc762ead6a01485f744d46e42b907a83dR59 and https://github.com/chanzuckerberg/napari-hub/pull/1132/files#diff-4f6c6f12cffea8b1d0b859d90ae31e7bc762ead6a01485f744d46e42b907a83dR67). 

It's pretty minor and in this case I think the redundancy is worth the clarity of what exactly is being returned rather than trying to abstract it; just wanted to mention though :)

richaagarwal (approved):

PR #1160, opened 2023-07-19
Fix home page redesign bugs

codemonkey800:
## Description

Should fix most of the bugs found during testing from QA and Design:

- QA Bugs
  - https://github.com/chanzuckerberg/napari-hub/issues/1135
  - https://github.com/chanzuckerberg/napari-hub/issues/1135
  - https://github.com/chanzuckerberg/napari-hub/issues/1139
  - https://github.com/chanzuckerberg/napari-hub/issues/1140
  - https://github.com/chanzuckerberg/napari-hub/issues/1141
  - https://github.com/chanzuckerberg/napari-hub/issues/1142
- [Design bugs](https://github.com/chanzuckerberg/napari-hub/issues/1005#issuecomment-1625815305)
  - Fix navigation to `/plugins` page
  - Fix page transitions when navigating away from home page
  - Increase card spacing for narrow screens
  - Change icon based on plugin type
  - Clear all should be working on desktop
  - Update category background color when selected to primary 200
  - Fix right alignment of hamburger menu
  - Fix categories being all caps on narrow screens
  - Stack content of search result on narrow screens

PR #1167, opened 2023-07-24
Fix search result columns

codemonkey800:
## Description

Fixes issue introduced in #1160 around the column span for search result metadata

### Before

image

### After

image


manasaV3 (approved):
:shipit:

PR #1176, opened 2023-07-27
feature flag cache

codemonkey800:
## Description

Adds TTL cache to frontend feature flags so that flags are cached on fetch and can be reused on other pages when they render. This sets the cache to 10 minutes, but the actual time it takes for feature flag changes to propagate depend on when the page was cached on the CDN

### DIFF
--- a/frontend/src/utils/ssr.ts
+++ b/frontend/src/utils/ssr.ts
@@ -48,8 +48,8 @@ type ServerSidePropsHandlerProps<P> = P &
     featureFlags?: FeatureFlagMap;
   };
 
-// Expire feature flags after 10 minutes
-const FLAG_CACHE_TTL = 1000 * 60 * 10;
+// Expire feature flags after 5 minutes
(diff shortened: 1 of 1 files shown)


richaagarwal (approved):

frontend/src/utils/ssr.ts
```diff
@@ -46,6 +48,11 @@ type ServerSidePropsHandlerProps<P> = P &
     featureFlags?: FeatureFlagMap;
   };
 
+// Expire feature flags after 10 minutes
+const FLAG_CACHE_TTL = 1000 * 60 * 10;
```
manasaV3, line 52: As we cache the pages for 5 minutes in prod, would it be good to match that?

manasaV3: If the first time the cache is loaded is on page load, wouldn't the CDN cache be more or less in sync with the split.io cache? 😅

manasaV3 (approved):
:shipit:

PR #1184, opened 2023-07-29
Recomputing plugin with plugin_blocked trigger

manasaV3:
## Description

Relates to https://github.com/chanzuckerberg/napari-hub/issues/1182

Depends on: https://github.com/chanzuckerberg/napari-hub/pull/1183

Uses `plugin_blocked` stream for triggering recompute of the current plugin record.

## Notes
Due to how terraform plans and applies changes, the changes from https://github.com/chanzuckerberg/napari-hub/pull/1183 and this PR need to go out as a part of different releases.


napari-hub-commons/src/nhcommons/models/plugin.py
```diff
@@ -46,14 +47,27 @@ class Meta:
...
+        {
+            "hash_key": name,
+            "attributes_to_get": ["name", "version", "release_date"],
+        }
     )
+    plugins = [plugin for plugin in result]
+    if not plugins:
+        return None
+    plugin = sorted(plugins, key=lambda p: p.release_date, reverse=True)[0]
+    return plugin.version
```
aganders3, line 65: Should this always return a singly entry (if any)? I don't think anything in the DB enforces it, but I assume there should only been one `is_latest` plugin for each name.

aganders3 (approved):
Looks good with just one tiny question. I didn't dive into the bookkeeping to see if my assumption is correct but the naming implies it (to me).

napari-hub-commons/src/nhcommons/models/plugin.py
manasaV3: In an ideal world, yes. But, when we have a new plugin, we update the older version of the plugin (if it exists) and add the new version in an eventually consistent way. So, depending on how the execution happens, there could be a very brief period where there might be two versions of the plugin marked as the latest. This handles for that edge case. 😁

richaagarwal: @manasaV3 If it's expected to be a very brief period, would it be worth the tradeoff to simplify this code and optimize for the more common case where there's only one `is_latest`?  Given that there's an expectation for data to be stale for anywhere from 5-10 minutes while the new updates are running, and that we cache for 5 minutes, it seems we could tolerate this brief period of showing the slightly older plugin version.

richaagarwal: Oh that makes sense, thanks for clarifying! In that case, does this code only get run upon certain triggers such as the `plugin_blocked` trigger introduced in this PR, or does it also get called any time we're displaying the plugin page?  If the latter, I wonder if we could make the visibility field not version-dependent, but if the former, then I don't think it's a huge concern.

manasaV3: You are correct, it gets run only when changes to the `plugin_blocked` table triggers the recompute. 😸

richaagarwal (approved):

Issue #1197, opened 2023-08-02
Use nhcommons in modules to remove code duplication


manasaV3:
Update the use of the dynamo models, request package usage, etc. in the `backend`, and `plugin` module to use `nhcommons` module.  

This requires an update on how the image is generated. 

- [x] backend
- [x] plugins

PR #1218, opened 2023-08-14
Refactoring to use nhcommons in plugins module

manasaV3:
## Description
Relates to https://github.com/chanzuckerberg/napari-hub/issues/1197

### Changes Introduced:
- Updates Docker related components to allow napari-hub-commons install
- Restructures tests into a single folder
- Removes writes and checks to s3
- Update to remove s3 permissions and allowing query on plugin_metadata dynamoDB
- Updates to install `libffi7` to support tests on ubuntu version > 20.10


aganders3:
Ah I see this was not caught in CI because I did `pip install -r requirements.txt -r test-requirements.txt` instead of doing it in two steps. This means in CI you first install `npe2 0.7.2`, which correctly pins pydantic, then install `npe2 0.6.1` that is happy with the already-installed pydantic <2.

plugins/Dockerfile
```diff
@@ -1,11 +1,13 @@
-FROM public.ecr.aws/lambda/python:3.8
+FROM public.ecr.aws/lambda/python:3.9
 
 ENV NPE2_NOCACHE = 1
 
-RUN ["yum", "install", "-y", "mesa-libGL"]
+RUN yum install -y mesa-libGL
```
aganders3, line 5:
```suggestion
RUN yum install -y mesa-libGL && yum clean all
```

plugins/Dockerfile
```diff
@@ -1,11 +1,13 @@
...
 
-COPY requirements.txt .
-RUN ["pip", "install", "-r", "requirements.txt"]
+COPY ./napari-hub-commons ../napari-hub-commons
+COPY ./plugins/requirements.txt .
+RUN pip install -r requirements.txt
```
aganders3, line 9:
```suggestion
RUN pip install --no-cache-dir -r requirements.txt
```

aganders3 (approved):
LGTM - thanks for fixing the deps.

These suggested changes to the Dockerfile save about 300 MB in the final image for me.

I'm not sure when/where the production image is built but we should also be careful that the general `COPY` commands to not pull in extra stuff using dockerignore or other techniques. For example locally it was pulling in virtual environments and/or build dirs.

manasaV3:
> I'm not sure when/where the production image is built but we should also be careful that the general COPY commands to not pull in extra stuff using dockerignore or other techniques. 

The happy workflow builds it in a github action, so we don't have to worry about venv getting added to the docker image.  

But, a great call out on `.dockerignore`.  For this project, files related to tests are the only ones I can think of that would benefit from, as there are only a couple of those files, I think we will be fine skipping it in this case.

aganders3 (approved):

Issue #1156, opened 2023-07-17
Get clean code_repository url from pypi "project_urls" data


manasaV3:
To get the code repository, we currently check if there is a `Source Code` field in the  `project_urls` section of the PyPI response. In case the `Source Code` field does not exist, we iterate through the URLs, and fetch the first one, which is a GitHub url.  In those cases, we might fetch options that aren't the best fit. 

For example, look at the snippet below:
```
"project_urls": {
    "documentation": "https://github.com/imagej/napari-imagej#README.md",
    "download": "https://pypi.org/project/napari-imagej/",
    "homepage": "https://github.com/imagej/napari-imagej",
    "source": "https://github.com/imagej/napari-imagej",
    "tracker": "https://github.com/imagej/napari-imagej/issues"
}
```

In here, the code repository, gets set as `"https://github.com/imagej/napari-imagej#README.md"`. This leads to incorrect fetching of data from gitHub. 

We should parse the url to remove any additional anchor tags, query parameter in the url for code repository. 

Also, we should check if the `source` field exists in the project url before iterating over the urls in that section.

bnelson-czi:
I went through and found every plugin with a malformed or no `code_repository`. This include repos with a technically correct GitHub url formats but the repo doesn't exist (maybe it got moved or taken down).

```
ERROR WHEN REQUESTING napari-bud-cell-segmenter REPO: https://github.com/AurelienMaillot/napari-bud-cell-segmenter
ERROR WHEN PARSING napari-nibabel REPO: None
ERROR WHEN PARSING napari-hello REPO: None
ERROR WHEN PARSING napari-svetlana REPO: https://bitbucket.org/koopa31/napari_svetlana/src/main/
ERROR WHEN PARSING napari-pram REPO: None
ERROR WHEN REQUESTING napari-imagej REPO: https://github.com/imagej/napari-imagej#README.md
ERROR WHEN PARSING napari-bioimageio REPO: None
ERROR WHEN PARSING eats-worm REPO: None
ERROR WHEN REQUESTING recOrder-napari REPO: https://github.com/mehta-lab/recOrder/tree/main/recOrder
ERROR WHEN REQUESTING napari-indices REPO: https://github.com/Emmanulla0/napari-indices
ERROR WHEN REQUESTING allencell-ml-segmenter REPO: https://github.com/AllenCell/allencell-ml-segmenter
ERROR WHEN PARSING napari-mri REPO: None
[...]

PR #1219, opened 2023-08-15
Handling code_repository from non "Source Code"

manasaV3:
## Description

Relates to: https://github.com/chanzuckerberg/napari-hub/issues/1156

### Changes introduced:
When fetching code_repository from non-"Source Code" fields, we parse it and clean it if necessary to ensure it conforms to the expected format. 

#### Changes in validation regex explained

Before we validated the urls against this regex: `"^https://github\\.com/([^/]+)/([^/]+)"`.

Now we are validating the urls against the regex: `"^https://github\\.com/[^/]+/[^#./]+"`.

In both cases, we check that the regex starts with `https://github.com/{org}/{repo}` and has a non-zero number of characters that are not `/`. In the newer validation, the repo value is also validated not to have `.` and `#`.

| Input      | Old validation  | New Validation |
| ----------- | ----------- | ----------- |
| "https://github.com/test123/foo1/tree/main/foo"  |  "https://github.com/test123/foo1" |  "https://github.com/test123/foo1" |
| "https://github.com/test123/foo1#README"  |  "https://github.com/test123/foo1#README" |  "https://github.com/test123/foo1" |
| "https://github.com/test123/foo1.git"  |  "https://github.com/test123/foo1.git" |  "https://github.com/test123/foo1" |

#### How is_valid_repo_url is impacted

With the changes, the `is_valid_repo_url` returns `true` whenever there is a match to the expected format `https://github.com/{org}/{repo}` even if they are followed by `#`, `.`, `/`.

| Input      | Old validation  | New Validation |
| ----------- | ----------- | ----------- |
| "https://github.com/test123/foo1/tree/main/foo"  | true |  true |
| "https://github.com/test123/foo1#README"  |  true  |  true  |
| "https://github.com/test123/foo1.git"  |  true |  true |
| "https://github.com/test123/foo1"  | true |  true |
| "https://github.com/test123"  | false |  false |

### DIFF
--- a/napari-hub-commons/src/nhcommons/utils/github_adapter.py
+++ b/napari-hub-commons/src/nhcommons/utils/github_adapter.py
@@ -1,27 +1,34 @@
 import re
-from typing import Dict, Optional
+from typing import Dict, Optional, Match
 
 import yaml
 
 from .adapter_helpers import GithubClientHelper, CitationHelper
 from ..models.plugin_utils import PluginVisibility
 
-_URL_PATTERN = re.compile("^https://github\\.com/([^/]+)/([^/]+)")
-_DEFAULT_DESCRIPTION = "The developer has not yet provided a napari-hub " \
-                       "specific description."
+_GITHUB_URL_PATTERN = re.compile("^https://github\\.com/[^/]+/[^#./]+")
+_DEFAULT_DESCRIPTION = (
+    "The developer has not yet provided a napari-hub " "specific description."
+)
 _PROJECT_URL_NAMES = {
-    'Project Site': 'project_site',
-    'Documentation': 'documentation',
-    'User Support': 'support',
-    'Report Issues': 'report_issues',
-    'Twitter': 'twitter'
+    "Project Site": "project_site",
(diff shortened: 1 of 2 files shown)


aganders3 (approved):

richaagarwal:
@manasaV3 Could you give this PR a more descriptive name?  Also, small nitpick but I think the field name is `code_repository` so it might be better to use that full name when referencing it.

napari-hub-commons/src/nhcommons/utils/github_adapter.py
```diff
@@ -1,27 +1,34 @@
...
+_HUB_CONFIG_KEYS = {"summary", "authors", "labels", "visibility"}
+
+
+def _get_repo_match(url: str) -> Optional[Match]:
+    if not url:
+        return None
+    return _GITHUB_URL_PATTERN.match(url)
 
 
 def is_valid_repo_url(url: str) -> bool:
```
richaagarwal, line 30: Where is this being called?

richaagarwal (approved):

Issue #1199, opened 2023-08-02
Create data parity workflow for commit activity


manasaV3:
Create a workflow for commit activity when we need to update/backfill records for the entire table. This is needed to remove records from dynamo that have been removed in snowflake.

manasaV3:
Example use cases: 

- **napari-bud-cell-segmenter:**  It's [code_repo](https://github.com/AurelienMaillot/napari-bud-cell-segmenter) has been made private now, I believe. Snowflake table had a rewrite, and after the rewrite, has no data for it. But, the dynamo is surfacing old data that was in Snowflake.

- **bfio:** Snowflake had wrong commit data in the past, which still continues to live on in Dynamo. You can see this reflected in the Maintenance timeline for September 2022. The [repo's commit history](https://github.com/PolusAI/bfio/commits/master) doesn't have any data for that time period. The current data is snowflake reflects the correct commit history for that repo but dynamo doesn't.

PR #1222, opened 2023-08-16
Updating maintenance metrics workflow to support overwrite

manasaV3:
## Description
Addresses https://github.com/chanzuckerberg/napari-hub/issues/1199

### Changes introduced
- Adds expiry to the `github-activity` dynamo table infrastructure
- Updates GitHubActivityType to include expiry formatter by type
- Updates query to fetch data for the last 14 months if `type=MONTH` for commit activity
- Updates data accumulator for `type=MONTH` to fill months that don't have an entry with count=0
- Cleans up tests


data-workflows/activity/snowflake_adapter.py
```diff
@@ -137,66 +173,85 @@ def _cursor_to_timestamp_by_name_mapper(accumulator: dict[str, datetime], cursor
...
-    cursor record and added to the accumulator keyed on name.
-    The cursor contains the fields name, timestamp, and count
-    :param dict[str, List] accumulator: Accumulator that will be updated with new data
-    :param SnowflakeCursor cursor:
-    :returns: Accumulator after data from cursor has been added
-   """
-    for name, timestamp, count in cursor:
-        accumulator.setdefault(name, []).append({'timestamp': timestamp, 'count': count})
-    return accumulator
+def _cursor_to_plugin_activity_mapper(
```
aganders3, line 176: The name here was a little confusing (where used above) as this now returns a callable. Maybe `_get_cursor_to_plugin_activity_mapper` or `_make_cursor_to_plugin_activity_mapper`?

data-workflows/activity/tests/test_github_activity_model.py
```diff
@@ -1,30 +1,26 @@
...
-    REPO1: datetime.strptime("05/16/2023 10:24:20", "%m/%d/%Y %H:%M:%S"),
-    REPO2: datetime.strptime("06/26/2023 20:30:00", "%m/%d/%Y %H:%M:%S"),
-}
-FORMATTED_PLUGIN_BY_TS = {
-    repo: ts.replace(day=1) for repo, ts in PLUGIN_BY_EARLIEST_TS.items()
-}
 
 
-def generate_expected(data, granularity, type_id, ts_formatter):
+def generate_expected(data, granularity, type_id, ts_formatter, include_expiry=False):
```
aganders3, line 18:
```suggestion
def generate_expected(data, granularity, type_id, ts_formatter, *, include_expiry=False):
```
nitpick (super-nit as this is in test code): I like boolean args to be keyword-only for clarity at the call site

aganders3 (commented):
I have a couple quick questions/comments on this but mostly looks okay and tests all pass.

As a small request - it looks like you're running a code-formatter in some of these PRs. I totally support this, but can you possibly try to split those changes into a separate commit? It makes it harder to read for actual changes especially as I am not so intimately familiar with this codebase.

data-workflows/activity/github_activity_model.py
```diff
@@ -1,72 +1,85 @@
...
         lambda timestamp: None,
         "TOTAL:{repo}",
         "COUNT(*) AS commit_count",
-        "name"
+        "name",
+        lambda timestamp: None,
     )
 
-    def format_to_timestamp(self, timestamp: datetime) -> Union[int, None]:
+    def format_to_timestamp(self, timestamp: Optional[date]) -> Optional[int]:
```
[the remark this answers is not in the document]
manasaV3: > It's a little hard to trace through here but I don't think this arg can be None. For example it gets passed directly to to_datetime in at least some of the formatters, which will fail. I think this is only an annotation issue though.

In the cases such as when we are fetching `TOTAL`, we end up having `None` passed in for the `timestamp` value.

manasaV3:
> As a small request - it looks like you're running a code-formatter in some of these PRs. I totally support this, but can you possibly try to split those changes into a separate commit (doesn't need to be a separate PR)? It makes it harder to read for actual changes especially as I am not so intimately familiar with this codebase.

Sorry, that it has made it hard to review. I had set the IDE to reformat the file on saving, making it harder to move all formatting changes to a separate commit. I will disable that going forward.

aganders3:
> Sorry, that it has made it hard to review. I had set the IDE to reformat the file on saving, making it harder to move all formatting changes to a separate commit. I will disable that going forward.

No worries! Maybe we can just do one bulk PR first that applies formatting only to all files. Then you can leave on this setting in your IDE to maintain the format for new changes. If this gets in the way for history (blame) you can add that one commit to a `.git-blame-ignore-revs` file in the root dir.

PR #1226, opened 2023-08-18
Return 404 for plugins that don't exist

richaagarwal:
Addresses https://github.com/chanzuckerberg/napari-hub/issues/947

See https://dev-shared-frontend.dev.imaging.cziscience.com/plugins/foobar, vs. what we display now - https://www.napari-hub.org/plugins/foobar


manasaV3:
Yay for more relevant HTTP Status codes!! 🔥

PR #1234, opened 2023-08-25
Adding back random_uuid as patch

manasaV3:
## Description
This is a potential band-aid to help the issue with the cyclic dependency in the production causing deploys to fail.


manasaV3:
No longer needed, fixed without patch.

PR #1235, opened 2023-08-25
Remove `authors` and `summary` from valid `config.yml` keys

DragaDoncila:
## Description
As per the `config.yml` [deprecation plan](https://docs.google.com/document/d/1XNf0KOpJqj1Tzyu9u6KZRdkBTP7m6xSRLExJ6KPhMfM/edit?usp=sharing), this PR removes `authors` and `summary` from the set of valid keys used to parse metadata from `config.yml`. Following this PR, only author information listed in CITATION.cff or core metadata will be parsed and displayed on the hub, and summaries will only be read and displayed from core metadata.


DragaDoncila:
@manasaV3 I couldn't find any tests that check where the metadata is coming from so I'm not sure what the best way would be to verify this is the only required change. Any insight would be appreciated!

DragaDoncila:
@richaagarwal the currently failing preview test will fail until after we release this PR, because it checks that hub metadata matches preview. I think we should remove the test to be honest, if we're going to be deprecating the preview page, but curious what you think? I can remove it in a separate PR if we are happy to.

richaagarwal:
@DragaDoncila I'm fine with removing the preview page tests at this time. If it's easier for you to ignore those failing tests in the meantime, that's fine with me as well (as we'll definitely remove the tests as part of the clean-up after deprecation).

manasaV3 (approved):
🚀

PR #1238, opened 2023-08-28
Remove state dependent preview test to avoid spurious failures

DragaDoncila:
## Description
This preview test fails frequently as we change where we read metadata from, because changes to the metadata need to be released before we can expect the napari hub to match. Since this is so state dependent, and we're deprecating the preview page, I'm removing this test.


richaagarwal (approved):

DragaDoncila:
@manasaV3 even better! Will close this 🎉

PR #1241, opened 2023-08-31
add run_attempt to e2e fixture cache key

codemonkey800:
## Description

Adds `github.run_attempt` to the e2e fixture cache key so that it re-fetches the hub API data fixture

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -21,6 +21,9 @@ defaults:
   run:
     working-directory: frontend/
 
+env:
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -79,7 +81,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: frontend/e2e/fixtures
-          key: e2e-fixtures-${{ github.run_id }}
+          key: e2e-fixtures-${{ github.run_id }}-${{ github.run_attempt }}
```
manasaV3, line 84: Not a blocker: Could this be an env variable that gets referenced in multiple places?

manasaV3 (dismissed):

manasaV3 (approved):

PR #1247, opened 2023-09-06
Set up AWS RUM for the frontend

codemonkey800:
## Description

Sets up AWS RUM on the frontend. This works by passing the AWS RUM credentials to the client at render time on the server, that way we don't need to build the frontend with the credentials already inlined in the code.

- Adds env variables for accessing Cloudwatch RUM config. This includes the app ID, app name, identity pool ID, and group role ARN
- Adds `AwsRumProvider` for giving the application access to the AWS RUM credentials / instance globally
- Sets up happy infra to pass RUM credentials to the frontend server

## Demos

image

### DIFF
--- a/frontend/src/global.d.ts
+++ b/frontend/src/global.d.ts
@@ -5,7 +5,6 @@ declare namespace NodeJS {
     readonly BASE_PATH: string;
     readonly CLOUDWATCH_RUM_APP_ID?: string;
     readonly CLOUDWATCH_RUM_APP_NAME?: string;
-    readonly CLOUDWATCH_RUM_CONFIG?: string;
     readonly CLOUDWATCH_RUM_IDENTITY_POOL_ID?: string;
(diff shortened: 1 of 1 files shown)


frontend/src/global.d.ts
```diff
@@ -3,8 +3,13 @@ declare namespace NodeJS {
     readonly API_URL_HOST: string;
     readonly API_URL: string;
     readonly BASE_PATH: string;
-    readonly ENV: 'local' | 'dev' | 'staging' | 'prod';
+    readonly CLOUDWATCH_RUM_APP_ID?: string;
+    readonly CLOUDWATCH_RUM_APP_NAME?: string;
+    readonly CLOUDWATCH_RUM_CONFIG?: string;
```
manasaV3, line 8: How will this variable be populated?

manasaV3 (approved):
🚢

PR #1262, opened 2023-09-15
Add parsing and mapping for manifest categories to EDAM terms

DragaDoncila:
## Description
As proposed in the `config.yml` deprecation [plan](https://docs.google.com/document/d/1XNf0KOpJqj1Tzyu9u6KZRdkBTP7m6xSRLExJ6KPhMfM/edit?usp=sharing), this PR adds code to parse the categories listed in the `npe2` manifest of a plugin and map them to their corresponding EDAM terms. Manifest categories that do not map to an EDAM term are currently discarded.


data-workflows/plugin/aggregator.py
```diff
@@ -101,3 +102,40 @@ def _get_visibility(name, aggregate, blocked_plugins) -> PluginVisibility:
...
+def _merge_metadata_manifest_categories(metadata, manifest):
+    """Merge categories and hierarchy in PyPI and manifest.
+
+    Keeps union of keys and values present in both dictionaries.
+    """
+    meta_category = metadata.get('category', {})
+    man_category = manifest.get('category', {})
+    meta_category_hierarchy = metadata.get('category_hierarchy', {})
+    man_category_hierarchy = manifest.get('category_hierarchy', {})
+
```
manasaV3, line 115: Nit: if we don't have `category` in the manifest, we can return `metadata` here.

data-workflows/plugin/manifest.py
```diff
@@ -50,6 +61,25 @@ def _get_raw_manifest(manifest_data: Optional[dict[str, Any]],
...
+
+    :param manifest_categories: categories read from the manifest
+    """
+    terms = []
+    for category in manifest_categories:
+        if category in MANIFEST_EDAM_MAPPING:
+            terms.extend(MANIFEST_EDAM_MAPPING[category])
+    labels = {
+        'ontology': ONTOLOGY_VERSION,
+        'terms': terms
```
manasaV3, line 77: Nit: Could we use double quotes here to be more consistent with other files?

data-workflows/plugin/manifest.py
```diff
@@ -50,6 +61,25 @@ def _get_raw_manifest(manifest_data: Optional[dict[str, Any]],
...
+    }
+    categories, hierarchy = process_for_categories(labels, ONTOLOGY_VERSION)
+
+    return categories, hierarchy
```
manasaV3, line 81: Nit: As we aren't using the `categories, hierarchy` variables, we could skip the assignment and directly return the process_for_categories functions result.

data-workflows/plugin/manifest.py
```diff
@@ -50,6 +61,25 @@ def _get_raw_manifest(manifest_data: Optional[dict[str, Any]],
...
+    """Maps raw manifest categories to the appropriate EDAM ontology.
+
+    Any categories that don't map to an EDAM term are discarded.
+
+    :param manifest_categories: categories read from the manifest
+    """
+    terms = []
+    for category in manifest_categories:
+        if category in MANIFEST_EDAM_MAPPING:
+            terms.extend(MANIFEST_EDAM_MAPPING[category])
```
manasaV3, line 74: Nit: Instead of appending to the list, we could create the list with all the values. 
```
terms = [ 
    MANIFEST_EDAM_MAPPING[cat] for cat in manifest_categories if cat in MANIFEST_EDAM_MAPPING
]
```

data-workflows/plugin/aggregator.py
```diff
@@ -101,3 +102,40 @@ def _get_visibility(name, aggregate, blocked_plugins) -> PluginVisibility:
         return PluginVisibility[visibility]
 
     return PluginVisibility.PUBLIC
+
+def _merge_metadata_manifest_categories(metadata, manifest):
```
manasaV3, line 106: Nit: Can we add type hints here?

data-workflows/plugin/manifest.py
```diff
@@ -50,6 +61,25 @@ def _get_raw_manifest(manifest_data: Optional[dict[str, Any]],
 
     return manifest_data
 
+def _map_categories(manifest_categories):
```
manasaV3, line 64: Nit: Can we add type hints here as well?

data-workflows/plugin/tests/test_manifest.py
```diff
@@ -58,6 +60,8 @@ def default_result(self) -> dict[str, Any]:
         ({"npe1_shim": True}, {"npe2": False}),
         ({"npe1_shim": False}, {"npe2": True}),
         ({"display_name": "foo"}, {"display_name": "foo", "npe2": True}),
+        ({"categories": ['not-mapped']}, {"category": {}, "category_hierarchy": {}, "npe2": True}),
+        ({"categories": ['Segmentation', 'other']}, {"category": {"Workflow step": ['Image segmentation']}, "category_hierarchy": {"Workflow step": [["Image segmentation"]]}, 'npe2': True}),
```
manasaV3, line 64: Nit: Can we update this to use double quotes for consistency?

data-workflows/plugin/aggregator.py
```diff
@@ -101,3 +102,40 @@ def _get_visibility(name, aggregate, blocked_plugins) -> PluginVisibility:
...
+        merged_labels = list(set(meta_category.get(key, [])).union(set(man_category.get(key, []))))
+        merged_category[key] = merged_labels
+
+    merged_hierarchy = {}
+    merged_keys = list(set(meta_category_hierarchy.keys()).union(set(man_category_hierarchy.keys())))
+    for key in merged_keys:
+        man_hierarchy = man_category_hierarchy.get(key, [])
+        meta_hierarchy = meta_category_hierarchy.get(key, [])
+        # the lowest level of the hierarchy will be the comparison key for removing duplicates
+        meta_leaves = [hi_list[-1] for hi_list in meta_hierarchy]
```
manasaV3, line 129: It would be more optimal to have the `meta_leaves` be a set, as we are performing a check to see if it contains specific values.

data-workflows/plugin/aggregator.py
```diff
@@ -101,3 +102,40 @@ def _get_visibility(name, aggregate, blocked_plugins) -> PluginVisibility:
...
+        # only keep manifest hierarchies we don't already have
+        man_filtered = list(filter(lambda hi_list: hi_list[-1] not in meta_leaves, man_hierarchy))
+        meta_hierarchy.extend(man_filtered)
+        merged_hierarchy[key] = meta_hierarchy
```
manasaV3, line 133: Nit: Could this be simplified to `merged_hierarchy[key] =  meta_hierarchy + meta_hierarchy`?

manasaV3 (commented):
I have some minor suggestions, otherwise LGTM. 🚀

data-workflows/plugin/aggregator.py
DragaDoncila: I added `dict[str, Any]` as there's a fair few options for the values - let me know if this is sufficient!

data-workflows/plugin/aggregator.py
DragaDoncila: Done!

data-workflows/plugin/aggregator.py
DragaDoncila: Done!

data-workflows/plugin/aggregator.py
DragaDoncila: Done!

data-workflows/plugin/manifest.py
DragaDoncila: Done!

data-workflows/plugin/manifest.py
DragaDoncila: Thanks, I've updated this to be a list comprehension

data-workflows/plugin/manifest.py
```diff
@@ -14,11 +15,23 @@
...
+MANIFEST_EDAM_MAPPING = {
+    "Annotation": ["Image Annotation"],  # Operation
+    "Image_Processing": ["Image Processing"],  # Operation
+    "Measurement": ["Image Characteristics"],  # Data
+    "Segmentation": ["Image Segmentation"],  # Operation > Image Processing
+    "Transformations": [
+        "Geometric Transformation",
+        "Morphological Operation",
+    ],  # Operation > Image Processing
+    "Visualization": ["Visualisation"],  # Operation
```
manasaV3, line 28: Wanted to call out, that I do not see an entry for `"Image Processing"`, `"Image Characteristics"`, `"Geometric Transformation"` in the categories.

manasaV3 (approved):

data-workflows/plugin/aggregator.py
manasaV3: That's good. :)

data-workflows/plugin/manifest.py
DragaDoncila: @manasaV3 thanks for flagging! `Geometric Transformation` was an error on my end - it's `Geometrical Transformation` in the ontology. However, `Geometrical Transformation` and its children are also not present in the category table, so I'm curious what the logic is behind what is and isn't included? Just as an idea, `Morphological Operation` and all of its children are present in the table, and they are all children of `Image Processing`. Other children of `Image Processing` are in the hierarchy, so I would've imagined `Image Processing` also would be, since the initial guidelines were that labels at any level of the hierarchy are supported.

As for `Image Characteristics`, I've removed this one as it anyway would not be useable for filtering, since it's under the `Data` dimension.

manasaV3: @neuromusic might be able to provide more clarity on the process of identifying the categories that were included.

manasaV3 (dismissed):

richaagarwal (approved):

PR #1287, opened 2023-10-17
add metric filter for RUM uncaught errors

codemonkey800:
## Description

Adds a metric filter for RUM error events so that the uncaught error doesn't get triggered for errors that we can safely ignore.

## Demos

The pattern for the filter can be verified using the AWS cli:

```sh
AWS_PROFILE=sci-imaging aws logs describe-metric-filters \
  | jq --raw-output '.metricFilters[] | select(.filterName=="dev-shared-frontend-uncaught-error") | .filterPattern'
```
```
"com.amazon.rum.js_error_event" -"CWR: Failed to retrieve Cognito identity" -"CWR: Failed to retrieve credentials" -"ResizeObserver loop" -"Script error" -"The provided `href`" -"The request is not allowed by the user agent"
```

Tested in `dev-shared` with artificial errors created from the frontend:

image

The first 3 buttons are all known errors that are filtered while the last one is not. Clicking on any of the first 3 buttons correctly reports in RUM:

image

But it does not trigger the alert. However, clicking on the last button will trigger the alert:

image

image

### DIFF
--- a/.happy/terraform/modules/cloudwatch-alarm/main.tf
+++ b/.happy/terraform/modules/cloudwatch-alarm/main.tf
@@ -99,11 +99,38 @@ resource aws_cloudwatch_log_metric_filter frontend_error {
   }
 }
 
+resource aws_cloudwatch_log_metric_filter frontend_uncaught_error {
+  name            = "${var.stack_name}-frontend-uncaught-error"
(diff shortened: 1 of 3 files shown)


manasaV3 (approved):
:shipit:

PR #1291, opened 2023-10-26
Filter out empty support links

codemonkey800:
## Description

Fixes issue where links like `https://napari-hub.org/plugins/[name]` were being clicked on because of an empty value for the documentation link.

### Before

image
image

### After

image


richaagarwal (approved):

manasaV3 (approved):
:shipit:

Issue #1275, opened 2023-09-29
Sporadic issue with PyPI plugins fetch results in batch of plugins incorrectly marked as stale


DragaDoncila:
#### Description
Sporadically, fetching of plugins with the `Framework :: napari` classifier from PyPI results in far fewer plugins than currently listed on the napari hub (for example the [latest such occurrence](https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fprod-data-workflows/log-events/2023$252F09$252F27$252F$255B$2524LATEST$255Ded73c58030ec4fc88ade0ddfbc110719$3Fstart$3D1695796660375$26refEventId$3D37817529233772121602487116682004329335615982358589931544) resulted in just 160 plugins being returned instead of 360). This mismatch leads to the missing plugins being marked stale in Dynamo, and temporarily being removed from the napari hub. Typically, the next workflow run (5 minutes later) restores the missing plugins and they become available again. 

First reported in [this zulip thread](https://napari.zulipchat.com/#narrow/stream/212875-general/topic/napari.20hub.20stream.20noise), it appears this error occurs rarely (twice in September, twice in May and twice in February going back to February). We've opened a PR to silence the zulip stream for now to avoid the noise, but should look into whether there's a way to avoid marking the plugins as stale in this instance.

richaagarwal:
Thank you @DragaDoncila for reporting this and @kephale bringing it to our attention in Zulip. While this is not a critical bug that we will be able to address immediately, we have discussed and will have @DragaDoncila look into the feasibility of addressing this in the coming weeks and then regroup on potential implementation/fix. Once this is addressed, we plan to re-enable the zulip stream as well. 

For transparency, I have added our bug prioritization definitions to this wiki page - https://github.com/chanzuckerberg/napari-hub/wiki/Bug-prioritization.  I am labeling this a P2/ medium priority due to the fact that it is a sporadic enough issue that has yet to prevent users from finding/looking up napari hub plugins. However, since it has the potential to lead to a more critical issue, we are spending time investigating.

DragaDoncila:
[...]

PR #1301, opened 2024-01-16
Validate plugins removed from hub before marking them stale

manasaV3:
## Description
Relates to: https://github.com/chanzuckerberg/napari-hub/issues/1275

In cases where a plugin is being marked stale without a new version, we validate if the version is no longer active by validating it against `npe2api` repository's `public/classifiers.json`. 

### TODO:
Update the documentation to reflect it can now take upto 2 hours for a plugin to be removed from the site.

### DIFF
--- a/.github/workflows/backend-tests.yml
+++ b/.github/workflows/backend-tests.yml
@@ -26,7 +26,7 @@ jobs:
   # Runs pytest for backend code
   tests:
     name: Unit test with pytest
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
 
     steps:
       - name: Checkout Repo
--- a/.github/workflows/backend-tests.yml
+++ b/.github/workflows/backend-tests.yml
@@ -47,7 +47,6 @@ jobs:
       - name: Install dependencies
         # need npe2 here to test preview page
         run: |
-          sudo apt install libffi7
           pip install --upgrade pip
           pip install -r requirements.txt
           pip install -r test-requirements.txt
--- a/.github/workflows/backend-tests.yml
+++ b/.github/workflows/backend-tests.yml
@@ -47,6 +47,7 @@ jobs:
       - name: Install dependencies
         # need npe2 here to test preview page
         run: |
+          sudo apt install libffi7
           pip install --upgrade pip
           pip install -r requirements.txt
           pip install -r test-requirements.txt
--- a/.github/workflows/backend-tests.yml
+++ b/.github/workflows/backend-tests.yml
@@ -26,7 +26,7 @@ jobs:
   # Runs pytest for backend code
   tests:
     name: Unit test with pytest
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
 
     steps:
       - name: Checkout Repo
--- a/.happy/terraform/modules/ecs-stack/main.tf
+++ b/.happy/terraform/modules/ecs-stack/main.tf
@@ -341,6 +341,7 @@ module data_workflows_lambda {
     "GITHUB_CLIENT_ID"      = local.github_client_id
     "GITHUB_CLIENT_SECRET"  = local.github_client_secret
     "PLUGINS_LAMBDA_NAME"   = local.plugins_function_name
+    "ZULIP_CREDENTIALS"     = local.zulip_credentials
   }
 
   log_retention_in_days   = local.log_retention_period
--- a/.happy/terraform/modules/ecs-stack/main.tf
+++ b/.happy/terraform/modules/ecs-stack/main.tf
@@ -341,7 +341,6 @@ module data_workflows_lambda {
     "GITHUB_CLIENT_ID"      = local.github_client_id
     "GITHUB_CLIENT_SECRET"  = local.github_client_secret
     "PLUGINS_LAMBDA_NAME"   = local.plugins_function_name
-    "ZULIP_CREDENTIALS"     = local.zulip_credentials
   }
 
   log_retention_in_days   = local.log_retention_period
--- a/data-workflows/plugin/classifier_adapter.py
+++ b/data-workflows/plugin/classifier_adapter.py
@@ -18,11 +18,11 @@ def _get_recent_query_data() -> dict:
     return data
 
 
-def is_plugin_live(name: str, version: str) -> bool:
+def is_plugin_active(name: str, version: str) -> bool:
     try:
         recent_query_update = _get_recent_query_data()
         active_versions = set(recent_query_update.get("active", {}).get(name, []))
         return version in active_versions
     except RuntimeError:
-        LOGGER.warning(f"Returning {name} {version} is live due to RuntimeError")
+        LOGGER.warning(f"Returning {name} {version} is active due to RuntimeError")
         return True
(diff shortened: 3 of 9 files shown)


.happy/terraform/modules/ecs-stack/main.tf
```diff
@@ -341,6 +341,7 @@ module data_workflows_lambda {
     "GITHUB_CLIENT_ID"      = local.github_client_id
     "GITHUB_CLIENT_SECRET"  = local.github_client_secret
     "PLUGINS_LAMBDA_NAME"   = local.plugins_function_name
+    "ZULIP_CREDENTIALS"     = local.zulip_credentials
```
DragaDoncila, line 344: Is this re-enabling the zulip stream @manasaV3? Perhaps it would be good to do that in a separate PR once we've merged this and confirmed all is working as intended? Ideally tbh I would like to actually capture one of the events reported in #1725 before we re-enable.

Also regardless of that, we should flag to folks on zulip before we re-enable the channel.

data-workflows/plugin/tests/test_classifier_adapter.py
```diff
@@ -0,0 +1,66 @@
...
+        self._github_client_helper.get_file.assert_called_once_with(FILE_NAME, "json")
+
+    def test_handle_invalid_query_data(self):
+        self._classifier_json = {"inactive": []}
+        assert False == plugin.classifier_adapter.is_plugin_live("foo", "1.0.2")
+        assert False == plugin.classifier_adapter.is_plugin_live("bar", "1.0.2")
+        self._github_client_helper_call.assert_called_once_with(REPOSITORY)
+        self._github_client_helper.get_file.assert_called_once_with(FILE_NAME, "json")
+
+    def test_is_plugin_live_caching(self):
```
DragaDoncila, line 49: I don't really understand how this is different from `test_handle_valid_query_data`

data-workflows/plugin/processor.py
```diff
@@ -39,15 +40,19 @@ def _is_new_plugin(plugin_version_pair):
...
-                plugin_metadata_type=PluginMetadataType.PYPI,
-            )
-            if pypi_plugin_version is None:
-                zulip.plugin_no_longer_on_hub(name)
+        if pypi_plugin_version == version or (
+            pypi_plugin_version is None and is_plugin_live(name, version)
+        ):
+            continue
+
+        logger.info(f"Updating old plugin={name} version={version}")
```
DragaDoncila, line 48: What other events (besides a plugin being removed) reach this line of code? The comment says `"update for removed plugins and existing older version of plugins"`, but afaik we only update metadata when a plugin is first released?

data-workflows/plugin/classifier_adapter.py
```diff
@@ -0,0 +1,28 @@
...
+    return data
+
+
+def is_plugin_live(name: str, version: str) -> bool:
+    try:
+        recent_query_update = _get_recent_query_data()
+        active_versions = set(recent_query_update.get("active", {}).get(name, []))
+        return version in active_versions
+    except RuntimeError:
+        LOGGER.warning(f"Returning {name} {version} is live due to RuntimeError")
```
DragaDoncila, line 27: would be good to actually print out the message from the `RuntimeError` here

data-workflows/plugin/tests/test_classifier_adapter.py
```diff
@@ -0,0 +1,66 @@
...
+        self._classifier_json = {"active": {"foo": ["1.0.0", "1.0.1", "1.0.2"]}}
+        assert True == plugin.classifier_adapter.is_plugin_live("foo", "1.0.2")
+        assert False == plugin.classifier_adapter.is_plugin_live("foo", "1.0.4")
+        self._github_client_helper_call.assert_called_once_with(REPOSITORY)
+        self._github_client_helper.get_file.assert_called_once_with(FILE_NAME, "json")
+
+    def test_is_plugin_live_does_not_cache_error(self):
+        self._classifier_json = None
```
DragaDoncila, line 57: would be good to add a comment here, this test is very opaque.

.github/workflows/backend-tests.yml
```diff
@@ -47,6 +47,7 @@ jobs:
       - name: Install dependencies
         # need npe2 here to test preview page
         run: |
+          sudo apt install libffi7
```
DragaDoncila, line 50: What prompted this change?

DragaDoncila (commented):
@manasaV3 thanks for working on this! Left a couple of comments.

.github/workflows/backend-tests.yml
manasaV3: This was to fix issues with ubuntu's latest distro not having the required files, which was causing the tests to fail. I should move this patch to a separate PR.

data-workflows/plugin/classifier_adapter.py
manasaV3: As the run time error generated [here](https://github.com/chanzuckerberg/napari-hub/pull/1301/files#diff-d3d31a1bbe0f387eac4ae3570367d4a7ae2c112ce5d4dce794883d1d5c376bd5) doesn't provide any additional information, I don't think it would be useful to log it. 

Instead, updating the `github_client_helper.get_file(..)` to print the exception before it handles it gracefully will provide the context.

data-workflows/plugin/tests/test_classifier_adapter.py
manasaV3: You are right, this is a duplicate test that should be removed.

.happy/terraform/modules/ecs-stack/main.tf
manasaV3: Agreed that we should decouple adding the Zulip credentials from this PR. Opened [this PR](https://github.com/chanzuckerberg/napari-hub/pull/1304) for that. 

We could let this run in prod for up to a month to see if we can catch the issue occurring. But, if we aren't able to, I think we could add Zulip notifications to let the community know that we are cautiously positive about having fixed the issue. What do you think?

data-workflows/plugin/classifier_adapter.py
```diff
@@ -0,0 +1,28 @@
...
+FILE_NAME = "public/classifiers.json"
+LOGGER = logging.getLogger(__name__)
+
+
+@cache
+def _get_recent_query_data() -> dict:
+    github_client_helper = GithubClientHelper(REPOSITORY)
+    data = github_client_helper.get_file(FILE_NAME, "json")
+    if not data:
+        raise RuntimeError(f"Unable to fetch {FILE_NAME} from {REPOSITORY}")
```
DragaDoncila, line 17: might be worth logging here?

data-workflows/plugin/processor.py
DragaDoncila: Ah yeah of course. Thanks for clarifying!

.happy/terraform/modules/ecs-stack/main.tf
DragaDoncila: @manasaV3 yep I think that sounds reasonable!

DragaDoncila (approved):
@manasaV3 looks good to me - just left a small comment about perhaps logging when the `classifiers` file is unavailable

PR #1303, opened 2024-02-07
Updating ubuntu distros to latest for running tests

manasaV3:
## Description

Updates backend and data workflows tests to run on the latest ubuntu distros.

### DIFF
--- a/.github/workflows/backend-tests.yml
+++ b/.github/workflows/backend-tests.yml
@@ -47,6 +47,7 @@ jobs:
       - name: Install dependencies
         # need npe2 here to test preview page
         run: |
+          sudo apt install libffi7
           pip install --upgrade pip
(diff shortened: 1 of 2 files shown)


richaagarwal (approved):

.github/workflows/backend-tests.yml
```diff
@@ -47,6 +47,7 @@ jobs:
       - name: Install dependencies
         # need npe2 here to test preview page
         run: |
+          sudo apt install libffi7
```
richaagarwal, line 50: Is this a dependency that the latest version requires?

manasaV3: Yeah, the newer versions come with `libffi8` but `libffi7` is needed.

PR #1330, opened 2025-01-07
Add mechnism to ignore commits in release notes

aganders3:
## Description
The release notes workflow depends on specific commit/PR structures, and is currently broken. This PR adds a mechanism to explicitly ignore certain commits. It also ignores the specific problematic commit.

### DIFF
--- a/.github/release-notes.config.js
+++ b/.github/release-notes.config.js
@@ -40,4 +40,11 @@ module.exports = {
       labels: [],
     },
   ],
+
+  /**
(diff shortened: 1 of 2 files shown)


aganders3:
I got impatient and ran the workflow from this branch to test it. It seems to work!
https://github.com/chanzuckerberg/napari-hub/actions/runs/12656760459

richaagarwal (approved):
Thanks for adding this, and sorry for my commit that caused this in the first place 😅

PR #1331, opened 2025-01-07
Use npe2api instead of scraping PyPI for active plugins

aganders3:
## Description
Recent changes to the PyPI website to prevent scraping search results have...prevented us from scraping search results.

npe2api faced the same issue but also uses the public BigQuery PyPI metadata database as a source of truth for active plugins (with the Framework::napari trove classifier). This means the index is only updated every 2h (could possibly be reduced to 1h), but it's a more robust and future-proof solution.

I believe relying on npe2api (https://api.napari.org) as a source of truth is reasonable given the plan to move napari-hub to community support.

### Related
https://github.com/napari/npe2api/pull/35
https://github.com/napari/npe2api/pull/37
https://github.com/napari/npe2api/pull/41
[napari Zulip Thread](https://napari.zulipchat.com/#narrow/channel/309872-plugins/topic/.E2.9C.94.20Plugin.20not.20updating.20in.20napari.20hub/near/491446946)
[PyPI Discussion](https://discuss.python.org/t/fastly-interfering-with-pypi-search/73597/6)

### DIFF
--- a/napari-hub-commons/src/nhcommons/tests/utils/test_pypi_adapter.py
+++ b/napari-hub-commons/src/nhcommons/tests/utils/test_pypi_adapter.py
@@ -137,31 +137,9 @@ class TestPypiAdapter:
     def setup_method(self, monkeypatch):
         monkeypatch.setattr(requests, "get", self._mocked_requests_get)
 
-    def _generate_html_data(self, plugin_version_list: List[Tuple[str, str]]):
-        data = [
-            f"""
-                <div>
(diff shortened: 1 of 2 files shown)


aganders3:
I reverted [ea2562d](https://github.com/chanzuckerberg/napari-hub/pull/1331/commits/ea2562dd311586ab764ee390d38e4c4a44bba142) here and moved it into #1332. Checks will fail on this branch until that is merged and we can rebase.

manasaV3 (approved):
:shipit:

PR #1335, opened 2025-01-14
Remove redundant check for "active" plugins on npe2api

### LINKED ISSUE #1334: Duplicate plugin entries
## Description

On running the plugin update process workflow, we found 499 public plugins (i.e., plugins that are neither hidden nor blocked) on the hub. This was unexpected, as PyPI lists only 496 plugins in total, and the npe2api lists 483.

The reason for this discrepancy is due to duplicate entries for some plugins in the Napari Hub. For example, [the plugin anchor-droplet-chip appears twice](https://www.napari-hub.org/plugins?search=anchor-droplet-chip): once as `anchor-droplet-chip` and once as `anchor_droplet_chip`, even though both refer to the same name and version of the plugin.

**Why is this happening?**

The issue arises from how plugin data is retrieved and processed. The workflow fetches plugin list from the [npe2 API](https://api.napari.org/api/plugins), where the plugin `anchor_droplet_chip` is listed with underscores. However, we also have an old entry for `anchor-droplet-chip` from when PyPI was used as the data source. This older entry should have been marked as stale, since it doesn’t match the entry from npe2api (the new source of truth for the latest plugin version).

To prevent problems caused by intermittent issues with PyPI returning empty lists (which led to the deletion of all plugins on the hub multiple times), we added an extra check before removing any plugin. This check ensures that a plugin is only removed if it is not considered as active by npe2api. (ref: https://github.com/napari/npe2api/blob/main/public/classifiers.json)

However, the issue is that the list we use for this check contains entries with hyphenated names, like pypi  (eg. `anchor-droplet-chip`), and not the underscored versions (e.g., `anchor_droplet_chip`). As a result, the duplicate entry with underscores is skipped for deletion, and continues to be retained.

**TLDR:** The difference in how certain plugins are named in different parts of the npe2api is resulting in duplicate entries.

aganders3:
## Description
This removes the check with npe2api for "active" plugins that prevents plugins from being marked stale. The check was made redundant by #1331. This closes #1334, because relevant plugins (those already in Dynamo with matching non-normalized names on npe2api/plugins) will now be marked "stale" and removed from the public view of the hub.


data-workflows/plugin/processor.py
```diff
@@ -37,25 +37,39 @@ def _is_new_plugin(plugin_version_pair):
...
+                f"(normalized={normalized})"
+            )
+            _mark_version_stale(name, version)
+            continue
+
         pypi_plugin_version = pypi_latest_plugins.get(name)
         if pypi_plugin_version == version:
             continue
+
         if pypi_plugin_version is None and is_plugin_active(name, version):
```
manasaV3, line 62: As we are fetching data from npe2api and not pypi, could we simply remove the check for `is_plugin_active`?

aganders3: Yes, I think we can remove it. Do you mean in addition to the other changes here, or instead?

I still think it makes sense to do the PyPI normalization to ensure a single version of each plugin. In prod I think this also gives continuity with past versions (though I'm not sure this really matters).

aganders3:
I'm not sure why the [Push Remote Dev workflow](https://github.com/chanzuckerberg/napari-hub/actions/runs/12771744643) is failing. It looks okay to me in Terraform.

data-workflows/plugin/processor.py
aganders3: I removed this check in [bcbdd73](https://github.com/chanzuckerberg/napari-hub/pull/1335/commits/bcbdd7310842f18c66100756a0599f86d7e26c29). The check is then unused so I removed the code and tests with it. I'm happy to do whatever is cleanest here 😅.

manasaV3: If we removed the `is_plugin_active` check, it should also take care of removing the duplicate entry. So one of those checks should be enough.

data-workflows/plugin/processor.py
```diff
@@ -42,19 +41,14 @@ def _is_new_plugin(plugin_version_pair):
         pypi_plugin_version = pypi_latest_plugins.get(name)
         if pypi_plugin_version == version:
             continue
-        if pypi_plugin_version is None and is_plugin_active(name, version):
-            logger.info(
-                f"Skipping marking plugin={name} version={version} stale as the "
-                f"plugin is still active in npe2api"
-            )
-            continue
```
aganders3, line 50: This is the main feature of this PR. This should result in removal of the duplicate plugins we were seeing on prod.

manasaV3 (approved):
:shipit:

